PR #9, opened 2020-02-11
remove defaults and trial options

tangiel:


asm/cluster/cluster.yaml
```diff
@@ -31,17 +31,3 @@ spec:
...
-      disabled: false
-    httpLoadBalancing:
-      disabled: false
-    cloudrunConfig:
-      disabled: true # {"type":"boolean","x-kustomize":{"setter":{"name":"cloud-run-disabled","value":"true"}}}
-  ipAllocationPolicy:
-    useIpAliases: true
-  masterAuth:
-    username: admin
-  network: default
```
davidebbo: It might make sense to leave some fields as commented out, so that people who do want to change things can do it more easily. e.g. we know people want to commonly set network/subnetwork, so I'd keep at least that (commented out).

asm/cluster/nodepool.yaml
```diff
@@ -24,25 +24,8 @@ spec:
...
-    - "https://www.googleapis.com/auth/service.management.readonly"
-    - "https://www.googleapis.com/auth/servicecontrol"
-    - "https://www.googleapis.com/auth/trace.append"
-    metadata:
-      disable-legacy-endpoints: "true"
     workloadMetadataConfig:
       nodeMetadata: GKE_METADATA_SERVER
-  management:
-    autoRepair: true
-    autoUpgrade: true
```
davidebbo: I think default is `false` for these two. Which may still be perfectly fine, but this is a small change. From what I can see, the manual docs never had this on, so I'm not worried.

tangiel: `autoRepair` defaults to true for COS containers. But I believe these settings came from the trial.

asm/cluster/cluster.yaml
tangiel: Added network/subnetwork. I don't think the rest need to be included.

PR #25, opened 2020-03-13
set asm branch for anthos sample deployment.

zhihansh:


asm/cluster/nodepool.yaml
```diff
@@ -24,8 +24,19 @@ spec:
...
     machineType: e2-standard-4
+    preemptible: false
+    oauthScopes:
+      - "https://www.googleapis.com/auth/cloud-platform"
+    metadata:
+      disable-legacy-endpoints: "true"
+    serviceAccount: anthos-svc-account@PROJECT_ID.iam.gserviceaccount.com # {"type":"string","x-kustomize":{"setter":{"name":"service-account"}}}
     workloadMetadataConfig:
-      nodeMetadata: GKE_METADATA_SERVER
+      nodeMetadata: EXPOSE
```
elfinhe: Is this EXPOSE required for anthos sample deployment? At least from ASM side, we've moved away from it.

asm/project/enable-services.yaml
```diff
@@ -1,85 +0,0 @@
-# Copyright 2020 Google LLC
```
elfinhe: why delete this file?

asm/cluster/nodepool.yaml
zhihansh: Do you mean the nodeMetadata was EXPOSE and then changed to GKE_METADATA_SERVER in ASM repo? If that's the case, It should be changed to GKE_METADATA_SERVER.

asm/project/enable-services.yaml
zhihansh: The Deployment manager has already enabled all these APIs before installing ASM.

asm/cluster/nodepool.yaml
elfinhe: correct.

asm/project/enable-services.yaml
elfinhe: cool

zhihansh: done

asm/cluster/nodepool.yaml
zhihansh: done

PR #38, opened 2020-04-17
Add asm-patch packages

nan-yu:
The patch package is used to enable ASM on an existing cluster.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,12 @@
+.idea
+*.iml
+.vscode
+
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -0,0 +1,12 @@
+.idea
```
elfinhe, line 1: .gitignore should not be submitted.

nan-yu:
FYI: @justinsb @janetkuo @monopole

PR #39, opened 2020-04-17
Add the validation function to validate ASM features

nan-yu:
This function can be served as a kpt function to validate if the cluster
configurations meet all ASM requirement.
It can be run as:
```
kpt fn source ${CONFIG_DIR} | kpt fn run --image
gcr.io/kustomize-functions/validate-asm:v0.1.0
```


nan-yu:
FYI: @justinsb @janetkuo @monopole

kustomize-functions/validate-asm/image/main.go
```diff
@@ -0,0 +1,347 @@
...
+
+	var errList []error
+
+	meta, err := r.GetMeta()
+	if err != nil {
+		errList = append(errList, err)
+		return errList
+	}
+
+	if meta.Kind == "ContainerCluster" {
```
janetkuo, line 73: Make `ContainerCluster` a const. 

Check API version as well?

kustomize-functions/validate-asm/image/main.go
```diff
@@ -0,0 +1,347 @@
...
+			errList = append(errList, err)
+		}
+
+		// validate if mesh_id lable is set
+		if _, err := validateNodeExists(r, meta, "spec", "labels", "mesh_id"); err != nil {
+			errList = append(errList, err)
+		}
+
+		// validate release channel
+		if err := validateNodeValueIn(r, meta, supportedReleaseChannels,"spec", "releaseChannel", "channel"); err != nil {
```
janetkuo, line 94: It seems that `spec.releaseChannel` is already a required field.

kustomize-functions/validate-asm/Makefile
```diff
@@ -0,0 +1,42 @@
+# Copyright 2020 The Kubernetes Authors.
```
janetkuo, line 1: What copyright should we use?

kustomize-functions/validate-asm/image/main.go
nan-yu: DONE

kustomize-functions/validate-asm/image/main.go
nan-yu: yes, if `spec.releaseChannel` node is missing, it also returns an error with a missing node.

kustomize-functions/validate-asm/Makefile
nan-yu: Changed to the same Google copyright as the other config files.

PR #41, opened 2020-04-28
Migrate kpt setter to openAPI.

yuwenma:
PartialSetter is deprecrated and replaced by kpt substition

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -1,0 +1,103 @@
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+metadata:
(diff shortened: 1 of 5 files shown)


elfinhe:
This may affect our prod installation workflow, let me test it first.

a double-check: have you changed some of the setter names?

yuwenma:
> This may affect our prod installation workflow, let me test it first.
> 
> a double-check: have you changed some of the setter names?

No setter name is changed. But it introduces a new kpt feature substitution. and this requires kpt to be v0.17.0+

PR #48, opened 2020-05-09
Update the kpt validation function

nan-yu:
It includes the following changes:
- Ignore kpt annotations in the node value field
- Remove validation on the master node version as the new requirement only requires using a release channel, but no specific master node versions
- Validate containerCluster's machineType only when it exists
- Make the validator understand both nodeCount and initialNodeCount
- Support Warning level validation errors


nan-yu:
@janetkuo, I'm not able to add you to the reviewers. Here are the fixes to make the kpt function work with both new and existing clusters.

janetkuo:
lgtm

PR #57, opened 2020-05-15
Add configs for Mesh CA in prometheus

elfinhe:


elfinhe:
Tested with staging anthoscli

PR #61, opened 2020-06-05
Update to 1.4.9-asm.1

katze120:
Fixes the security issue, CVE-2020-10739


katze120:
@elfinhe for your review, thanks!

PR #62, opened 2020-06-06
Fix servicemesh typo in config address setters.

incfly:


incfly:
@elfinhe @stewartbutler could you take a look thanks!

PR #72, opened 2020-06-23
Adds the env variables to the ingressgateway.

maasen:


maasen:
@elfinhe can you take a look?

PR #76, opened 2020-06-24
Rev asm version to 1.5.6-asm.0

kenthua:
Instructions in docs reference 1.5.6-asm.0, but 1.5.5-asm.2 is pulled.

Instructions for upgrades in docs has users pulling `istio-1.5.6-asm.0-linux.tar.gz`, which `istioctl` conflicts with our packaged manifests.


elfinhe:
Thanks!

PR #78, opened 2020-06-25
Enable Stackdriver Tracing by default

gargnupur:
Enable Stackdriver Tracing by default and add custom tags  for canonical service and revision


gargnupur:
@maasen: we need this enabled only after a RC is available by noon today... would submitting this before that can cause problems right?

PR #99, opened 2020-07-24
Updating ASM Managed Control Plane configs

stewartbutler:


asm/trafficdirector/td-galley.yaml
```diff
@@ -61,7 +61,7 @@ spec:
         - --validation-port=9443
         - --sinkAuthMode=$(SINKAUTHMODE)
         - --sinkAddress=$(SINKADDRESS)
-        - --sinkMeta=project_id=$(PROJECT_ID),sds_path=$(SDSPATH),trace=0,enable_security_beta=false # {"$ref":"#/definitions/io.k8s.cli.substitutions.sink-meta"}
+        - --sinkMeta=project=$(PROJECT_ID),location=$(GCP_LOCATION),network=projects/$(PROJECT_ID)/global/networks/default,sds_path=$(SDSPATH),trace=0,enable_security_beta=true # {"$ref":"#/definitions/io.k8s.cli.substitutions.sink-meta"}
```
andraxylia, line 64: enable_security_beta=false

andraxylia: Better remove it completely, Galley is not supposed to set this at all.

andraxylia: It turns out we use it for testing, so please leave it but change it to false.

stewartbutler: Set to false.

asm/Kptfile
```diff
@@ -132,17 +142,22 @@ openAPI:
...
         setter:
           name: anthos.servicemesh.enable-security-beta
-          value: "true"
+          value: "false"
     io.k8s.cli.substitutions.sink-meta:
       x-k8s-cli:
         substitution:
           name: anthos.servicemesh.debug.trace-id
-          pattern: "--sinkMeta=project_id=$(PROJECT_ID),sds_path=$(SDSPATH),trace=TRACEID,enable_security_beta=ENABLE_SECURITY_BETA"
+          pattern: "--sinkMeta=project=$(PROJECT_ID),location=$(GCP_LOCATION),network=projects/NETWORK_PROJECT_ID/global/networks/NETWORK_NAME,sds_path=$(SDSPATH),trace=TRACEID,enable_security_beta=ENABLE_SECURITY_BETA"
```
andraxylia, line 150: Can you please remove trace=TRACEID ?

asm/trafficdirector/td-galley.yaml
```diff
@@ -61,7 +61,7 @@ spec:
         - --validation-port=9443
         - --sinkAuthMode=$(SINKAUTHMODE)
         - --sinkAddress=$(SINKADDRESS)
-        - --sinkMeta=project_id=$(PROJECT_ID),sds_path=$(SDSPATH),trace=0,enable_security_beta=false # {"$ref":"#/definitions/io.k8s.cli.substitutions.sink-meta"}
+        - --sinkMeta=project=$(PROJECT_ID),location=$(GCP_LOCATION),network=projects/$(PROJECT_ID)/global/networks/default,sds_path=$(SDSPATH),trace=0,enable_security_beta=false # {"$ref":"#/definitions/io.k8s.cli.substitutions.sink-meta"}
```
andraxylia, line 64: network=projects/$(NETWORK_PROJECT_ID)/global/networks/$(NETWORK_NAME)

stewartbutler: These are actually set by KPT directly. The $(PROJECT_ID) here is because it is using the project ID interpolation from the env variables that are passed. If a user set the kpt setter for network project, it would be replaced by the value of that setter as a plain string (not as an interpolated variable).

asm/Kptfile
stewartbutler: Needed for now, will remove in future.

PR #103, opened 2020-08-06
Support installation with citadel

yangminzhu:
This allows to use Citadel with the asm-gcp profile.

**For new cluster**
1. Get the kpt package with `kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/asm-citadel@release-1.6-asm asm`

**For existing cluster**
1. Get the kpt package with `kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/asm-patch-citadel@release-1.6-asm asm-patch`


yangminzhu:
@elfinhe can you review, we also need to cherry-pick it to release-1.6 and release-1.7 branch. Thanks.

asm-patch/Kptfile
```diff
@@ -124,4 +124,13 @@ openAPI:
...
\ No newline at end of file
+            ref: '#/definitions/io.k8s.cli.setters.gcloud.core.project'
+    io.k8s.cli.substitutions.citadel-patch-base-dir:
+      type: string
+      x-k8s-cli:
+        substitution:
+          name: citadel-patch-base-dir
+          pattern: "../base-dir"
+          values:
+          - marker: base-dir
```
elfinhe, line 135:
```
          pattern: "../base"
          values:
          - marker: base
```

In the `pattern`, it is the actual value we put in the template.

asm/citadel-patch/citadel-patch.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2020 Google LLC
```
elfinhe, line 1: We need discuss more about this asm/citadel-patch/citadel-patch.yaml, it will be detected by anthoscli and generate error like this:

```
Unexpected error: error reconciling objects: error reconciling IstioOperator:/default-istio-operator: gke.io/cluster is required for IstioOperator:/default-istio-operator
```

yangminzhu:
I think I need to move the citadel-patch out of the asm folder, otherwise
it would cause issues anyway because there are 2 IstioOperator in it, any
suggestions?

elfinhe:
hmm, I think the solution here brings more complexity for the problem and is kinda premature.

let's do this:

1. create a folder named `options`
1. create a full copy of istio-operator.yaml with citadel overrides as options/istio-operator-citadelca.yaml, together with options/Kptfile.
2. when user wants a CA option, he need to use extra commands below to override the asm/cluster/istio-operator.yaml and continue.

```
kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/options@release-1.7-asm .
cp -f options/istio-operator-citadelca.yaml asm/cluster/istio-operator.yaml
```

We can do the same to asm-patch. The general is to not change things in the current asm/ or asm-patch/ folders.

How to generate options/ folder when have more combinations is a different topic in the future.

yangminzhu:
> hmm, I think the solution here brings more complexity for the problem and is kinda premature.
> 
> let's do this:
> 
> 1. create a folder named `options`
> 2. create a full copy of istio-operator.yaml with citadel overrides as options/istio-operator-citadelca.yaml, together with options/Kptfile.
> 3. when user wants a CA option, he need to use extra commands below to override the asm/cluster/istio-operator.yaml and continue.
> 
> ```
> kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/options@release-1.7-asm .
> cp -f options/istio-operator-citadelca.yaml asm/cluster/istio-operator.yaml
> ```
> 
> We can do the same to asm-patch. The general is to not change things in the current asm/ or asm-patch/ folders.
> 
> How to generate options/ folder when have more combinations is a different topic in the future.

I agree we should put this in a separate folder but I prefer to use kustomize instead of `cp`. It looks fragile to maintain a copy of the istio-operator.yaml, this will break if it goes out-of-sync with the original one in asm and asm-patch.

asm-patch/Kptfile
yangminzhu: not needed.

asm/citadel-patch/citadel-patch.yaml
yangminzhu: I moved the citadel-patch out of asm and asm-patch, should fix this now.

elfinhe:
> > hmm, I think the solution here brings more complexity for the problem and is kinda premature.
> > let's do this:
> > 
> > 1. create a folder named `options`
> > 2. create a full copy of istio-operator.yaml with citadel overrides as options/istio-operator-citadelca.yaml, together with options/Kptfile.
> > 3. when user wants a CA option, he need to use extra commands below to override the asm/cluster/istio-operator.yaml and continue.
> > 
> > ```
> > kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/options@release-1.7-asm .
> > cp -f options/istio-operator-citadelca.yaml asm/cluster/istio-operator.yaml
> > ```
> > 
> > 
> > We can do the same to asm-patch. The general is to not change things in the current asm/ or asm-patch/ folders.
> > How to generate options/ folder when have more combinations is a different topic in the future.
> 
> I agree we should put this in a separate folder but I prefer to use kustomize instead of `cp`. It looks fragile to maintain a copy of the istio-operator.yaml, this will break if it goes out-of-sync with the original one in asm and asm-patch.

now there are 4 extra commands (3 for adding the patch, 1 for installing kustomize) for using citadel CA in asm, which is not good. Plus, we've got feedback that combining multiple tools in the onboarding flow will block users from adopting asm. Especially the mixed use of both `kustomize` and `kpt` for configuration is not easy to understand for our users who haven't used kustomize.

yangminzhu:
> now there are 4 extra commands (3 for adding the patch, 1 for installing kustomize) for using citadel CA in asm, which is not good. Plus, we've got feedback that combining multiple tools in the onboarding flow will block users from adopting asm. Especially the mixed use of both kustomize and kpt for configuration is not easy to understand for our users who haven't used kustomize.

A few more commands is still much better than disrupting the default flow. Also I don't think it's a requirement to let the user to understand the technical details of kustomize, it's already used in the asm-patch and we will also make it clear in the documentation that the command patches the installation to use Citadel (instead of meshCA).

yangminzhu:
Discussed offline, I changed to make a full copy of the whole asm and asm-patch directory for the citadel installation, it's less commands to run for users (also no kustomize) but harder to maintain all the copies are in sync (@elfinhe will add some presubmit tests for ensure this). Thanks!

PR #105, opened 2020-08-10
Add control plane monitoring option into asm operator file

bianpengyuan:
@elfinhe Not sure if I set the setter correctly, ptal, thanks!

I will cherry-pick this to release-1.7 once this is merged.

### DIFF
--- a/asm-citadel/Kptfile
+++ b/asm-citadel/Kptfile
@@ -133,3 +133,9 @@ openAPI:
           values:
           - marker: PROJECT_ID
             ref: '#/definitions/io.k8s.cli.setters.gcloud.core.project'
+    io.k8s.cli.setters.anthos.servicemesh.controlplane.monitoring.enabled:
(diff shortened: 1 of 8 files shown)


elfinhe:
instead you can use this command to create the config in Kptfile:

```
kpt cfg create-setter asm/ app anthos.servicemesh.monitoring.enabled "true" --type string
```
Then use 
` # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.monitoring.enabled""}`

Also maybe no need to name after stackdriver:
https://cloud.google.com/blog/products/management-tools/cloud-operations-grows-with-monitoring-logging-more

we can use something like `anthos.servicemesh.monitoring.enabled` or `anthos.servicemesh.monitoring.metrics.enabled`

bianpengyuan:
@elfinhe updated. ptal.

PR #112, opened 2020-08-11
Remove *-citadel/, now users can get them from pkg/ by using kpt

elfinhe:
Mesh CA:
```
kpt pkg get https://github.com/elfinhe/anthos-service-mesh-packages.git/asm .
```
Extra commands to use Citadel CA:
```
rm -rf asm/istio
kpt pkg get https://github.com/elfinhe/anthos-service-mesh-packages.git/pkg/istio-citadel asm/istio
```


elfinhe:
ptal @yangminzhu

elfinhe:
cc @yanyuan-meiyi

yangminzhu:
this seems refactoring the istio-operator.yaml to a separate directory and download it again for citadel? better than duplicating the whole package, does this need any documentation update in the current flow?

yangminzhu:
> > this seems refactoring the istio-operator.yaml to a separate directory and download it again for citadel? better than duplicating the whole package, does this need any documentation update in the current flow?
> 
> that's right. no change on the current asm flow.

I think we need a small change in the documentation for installation with istioctl because the path to the istio-operator.yaml is changed.

PR #113, opened 2020-08-11
[Cherrypick] Add control plane monitoring option into asm operator file (#105) (#110)

elfinhe:
Cherry pick control plane monitoring option into master.


bianpengyuan:
Thanks! I thought master is not up to dated. I will keep that in mind.

PR #117, opened 2020-08-12
Add packages to support multiproject profile

yanyuan-meiyi:
I decided to choose option 3. It probably makes the most sense.


yanyuan-meiyi:
cc: @elfinhe, @yangminzhu

yanyuan-meiyi:
> mind explaining 1. how the files are generated (e.g., steps used, starting from which branch/folder), and 2. what changes are made on existing istio-operator.yaml, cluster.yaml, or any other?

The files are generated from the master branch by running make all under pkg folder (see the updates in pkg/Makefile).

In the existing istio-operator.yaml (asm-gcp), there is no change. I added a new istio-operator.yaml for multiproject profile. For cluster.yaml, I added setters to configure service and cluster secondary range. This is required when we trying to use shared network (shared VPC or VPC peering). Anthoscli will ignore these if they are set as empty (when we are using VPC like `default`). Note that this change is only for creating new clusters. Note that there is no change in cluster.yaml in the patch folder. This is because network configuration cannot be updated once the cluster has been created.

PR #118, opened 2020-08-12
add pkgs for multiproject

yanyuan-meiyi:
The packages are generated by running Make all in this PR: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/117.


yanyuan-meiyi:
cc @elfinhe @yangminzhu

asm-multiproject-patch/Kptfile
```diff
@@ -0,0 +1,184 @@
...
+      x-k8s-cli:
+        setter:
+          name: gcloud.container.cluster.releaseChannel
+          value: REGULAR
+    io.k8s.cli.setters.anthos.servicemesh.trustDomainAliases:
+      type: array
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.trustDomainAliases
+          listValues:
```
elfinhe, line 47: can you talk with Phani to double check that array is well supported by kpt?

yanyuan-meiyi: Just confirmed with Phani, it is supported and we can use it.

asm-multiproject/istio-multiproject/istio-operator.yaml
```diff
@@ -0,0 +1,51 @@
...
+          value: "true"
+        - name: SPIFFE_BUNDLE_ENDPOINTS
+          value: "PROJECT_ID.svc.id.goog|https://storage.googleapis.com/mesh-ca-resources/spiffe_bundle.json" # {"$ref":"#/definitions/io.k8s.cli.substitutions.spiffe-bundle-endpoints"}
+  meshConfig:
+    defaultConfig:
+      proxyMetadata:
+        GCP_METADATA: "PROJECT_ID|PROJECT_NUMBER|asm-cluster|us-central1-c" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gke-metadata"}
+        TRUST_DOMAIN: "PROJECT_ID.svc.id.goog" # {"$ref":"#/definitions/io.k8s.cli.substitutions.trust-domain"}
+        GKE_CLUSTER_URL: "https://container.googleapis.com/v1/projects/PROJECT_ID/locations/us-central1-c/clusters/asm-cluster" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gke-cluster-url"}
+    trustDomainAliases: # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.trustDomainAliases"}
```
yangminzhu, line 39: I'm still not sure why we cannot reuse the asm directory and only change the asm/cluster/istio-operator.yaml?

It looks like we can just have the default empty values for trustDomainAliases and clusterName, or maybe there are some other changes we can not combine there?

yanyuan-meiyi: This is definitely a possibility. Actually I asked Jianfei about merging this two profile together. The issue here is we need to put something under trustDomainAlias (we cannot assign an empty array to the setter). If we think it is safe to put some stuff under trustDomainAlias (I was thinking to put "PROJECT_ID.svc.id.goog", but it looks like it will need some tests to confirm it will work), then we probably can merge them together.

yanyuan-meiyi:
> Also you didn't change the asm-citadel and asm-patch-citadel?

No. I need to add two more packages for asm-multiproject-citadel and asm-multiproject-patch-citadel.

yangminzhu:
> > Also you didn't change the asm-citadel and asm-patch-citadel?
> 
> No. I need to add two more packages for asm-multiproject-citadel and asm-multiproject-patch-citadel.

sorry I wasn't clear, I meant to say why not add them in the same PR.

asm-multiproject/istio-multiproject/istio-operator.yaml
```diff
@@ -0,0 +1,51 @@
+# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
yanyuan-meiyi: Moved the files so the folder structure of all packages are the same.

asm-multiproject/istio-multiproject/istio-operator.yaml
yangminzhu: it can just be an empty string by default, @incfly  to confirm. And if that's the only change plus a few more in istio-operator.yaml to support the multi-project profile, we should try to add them to asm, having 8 duplicate directories with only minor changes in istio-operator.yaml is just crazy...

yanyuan-meiyi:
> > > Also you didn't change the asm-citadel and asm-patch-citadel?
> > 
> > 
> > No. I need to add two more packages for asm-multiproject-citadel and asm-multiproject-patch-citadel.
> 
> sorry I wasn't clear, I meant to say why not add them in the same PR.

I was still working on those when I sent the PR. I will include those in this PR.

yanyuan-meiyi:
> > > > Also you didn't change the asm-citadel and asm-patch-citadel?
> > > 
> > > 
> > > No. I need to add two more packages for asm-multiproject-citadel and asm-multiproject-patch-citadel.
> > 
> > 
> > sorry I wasn't clear, I meant to say why not add them in the same PR.
> 
> I was still working on those when I sent the PR. I will include those in this PR.

Updated the asm-citadel package for multiproject. Note that we can directly use asm-patch-citadel for multiproject, since we cannot change the cluster network settings once the cluster is created.

asm-multiproject/istio-multiproject/istio-operator.yaml
incfly: An empty array for trust domain alias should be safe (for non-meshca-multi-project cases).

But @yanyuan-meiyi have you mentioned there're some kpt issues about that? I don't recall what it is, sometihng like set it with some value, unable to set it back to empty or something. I'm not sure.

incfly: (my first reply is too fast... deleted)

I verified that the alias as same as the trust domain would be a no-op, compared with not setting the aliases at all. That should address Yuan's concern of not able to resetting it back to empty.

```
trustDomain: "proj.id.svc.goog"
trustDomainAliases: ["proj.id.svc.goog"]
```

yanyuan-meiyi: > (my first reply is too fast... deleted)
> 
> I verified that the alias as same as the trust domain would be a no-op, compared with not setting the aliases at all. That should address Yuan's concern of not able to resetting it back to empty.
> 
> ```
> trustDomain: "proj.id.svc.goog"
> trustDomainAliases: ["proj.id.svc.goog"]
> ```

Thanks! So we will then merge asm-gcp and asm-gcp-multiproject into one kpt package. Other than the network configurations, we will add two more setters: trustdomainalias and an additional environment variable PILOT_SKIP_VALIDATE_TRUST_DOMAIN. By default (asm-gcp), trustdomainalias will be the same as trustdomain, PILOT_SKIP_VALIDATE_TRUST_DOMAIN will be set as "false", so users using asm-gcp don't need to worry about these new settings. @elfinhe, does it sound good to you?

yanyuan-meiyi: Thanks! So we will then merge asm-gcp and asm-gcp-multiproject into one kpt package. Other than the network configurations, we will add two more setters: trustdomainalias and an additional environment variable PILOT_SKIP_VALIDATE_TRUST_DOMAIN. By default (asm-gcp), trustdomainalias will be the same as trustdomain, PILOT_SKIP_VALIDATE_TRUST_DOMAIN will be set as "false", so users using asm-gcp don't need to worry about these new settings. @elfinhe, does it sound good to you?

asm-multiproject-patch/resources/istio-operator.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: default-istio-operator
+spec:
+  profile: asm-gcp-multiproject # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.profile"}
+  hub: gcr.io/gke-release/asm # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.hub"}
+  tag: 1.6.5-asm.7 # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.tag"}
+  components:
+    pilot:
+      k8s:
+        env:
+        - name: PILOT_SKIP_VALIDATE_TRUST_DOMAIN
```
incfly, line 29: Update: This one is no longer needed after we have the cherry-picked a fix into 1.6-asm branch. @yangminzhu right?

incfly: Let me quickly verify this before the end of today without setting this plus including the fix, see if that works.

incfly: Yes. You can now ignore the `PILOT_SKIP_VALIDATE_TRUST_DOMAIN`.  It should not matter after another fix is merged.

https://gke-internal-review.googlesource.com/c/istio/istio/+/160952

yanyuan-meiyi: Awesome. Thanks a lot!

asm-multiproject-patch/resources/istio-operator.yaml
```diff
@@ -0,0 +1,51 @@
+# Copyright 2020 Google LLC
```
yangminzhu, line 1: I'm still not clear how will the customer use citadel with multiproject with this PR? This istio-operator.yaml doesn't include anything related to Citadel and also you don't have another directory `asm-citadel-multiproject-patch`?

Maybe the PR is not updated with latest changes yet?

asm-multiproject/istio-multiproject/istio-operator.yaml
elfinhe: Sounds good if the behaviors are the same with `trustDomainAliases: [<trust-domain>]` and  `trustDomainAliases:`.

asm-multiproject-patch/resources/istio-operator.yaml
yanyuan-meiyi: The customer will actually use `asm-citadel-patch` to install multiproject, since we can completely reuse the same package.

yangminzhu: how? the `asm-citadel-patch` is quite different from the asm-multiproject-patch.

And also how does the customer use citadel with ` asm-multiproject`? Can you list the commands the user needed to run?

yanyuan-meiyi: For the current change in this PR. The customers need to use `asm-multiproject` pkg and `asm-multiproject-patch`pkg for mesh ca installation, and use `asm-citadel` pkg and `asm-citadel-patch` pkg for citadel installation. 

We can reuse `asm-citadel` and `asm-citadel-patch`.

asm-multiproject/istio-multiproject/istio-operator.yaml
yanyuan-meiyi: Actually, my previous comment was wrong, I think the user needs to configure trustdomainalias if we merge the packages together in this way. Right now, we use substitution in kpt to populate trust domain based on this pattern `PROJECT_ID.id.svc.goog`. However, we don't have something like substitution array. We need to use setters. That means the customer needs to manually set this field.

incfly: hmm, can we work around by writing yamls as 

```yaml
trustDomainAliases:
- PROJECT_ID.svc.goog.id
```

If the issue is about kpt's setters in the string prefix in a array. (on the other hand, i'm wondering what's the standard kpt's answer for this kinda flexiblity, we have run into multiple syntax limitations so far)

yanyuan-meiyi: Hmm.. For setters, I don't think we can do that. The setter will overwrite the whole string. Let me ask if there is something can behave like an array of substitution.

yanyuan-meiyi: Just pinged Phani on this. Right now, we cannot do something like "array of substitution" using kpt.

PR #121, opened 2020-08-16
Revert "Add packages to support multiproject profile"

yanyuan-meiyi:
Reverts GoogleCloudPlatform/anthos-service-mesh-packages#117


yanyuan-meiyi:
Revert this change in master branch since we will merge asm and asm-multiproject together.

PR #123, opened 2020-08-20
add canonical service controller to 1.6

PatchworkGuilt:
As we outlined in go/canonical-service-asm-build, this adds the canonical service controller to the ASM distro release.

To save your time during review: this is the exact same controller.yaml file copy/pasted across 4 folders.

### DIFF
--- a/asm-citadel/cluster/canonical-service/controller.yaml
+++ b/asm-citadel/cluster/canonical-service/controller.yaml
@@ -1,0 +1,341 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  annotations:
+    gke.io/cluster: "gke://PROJECT_ID/us-central1-c/asm-cluster" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gkeio-cluster"}
+  name: asm-system
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.5
+    gke.io/cluster: "gke://PROJECT_ID/us-central1-c/asm-cluster" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gkeio-cluster"}
+  creationTimestamp: null
+  name: canonicalservices.anthos.cloud.google.com
+spec:
+  group: anthos.cloud.google.com
+  names:
+    kind: CanonicalService
+    listKind: CanonicalServiceList
+    plural: canonicalservices
+    singular: canonicalservice
+  scope: Namespaced
+  validation:
+    openAPIV3Schema:
+      description: CanonicalService is the Schema for the canonicalservices API
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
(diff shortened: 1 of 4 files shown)


asm-patch-citadel/resources/canonical-service/controller.yaml
```diff
@@ -0,0 +1,341 @@
...
+        name: kube-rbac-proxy
+        ports:
+        - containerPort: 8443
+          name: https
+      - args:
+        - --metrics-addr=127.0.0.1:8080
+        - --enable-leader-election
+        command:
+        - /manager
+        image: gcr.io/gke-release-staging/asm/canonical-service-controller:1.6.8-asm.2
```
yangminzhu, line 332: probably make the hub/tag configurable in kpt so easier for testing other images. Also you shouldn't use gke-release-staging as it's not accessible for external users.

PatchworkGuilt:
@elfinhe Integration tests are run as part of the image build process.  They can be found here: https://team.git.corp.google.com/csm-ux-eng/canonical-service/+/refs/heads/master/controllers/

asm-patch-citadel/resources/canonical-service/controller.yaml
PatchworkGuilt: Peter set up a build flow for it, but the image isn't available on gke-release yet (only staging).  Is that something that will happen automatically as part of a release or do we need to do something?

asm-citadel/cluster/canonical-service/controller.yaml
```diff
@@ -0,0 +1,341 @@
...
+kind: ClusterRole
+metadata:
+  name: canonical-service-proxy-role
+  annotations:
+    gke.io/cluster: "gke://PROJECT_ID/us-central1-c/asm-cluster" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gkeio-cluster"}
+rules:
+- apiGroups:
+  - authentication.k8s.io
+  resources:
+  - tokenreviews
```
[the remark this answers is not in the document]
PatchworkGuilt: To be honest I'm not sure.  Those rules were made automatically by the kubebuilder boilerplate.

asm-citadel/cluster/canonical-service/controller.yaml
```diff
@@ -0,0 +1,341 @@
...
+  name: canonical-service-manager-rolebinding
+  annotations:
+    gke.io/cluster: "gke://PROJECT_ID/us-central1-c/asm-cluster" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gkeio-cluster"}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: canonical-service-manager-role
+subjects:
+- kind: ServiceAccount
+  name: default
```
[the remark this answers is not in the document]
PatchworkGuilt: Makes sense (this was what the default was for the boiler plate).  What does creating a Service Account with yaml look like?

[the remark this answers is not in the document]
yangminzhu: See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/, it's just a few lines, and then you need to modify the deployment to use the service account you created.

asm-patch-citadel/resources/canonical-service/controller.yaml
yangminzhu: this needs to be updated to gke-release otherwise the user couldn't install it (they will just see image pull failed error), the alternative is to let the user to configure it to point to the correct gke-release but that's an extra step for users and requires update in documentation.

asm-citadel/cluster/canonical-service/controller.yaml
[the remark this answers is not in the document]
yangminzhu: I don't think the canonical service controller would need the tokenreview permission, can you check if it works if you remove this.

asm-patch-citadel/resources/canonical-service/controller.yaml
PatchworkGuilt: Yeah I'm happy to I just meant: the image doesn't _exist_ in gke-release yet.  What do we need to do to make that happen?

yangminzhu: can you check with peter, not sure we should discuss the internal release on github.

michaelsun3:
what is the canonical-service-controller for?

PR #124, opened 2020-08-20
Update asm* with latest KCC compatible annotations to support canonical services installation

elfinhe:
Update asm* with latest KCC compatible annotations to support canonical services installation


asm-citadel/Kptfile
```diff
@@ -40,6 +40,16 @@ openAPI:
         setter:
           name: gcloud.container.cluster.releaseChannel
           value: REGULAR
+    io.k8s.cli.setters.anthos.servicemesh.trustDomainAliases:
```
yanyuan-meiyi, line 43: We probably don't need to configure trustDomainAliases for citadel.

asm-patch-citadel/Kptfile
```diff
@@ -41,6 +41,16 @@ openAPI:
         setter:
           name: anthos.servicemesh.tag
           value: 1.6.8-asm.0
+    io.k8s.cli.setters.anthos.servicemesh.trustDomainAliases:
```
yanyuan-meiyi, line 44: Same as asm-citadel.

PR #131, opened 2020-09-05
Update multiproject

yanyuan-meiyi:


asm/istio/istio-operator.yaml
```diff
@@ -33,9 +33,11 @@ spec:
   meshConfig:
     defaultConfig:
       proxyMetadata:
-        GCE_METADATA_HOST: "metadata.google.internal"
```
yanyuan-meiyi, line 36: This field doesn't exist in the pkg/istio folder. Should we keep this one?

elfinhe: feel free to remove it here. I will update it in the profile to use the cluster info from GKE metadata server.

elfinhe (commented):
I'd suggest we create an overlay for multiproject (e.g., multiproject.yaml), then we can use

```
istioctl install -f istio-operator.yaml -f multiproject.yaml
```
to enable multi-project features.

We didn't do this before because anthoscli does not support multiple IstioOperator overlays. Now we can do it in 1.7.

asm/istio/multiproject.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  meshConfig:
+    trustDomainAliases: # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.trustDomainAliases"}
+    - "PROJECT_ID.svc.id.goog" # {"$ref":"#/definitions/io.k8s.cli.substitutions.trust-domain"}
```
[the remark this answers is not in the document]
yanyuan-meiyi: Moved the yaml file to options folder.

elfinhe (approved):

elfinhe (approved):

PR #132, opened 2020-09-08
Set caAddress to emtpy to be able to support both revision and non-re…

elfinhe:
Set caAddress to emtpy to be able to support both revision and non-revision use cases.

The template in charts will choose the right CA address based on whether revision is set or not.


yangminzhu (commented):
Also change the one in asm-patch-citadel?

elfinhe:
> Also change the one in asm-patch-citadel?

yep, updated, thanks

yangminzhu (approved):

PR #133, opened 2020-09-08
Refactoring ASM pkgs

elfinhe:
1. asm-patch: remove istio related resources. In 1.7, asm-patch will be only used to update clusters by anthoscli (focusing on infra provisioning).
1. pkg/istio/istio-operator.yaml: remove all cluster related fields. Use auto-detection from metadata server.
1. pkg/istio/options: add all optional features as overlays, e.g., citadel-ca, multiproject, multicluster.


yanyuan-meiyi (approved):

yangminzhu (commented):
What's the difference between pkg/istio/options/citadel-ca.yaml and asm/istio/options/citadel-ca.yaml?

elfinhe:
> citadel-ca.yaml

no difference between them, probably I need to remove the pkg/ folder to reduce duplicates.

yangminzhu (approved):

PR #136, opened 2020-09-10
add canonical-service to 1.7

PatchworkGuilt:
These are the same changes we added to the 1.6 branch, simplified for 1.7.

The content of the two controller.yaml files are identical.

I will follow up with a similar PR adding these changes to master so that we don't have to "forward port" this for the next release.

### DIFF
--- a/asm-patch/Kptfile
+++ b/asm-patch/Kptfile
@@ -93,6 +93,11 @@ openAPI:
         setter:
           name: anthos.servicemesh.tag
           value: 1.6.8-asm.9
+    io.k8s.cli.setters.anthos.servicemesh.canonicalServiceHub:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.canonicalServiceHub
+          value: gcr.io/gke-release/asm/canonical-service-controller:1.6.8-asm.9
     io.k8s.cli.substitutions.mesh-id:
       type: string
       x-k8s-cli:
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -93,6 +93,11 @@ openAPI:
         setter:
           name: anthos.servicemesh.tag
           value: 1.6.8-asm.9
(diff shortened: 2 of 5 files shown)


asm-patch/Kptfile
```diff
@@ -93,6 +93,11 @@ openAPI:
         setter:
           name: anthos.servicemesh.tag
           value: 1.6.8-asm.9
+    io.k8s.cli.setters.anthos.servicemesh.canonicalServiceHub:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.canonicalServiceHub
+          value: gcr.io/gke-release/asm/canonical-service-controller:1.6.8-asm.9
```
elfinhe, line 100: will we update the canonical-service-controller tag to 1.7.1-asm.x?

asm/Kptfile
```diff
@@ -93,6 +93,11 @@ openAPI:
         setter:
           name: anthos.servicemesh.tag
           value: 1.6.8-asm.9
+    io.k8s.cli.setters.anthos.servicemesh.canonicalServiceHub:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.canonicalServiceHub
+          value: gcr.io/gke-release/asm/canonical-service-controller:1.6.8-asm.9
```
elfinhe, line 100: same version tag question - do we need to update?

elfinhe (commented):
LGTM

asm-patch/Kptfile
PatchworkGuilt: I'd like to, but that doesn't seem to have happened yet.  Right now this is the only available version tag.  We can absolutely update it when it's available, but functionally it would be a no-op.

asm/Kptfile
PatchworkGuilt: Responded above. same answer here.

elfinhe (approved):

PR #147, opened 2020-09-15
Add meshexpansion as an option in 1.7

jasonwzm:
Only needed in 1.7 release.

/cc @maasen


elfinhe (commented):
LGTM.

np: rename to mesh-expansion.yaml.

jasonwzm:
> LGTM.
> 
> np: rename to mesh-expansion.yaml.

done

PR #149, opened 2020-09-16
Update usage to reflect default of Mesh CA for new installs

djazayeri:


scripts/asm-installer/install_asm
```diff
@@ -166,16 +166,18 @@ OPTIONS:
...
-                                      MODE=install defaults to using Mesh CA
-                                      as the certificate authority.
-  -c|--ca                <CA>         The type of certificate authority to be
-                                      used. Allowed values for <CA> are
-                                      {citadel|mesh_ca}.
-
+                                      {install|migrate}. Choose MODE=install
+                                      for a new ASM installation or MODE=migrate
+                                      for an existing OSS installation to be
+                                      migrated to ASM. MODE=install defaults to
```
zerobfd, line 172: Can we change "MODE=install" to "--mode install"? Caps vs not is optional, but in the CA field we use lowercase so we should be standard. We shouldn't use an equal sign since the script doesn't support using it.

scripts/asm-installer/install_asm
```diff
@@ -166,16 +166,18 @@ OPTIONS:
...
+                                      using Mesh CA as the certificate
+                                      authority.
+
+  -c|--ca                <CA>         (optional) The type of certificate
```
zerobfd, line 176: If it's required for migrate mode, let's not put the (optional) tag in front.

zerobfd (approved):
LGTM after a couple of small style points

scripts/asm-installer/install_asm
djazayeri: Sure. I originally tried this but it led to an unfortunate line break which made things a bit confusing. But I'll do it that way, since it is in fact correct.

scripts/asm-installer/install_asm
djazayeri: I am following the precedent from

 -k|--key_file          <FILE PATH>  (optional) The key file for a service
                                      account. Required if --service_account is
                                      passed.

elfinhe:
do we know why the test is failing? I can either help make an exception merge if we have confidence that it is a flaky run.

scripts/asm-installer/install_asm
zerobfd: That's a good point. When I wrote it, service account and key file are both optional, so it made sense. This one seems different to me because the dependency is on a mandatory option, not an optional one.

If this might be confusing for users, I'd honestly rather just remove the (optional) from the beginning altogether and let users read the descriptions for more info. I don't see a way to have the two cases be consistent to everyone otherwise.

I won't block on this if you think this is the right way to handle it.

zerobfd:
We don't know why the test is failing--I'm looking at it now but don't have an idea on how long it will take. Since it's just a usage change I think an exception merge should be fine--runs before this PR also failed and I'm pretty doubtful this is the culprit.

PR #152, opened 2020-09-17
Add test to check downtime during migration

ruigulala:


elfinhe (commented):
LGTM, maybe we need an approval from @zerobfd

scripts/asm-installer/tests/run_migration_suite_citadel
```diff
@@ -57,6 +57,12 @@ main() {
   verify_demo_app "${GATEWAY}"
 
   # Test starts here
+
+  # Send persistent traffic to demo app to test any downtime.
+  echo "Start sending traffic to demo app"
+  send_persistent_traffic "${GATEWAY}" &
+  local SEND_TRAFFIC_CMD_PID="${!}"
```
zerobfd, line 64: It probably won't happen, but if for some reason we fail to fork then this could be empty or -1. Later on you kill the pid. Bad things happen if you try to kill pid -1, especially if you happen to be root. Let's check the value of this immediately after forking to make sure the PID exists before continuing.

I _believe_ that the way you're using ps -p makes this safe, but 1) someone might change that later and forget about this, and 2) if we fail to fork then we probably don't want to continue the test anyways.

scripts/asm-installer/tests/common.sh
```diff
@@ -190,6 +190,17 @@ verify_demo_app() {
...
 }
+
+send_persistent_traffic() {
+  local IP; IP="$1"
+  local PAGE
+  local COUNT
+  while true; do
+    PAGE="$(curl "${IP}" -L -N -s || echo "")"
+    COUNT="$(echo "${PAGE}" | grep -c Hipster || true)"
+    { [[ "${COUNT}" -eq 0 ]] && echo -e "Receive unexpected response from demo app:\n${PAGE}" && return 1; } || sleep 1
```
zerobfd, line 201: This still works as expected if you use exit 1 instead of return 1, right? If so then I'd prefer that as it makes the intention clearer.

zerobfd (approved):
Small issues otherwise LGTM. Let's try to get the tests to re-run, it looks like the failure wasn't related to this.

scripts/asm-installer/tests/run_migration_suite_citadel
ruigulala: Done.

scripts/asm-installer/tests/common.sh
ruigulala: Done.

PR #153, opened 2020-09-17
Make usage prompt more consistent

zerobfd:
* Remove (optional) tag since there is no reasonable way to be
 consistent between different types of conditionally optional options
 * Remove spacing between option types
 * Separate flags from options
 * Add a period to all descriptions
 * Move allowed values to the end of descriptions
 * Standardize descriptions for optional options on "This can be omitted
 if..."
 * Language tweaks for parallelism


djazayeri (approved):
lgtm

PR #154, opened 2020-09-17
Updated minimum resource requirements for script install

djazayeri:
We now require >= 8 vCPUs in node pools whose machine size is >= 4 vCPUs.
(So this changes the structure of the requirement.)

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -548,7 +548,12 @@ validate_node_pool() {
 
   info "Confirming node pool requirements..."
   local ACTUAL_CPU
-  ACTUAL_CPU="$(list_valid_pools "${MACHINE_CPU_REQ}" | jq '.[] | (if .autoscaling.enabled then .autoscaling.maxNodeCount else .initialNodeCount end) * (.config.machineType / "-" | .[-1] | tonumber)' )"
+  ACTUAL_CPU="$(list_valid_pools "${MACHINE_CPU_REQ}" | \
+      jq '.[] |
+        (if .autoscaling.enabled then .autoscaling.maxNodeCount else .initialNodeCount end)
+        *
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/install_asm
```diff
@@ -546,17 +543,19 @@ EOF
...
 validate_node_pool() {
-  local NODE_REQ; NODE_REQ=4; readonly NODE_REQ;
-  local CPU_REQ; CPU_REQ=4; readonly CPU_REQ;
+  local MACHINE_CPU_REQ; MACHINE_CPU_REQ=4; readonly MACHINE_CPU_REQ;
+  local TOTAL_CPU_REQ; TOTAL_CPU_REQ=8; readonly TOTAL_CPU_REQ;
 
   info "Confirming node pool requirements..."
-  if [[ -z "$(list_valid_pools "${NODE_REQ}" "${CPU_REQ}")" ]]; then
+  local ACTUAL_CPU
+  ACTUAL_CPU="$(list_valid_pools "${MACHINE_CPU_REQ}" | jq '.[] | (if .autoscaling.enabled then .autoscaling.maxNodeCount else .initialNodeCount end) * (.config.machineType / "-" | .[-1] | tonumber)' )"
```
zerobfd, line 551: Please add some line breaks so this is easier to read.

zerobfd (changes requested):
Also we'll need to update the tests/fake_gcloud to account for this so the cli tests pass. Something like

```bash
 29   if [[ "${*}" == *"machineType"* ]]; then                                                                               
 30     echo "999"                                                                                                           
 31   else                                                                                                                   
 32     echo "this_should_pass"                                                                                              
 33   fi    
```

Otherwise it looks fine, I'll approve when the test is fixed.

zerobfd (approved):

PR #158, opened 2020-09-18
Report missing args with higher priority that wrong arg values

djazayeri:
I did this specifically because without this change if you type `./install_asm` with no args, it doesn't report usage, but just complains about needing mode to be install or migrate.


zerobfd (approved):

zerobfd:
Tests are failing because GKE picked today to update the version on the release track, we can merge anyways. Fix is in another PR.

PR #159, opened 2020-09-18
Apply canonical service yaml (unless explicitly disabled)

djazayeri:


djazayeri:
Do not merge until we get improved wording from the team that owns the canonical service component.

scripts/asm-installer/install_asm
```diff
@@ -867,6 +877,12 @@ install_asm(){
   sleep 1
   info "...done!"
 
+  if [[ "$DISABLE_CANONICAL_SERVICE" != 1 ]]; then
+    info "Installing ASM CanonicalService controller..."
+    retry 3 run kubectl apply -f asm/canonical-service/controller.yaml
```
jasonwzm, line 882: Add verification on the deployment? kubectl wait?

djazayeri: Good point, PTAL.

jasonwzm: Also can we add a warning or error when the wait fails. Similar to what istioctl does during install.

djazayeri: I agree that's nice-to-have, but we don't do that elsewhere in this script.

scripts/asm-installer/install_asm
```diff
@@ -867,6 +877,15 @@ install_asm(){
   sleep 1
   info "...done!"
 
+  if [[ "$DISABLE_CANONICAL_SERVICE" != 1 ]]; then
```
zerobfd, line 880: `if [[ "${DISABLE_CANONICAL_SERVICE}" -ne 1 ]]`

Very intuitively, != and = are string comparison operators in shell, you use -eq and -ne for numeric comparisons.

https://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-11.html

scripts/asm-installer/install_asm
```diff
@@ -867,6 +877,15 @@ install_asm(){
...
+    info "Installing ASM CanonicalService controller..."
```
zerobfd, line 881: It may be worth calling out here that it's going to the asm_system namespace. Someone might delete it if they don't recognize it otherwise.

zerobfd (commented):
lgtm after the comparison fix, will approve after the wording update as per your earlier comment

scripts/asm-installer/install_asm
djazayeri: Will add.

scripts/asm-installer/install_asm
djazayeri: Will change.

zerobfd (approved):

PR #164, opened 2020-09-23
Add environment variables to cloudbuild.yaml

zerobfd:
These are necessary for cloud build to run ci properly

### DIFF
--- a/scripts/asm-installer/tests/common.sh
+++ b/scripts/asm-installer/tests/common.sh
@@ -411,9 +411,9 @@ install_oss_istio() {
   TMPDIR="$(mktemp -d)"
   pushd "${TMPDIR}"
   curl -L "https://github.com/istio/istio/releases/download/\
-1.6.8/istio-1.6.8-linux-amd64.tar.gz" | tar xz
+1.7.1/istio-1.7.1-linux-amd64.tar.gz" | tar xz
 
-  ./istio-1.6.8/bin/istioctl operator init
+  ./istio-1.7.1/bin/istioctl operator init
   popd
   rm -r "${TMPDIR}"
 
--- a/scripts/asm-installer/tests/run_basic_suite
+++ b/scripts/asm-installer/tests/run_basic_suite
@@ -75,7 +75,6 @@ main() {
   echo "Performing a rolling restart of the demo app..."
   roll "${NAMESPACE}"
 
-
   local SUCCESS; SUCCESS=0;
   echo "Getting istio ingress IP..."
   GATEWAY="$(istio_ingress)"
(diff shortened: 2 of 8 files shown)


zerobfd:
Once CI was working, I added more commits to go ahead and make them green:
  * Fixed citadel installation to use the optional yaml
  * Updated the e2e tests to migrate from 1.7.1
  * Fixed the run_cli tests which weren't completely broken but weren't acting as expected
  * Added some more retries for the final step in the test

scripts/asm-installer/tests/common.sh
```diff
@@ -411,9 +411,9 @@ install_oss_istio() {
   TMPDIR="$(mktemp -d)"
   pushd "${TMPDIR}"
   curl -L "https://github.com/istio/istio/releases/download/\
-1.6.8/istio-1.6.8-linux-amd64.tar.gz" | tar xz
+1.7.1/istio-1.7.1-linux-amd64.tar.gz" | tar xz
```
djazayeri, line 414: Nit: extract a constant for this.

scripts/asm-installer/tests/run_basic_suite
```diff
@@ -85,6 +84,19 @@ main() {
   verify_demo_app "${GATEWAY}" || SUCCESS=1
   set -e
 
+  if [[ "${SUCCESS}" -eq 1 ]]; then
```
djazayeri, line 87: Nit: confusing that SUCCESS == 1 means failure

scripts/asm-installer/tests/run_basic_suite
```diff
@@ -85,6 +84,19 @@ main() {
...
+    echo "Failed to verify, restarting and trying again..."
```
djazayeri, line 88: Maybe comment some justification for why we try twice, versus considering this an error. (I assume flakiness, but this could cover a real issue.)

scripts/asm-installer/tests/run_migration_suite_citadel
```diff
@@ -119,6 +119,19 @@ main() {
   verify_demo_app "${GATEWAY}" || SUCCESS=1
   set -e
 
+  if [[ "${SUCCESS}" -eq 1 ]]; then
```
djazayeri, line 122: I'm curious if you could extract all this as a helper function instead of duplicating it in each test suite.

djazayeri (approved):

scripts/asm-installer/tests/run_migration_suite_citadel
zerobfd: I agree with all of these, but fixing these four wouldn't do much to solve the underlying issue that the tests are just really rough around the edges. I filed #165 as a followup to make the tests match the script style overall.

PR #166, opened 2020-09-24
Allow multiple operator overlays

zerobfd:
* Allow passing multiple -o options to specify multiple yaml files
 * Refactor citadel option to just add the file into the overlays instead of special casing it
 * Changed CLI tests to use meshca since validation changed location in the script

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -42,6 +42,7 @@ SCRIPT_NAME="${0##*/}"
 PROJECT_NUMBER=""
 KPT_URL=""
 KUBECONFIG=""
+ABS_OVERLAYS=""
 
 ### Option variables ###
 PROJECT_ID="${PROJECT_ID:=}"
@@ -193,7 +194,8 @@ OPTIONS:
   -o|--operator_overlay  <FILE PATH>  The location of a YAML file to overlay on
                                       the ASM IstioOperator. This option can be
                                       omitted if not installing optional
-                                      features.
+                                      features. To add multiple files, specify
+                                      them with multiple options one at a time.
   -s|--service_account   <ACCOUNT>    The name of a service account used to
                                       install ASM. If not specified, the gcloud
                                       user currently configured will be used.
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/install_asm
```diff
@@ -382,12 +384,19 @@ EOF
     fatal "Couldn't find key file ${KEY_FILE}."
   fi
 
-  if [[ -f "${OPERATOR_OVERLAY}" ]]; then
-    OPERATOR_OVERLAY="$(readlink -f "${OPERATOR_OVERLAY}")"
-    readonly OPERATOR_OVERLAY
-  elif [[ -n "${OPERATOR_OVERLAY}" ]]; then
-    fatal "Couldn't find yaml file ${OPERATOR_OVERLAY}."
+  while read -d ',' -r yaml_file; do
+    if [[ -f "${yaml_file}" ]]; then
```
jasonwzm, line 388: I guess here it should be `-e` to check if it is a file and then `readlink` will return the actual path?

jasonwzm (changes requested):

scripts/asm-installer/install_asm
zerobfd: -f also checks if it's a file. -e is any file-like object, -f is an fd that's not a device or directory.

jasonwzm: Ok, I thought -f does not check for symbolic link. Solved.

jasonwzm (approved):

scripts/asm-installer/install_asm
zerobfd: Just to double check, I ran these. It's true if the soft link points to a real file, and false if the link exists but points to a file that doesn't exist.

```bash
/tmp$ touch ./file
/tmp$ ln -s ./file one
/tmp$ ln -s nope two
/tmp$ [[ -f file ]]
/tmp$ echo $?
0
/tmp$ [[ -f one ]]
/tmp$ echo $?
0
/tmp$ [[ -f two ]]
/tmp$ echo $?
1
```

elfinhe (approved):

PR #167, opened 2020-09-24
Update ASM to 1.6.8

sandmman:


sandmman:
@elfinhe could you please review, thanks!

asm/Kptfile
```diff
@@ -70,7 +102,7 @@ openAPI:
           pattern: "proj-PROJECT_NUMBER"
           values:
           - marker: PROJECT_NUMBER
-            ref: '#/definitions/io.k8s.cli.setters.gcloud.project.projectNumber'
+            ref: '#/definitions/io.k8s.cli.setters.gcloud.project.environProjectNumber'
```
elfinhe, line 105: This line is only for multicluster setup in Environ. If you ASD do not use multicluster service mesh (cross cluster traffic). Please revert this line back.

Otherwise, you will have to add one step to set the environ project id: Step 5 in https://cloud.google.com/service-mesh/docs/gke-install-asm#preparing_resource_configuration_files.

sandmman: Ahh, that makes sense. Thank you and reverted.

elfinhe (approved):

sandmman:
@elfinhe thanks for the review, when you get a chance could you merge?

PR #171, opened 2020-09-24
Add --print-config flag

### LINKED ISSUE #169: Allow GitOps integration
Currently, the script generates and installs ASM config, which is not suitable for GitOps integration.  Ideally, users should treat their config as code, and review a diff of all changes before deploying to production via GitOps.  I think all that is needed here is a way to dump the config to stdout instead of applying it directly.

zerobfd:
Instead of installing anything, just print the config that would have been applied. This also skips verification and cluster/project setup. Closes #169.

Sample output (citadel was chosen to include an additional yaml file to overlay):

```
$>install_asm -l us-central1-c -n basic-suite-981hztoe -p test-cluster -m install -c citadel --print-config > testfile
$>cat testfile
---
kind: IstioOperator
spec:
  profile: asm-gcp
  hub: gcr.io/asm-staging-images/asm
  tag: 1.7.1-asm.0
  components:
    pilot:
      k8s:
        env:
        - name: SPIFFE_BUNDLE_ENDPOINTS
          value: "test-cluster.svc.id.goog|https://storage.googleapis.com/mesh-ca-resources/spiffe_bundle.json"
        - name: ENABLE_STACKDRIVER_MONITORING
          value: "true"
  meshConfig:
    defaultConfig:
      proxyMetadata:
        GCE_METADATA_HOST: "metadata.google.internal"
        TRUST_DOMAIN: "test-cluster.svc.id.goog"
  values:
    global:
      trustDomain: "test-cluster.svc.id.goog"
      sds:
        token:
          aud: "test-cluster.svc.id.goog"
---
kind: IstioOperator
spec:
  components:
    pilot:
      k8s:
        env:
        - name: SPIFFE_BUNDLE_ENDPOINTS
          value: ""
        - name: TOKEN_AUDIENCES
          value: "istio-ca,test-cluster.svc.id.goog"
        - name: ENABLE_STACKDRIVER_MONITORING
          value: "true"
  meshConfig:
    defaultConfig:
      proxyMetadata:
        TRUST_DOMAIN: "cluster.local"
        CA_PROVIDER: "Citadel"
        PLUGINS: ""
  values:
    global:
      trustDomain: "cluster.local"
      caAddress: ""
      pilotCertProvider: "istiod"
```

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -54,6 +54,7 @@ CA="${CA:=}"
 OPERATOR_OVERLAY="${OPERATOR_OVERLAY:=}"
 ENABLE_APIS="${ENABLE_APIS:=0}"
 DISABLE_CANONICAL_SERVICE="${DISABLE_CANONICAL_SERVICE:=0}"
+PRINT_CONFIG="${PRINT_CONFIG:=0}"
 SERVICE_ACCOUNT="${SERVICE_ACCOUNT:=}"
 KEY_FILE="${KEY_FILE:=}"
 OUTPUT_DIR="${OUTPUT_DIR:=}"
@@ -64,6 +65,11 @@ VERBOSE="${VERBOSE:=0}"
 
 main() {
   parse_args "${@}"
+  # make sure to redirect stdout as soon as possible if we're dumping the config
+  if [[ "${PRINT_CONFIG}" -eq 1 ]]; then
+    exec 3>&1
+    exec 1>&2
+  fi
   validate_args
   set_up_local_workspace
 
@@ -77,10 +83,15 @@ main() {
 
   set_up_project
   set_up_cluster
-  install_asm
-
-  info "Successfully installed ASM."
-  success
+  configure_package
+  if [[ "${PRINT_CONFIG}" -eq 1 ]]; then
+    print_config
+    success
+  else
+    install_asm
+    info "Successfully installed ASM."
+    success
+  fi
   return 0
 }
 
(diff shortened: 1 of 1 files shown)


backward-compatible (approved):
LGTM

elfinhe (approved):

scripts/asm-installer/install_asm
```diff
@@ -214,6 +225,11 @@ FLAGS:
   -e|--enable_apis                    Allow this script to enable necessary APIs
                                       on your behalf. Without this flag, it will
                                       abort if APIs are not already enabled.
+     --print_config                   Instead of installing ASM, print all of
+                                      the compiled YAML to stdout. All other
+                                      output will be written to stderr, even if
+                                      it would normally go to stdout. Skip all
```
djazayeri, line 231: Not sure I agree about "skip all validations" in this case. Is that standard/idiomatic for this kind of command? We can discuss offline.

zerobfd: I agree that it's a trade-off. In the context of the linked issue (GitOps / CI/CD integration) I thought it was safer for now to disable it by default. I don't know if the CI/CD user/service account is going to have the access to check everything, if the cluster names may be templated in later, etc. @therealmitchconnors do you have any guidance one way or the other?

therealmitchconnors: Is this validation against user-supplied config, or against runtime values in the cluster or project?

zerobfd: Runtime values. The CLI options/dependencies are still verified.

therealmitchconnors: Then it does seem like a runtime validation might not be possible in some CI scenarios.  Perhaps we should have a separate flag to control this?

zerobfd: Sure. Added #176 to track the extra flag.

PR #173, opened 2020-09-25
Fail earlier if kubectl can't connect

zerobfd:
Also fix a bug where an error would fail to print if a machine type that didn't end in a number was in a node pool.


backward-compatible (approved):
LGTM

djazayeri (approved):

elfinhe (approved):

PR #177, opened 2020-09-29
Profile multiproject

yanyuan-meiyi:


elfinhe (commented):
can we do the same change in pkg/ and use make to sync them into asm/ and asm/patch?

yanyuan-meiyi:
> can we do the same change in pkg/ and use make to sync them into asm/ and asm/patch?

elfinhe (approved):

PR #178, opened 2020-09-29
Fix node pool validation

zerobfd:
* Bump image version for staging tests
 * The validation fails if there is more than one node pool, this fixes it so that if the max CPU pool is valid then we pass


backward-compatible (approved):
LGTM

PR #190, opened 2020-10-06
Add support for upgrading ASM to the install scripts

backward-compatible:


scripts/asm-installer/install_asm
```diff
@@ -661,10 +668,9 @@ EOF
 validate_expected_control_plane(){
   info "Checking Istio installations..."
   check_no_istiod_outside_of_istio_system_namespace
-  if [[ "${MODE}" = "migrate" ]]; then
+  if [[ "${MODE}" = "migrate" || "${MODE}" = "upgrade" ]]; then
```
zerobfd, line 671: This can be a bit more succinct.

`if [[ "${MODE}" = @(migrate|upgrade) ]]; then` should work w/pattern matching.

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
   fi
 }
 
+validate_asm_version() {
+  info "Checking existing ASM version(s)..."
+  local OUTPUT
+  OUTPUT="$(retry 3 ./istio-${RELEASE}/bin/istioctl version -o json)"
+  local FOUND_INVALID_VERSION
```
zerobfd, line 740: I think the style consistent with the rest of the script is `local X; X=something;` for shorter lines.

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
...
+  FOUND_INVALID_VERSION="false"
```
zerobfd, line 741: Other places use 0 for false and 1 for true

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
   fi
 }
 
+validate_asm_version() {
+  info "Checking existing ASM version(s)..."
+  local OUTPUT
+  OUTPUT="$(retry 3 ./istio-${RELEASE}/bin/istioctl version -o json)"
```
zerobfd, line 739: Can you use ./"${ISTIOCTL_REL_PATH}" instead of istio-whatever? I think it should be set by this point in the script.

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
   fi
 }
 
+validate_asm_version() {
+  info "Checking existing ASM version(s)..."
+  local OUTPUT
```
zerobfd, line 738: VERSIONS or INSTALLED_VERSIONS or something is more descriptive

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
...
+
+    # if asm version found, pattern: 1.6.11-asm.1-586f900508ad482ed32b830dd15f6c54b32b93ed
+    local VERSION_MAJOR VERSION_MINOR VERSION_POINT VERSION_REV
+    IFS="." read VERSION_MAJOR VERSION_MINOR VERSION_POINT VERSION_REV <<< "$version"
+    VERSION_POINT="$(echo "$VERSION_POINT" | sed 's/-.*//')"
+    VERSION_REV="$(echo "$VERSION_REV" | sed 's/-.*//')"
+    if [[ "$VERSION_MAJOR" -ne 1 || "$VERSION_MINOR" -lt 6 || "$VERSION_MINOR" -ge "$MINOR" || \
+      ( "$VERSION_MINOR" -eq "$MINOR" && \
+        ( "$VERSION_POINT" -ge "$POINT" || \
+          ("$VERSION_POINT" -eq "$POINT" && "$VERSION_REV" -ge "$REV"))) ]]; then
```
zerobfd, line 755: This seems like a good candidate to pull into a different function.

```
version_less_than_installed() {
  VERSION=$1
  ....
  return 0/1
}
```

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
...
+  done
+  if [[ "$FOUND_INVALID_VERSION" = "true" ]]; then
+    fatal "Upgrade requires all existing control planes to be between versions 1.6.0-asm.0 (inclusive) and ${RELEASE} (exclusive)."
+  fi
+}
+
+validate_ca_consistency() {
+  info "Checking existing CA..."
+  local INSTALLED_CA
+  INSTALLED_CA="$(kubectl -n istio-system get pod -l istio=ingressgateway -o jsonpath='{.items[].spec.containers[].env[?(@.name=="CA_ADDR")].value}')"
```
zerobfd, line 770: Can you toss some line breaks in here?

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
...
+  if [[ ("${CA}" = "citadel" && ! "${INSTALLED_CA}" =~ istiod.*istio\-system\.svc) || \
```
zerobfd, line 771: Not sure if we need to handle it this in this commit, but do we know this will match even w/custom certs in Citadel?

scripts/asm-installer/install_asm
```diff
@@ -728,6 +733,47 @@ validate_istio_version() {
...
+        ("${CA}" = "mesh_ca" && ! "${INSTALLED_CA}" =~ meshca\.googleapis\.com) ]]; then
+    fatal "CA cannot be switched while performing upgrade. Please use the same CA."
```
zerobfd, line 773: We know the CA, instead of "same CA" we should use the value.

zerobfd (changes requested):
Logic looks perfect! Some style issues then should be gtg.

scripts/asm-installer/install_asm
backward-compatible: Currently, it's mixed style. I was following the existing **validate_istio_version** as a reference but I will updated both to match this.

scripts/asm-installer/install_asm
backward-compatible: SG

scripts/asm-installer/install_asm
backward-compatible: Currently, it's mixed style. I was following the existing **validate_istio_version** as a reference but I will updated both to match this.

scripts/asm-installer/install_asm
backward-compatible: SG

scripts/asm-installer/install_asm
backward-compatible: SG

scripts/asm-installer/install_asm
backward-compatible: I think we cannot yet support customer CA with scripts due to additional complexities so this will just fail fast right?

scripts/asm-installer/install_asm
backward-compatible: I like the suggestion, thanks!

scripts/asm-installer/install_asm
zerobfd: Fair: I think the rule of thumb is that if you can do it and stay within 80 char width then do it,  otherwise split them. If you think it's more readable like this then feel free to keep it.

scripts/asm-installer/install_asm
zerobfd: I think we should switch this. I much prefer using -eq and -ne to compare things rather than strings in bash.

scripts/asm-installer/install_asm
zerobfd: Someone might have installed manually and are trying to use the script to upgrade.

backward-compatible: updating this to pivot on meshCA check only

backward-compatible:
```
if [[ ("${INSTALLED_CA}" =~ meshca\.googleapis\.com && ! "${CA}" = "mesh_ca") || \
        ( ! "${INSTALLED_CA}" =~ meshca\.googleapis\.com && ! "${CA}" = "citadel") ]]; then
    fatal "CA cannot be switched while performing upgrade. Please use ${CA}."
fi
```

is this what you were thinking?

scripts/asm-installer/install_asm
backward-compatible: posh compliance fails with:

`if [[ "${MODE}" = @(migrate|upgrade) ]]; then` any thoughts?

zerobfd: Nope, posh is right, that's a bashism. Your original way is better.

scripts/asm-installer/install_asm
zerobfd: Maybe a bit easier to read:

```
is_meshca() {
  return [[ "${INSTALLED_CA}" =~ meshca\.googleapis\.com ]]
}

if is_meshca && "${CA}" == "citadel" || ! is_meshca && "${CA}" == "mesh_ca"; then
```

backward-compatible: agreed. I wonder if we need the condition to be grouped otherwise `is_meshca && ... || ! is_meshca && ..` will always be true? I'm not sure what's the posh way to group conditions..

backward-compatible: using `[[ ]]` for grouping

scripts/asm-installer/install_asm
backward-compatible: SG

scripts/asm-installer/install_asm
```diff
@@ -707,27 +713,108 @@ check_istio_not_deployed(){
 
 validate_istio_version() {
   info "Checking existing Istio version(s)..."
-  local OUTPUT
-  OUTPUT="$(retry 3 ./istio-${RELEASE}/bin/istioctl version -o json)"
-  local FOUND_VALID_VERSION
-  FOUND_VALID_VERSION="false"
-  for version in $(echo "${OUTPUT}" | jq -r '.meshVersion[].Info.version' -r); do
+  local VERSION_OUTPUT; VERSION_OUTPUT="$(retry 3 ./${ISTIOCTL_REL_PATH} version -o json)"
```
zerobfd, line 716: I missed this, but we should check that VERSION_OUTPUT isn't blank because istioctl messed up somehow. Just a quick

```
if [[ -z "${VERSION_OUTPUT}" ]]; then
  fatal "Couldn't validate existing Istio versions."
fi
```

scripts/asm-installer/install_asm
```diff
@@ -707,27 +713,108 @@ check_istio_not_deployed(){
...
   done
-  if [[ "$FOUND_VALID_VERSION" != "true" ]]; then
+  if [[ "$FOUND_VALID_VERSION" -eq 0 ]]; then
     fatal "Migration requires an existing control plane in the ${RELEASE_LINE} line."
   fi
 }
 
+validate_asm_version() {
+  info "Checking existing ASM version(s)..."
+  local VERSION_OUTPUT; VERSION_OUTPUT="$(retry 3 ./${ISTIOCTL_REL_PATH} version -o json)"
```
zerobfd, line 736: Same as above

```
if [[ -z "${VERSION_OUTPUT}" ]]; then
  fatal "Couldn't validate existing Istio versions."
fi
```

scripts/asm-installer/install_asm
```diff
@@ -707,27 +713,108 @@ check_istio_not_deployed(){
...
+  local FOUND_INVALID_VERSION; FOUND_INVALID_VERSION=0
+  for VERSION in $(echo "${VERSION_OUTPUT}" | jq -r '.meshVersion[].Info.version'); do
+    if ! [[ "${VERSION}" =~ "asm" ]]; then
+      fatal "Cannot upgrade from version ${VERSION}. Only upgrade from ASM distribution is supported."
```
zerobfd, line 740: nit: Only upgrades from ASM distributions are supported.

scripts/asm-installer/install_asm
```diff
@@ -707,27 +713,108 @@ check_istio_not_deployed(){
...
+
+    if version_valid_for_upgrade "${VERSION}"; then
+      info "  ${VERSION} (suitable for migration)"
+    else
+      info "  ${VERSION} (not suitable for migration)"
+      FOUND_INVALID_VERSION=1
+    fi
+  done
+  if [[ "$FOUND_INVALID_VERSION" -eq 1 ]]; then
+    fatal "Upgrade requires all existing control planes to be between versions 1.6.0-asm.0 (inclusive) and ${RELEASE} (exclusive)."
```
zerobfd, line 751: Can you change 1.6.0 to 1.$((MINOR-1)).0?

scripts/asm-installer/install_asm
```diff
@@ -707,27 +713,108 @@ check_istio_not_deployed(){
...
+  if is_major_minor_invalid || is_minor_point_rev_invalid; then
+    return 1
+  else
+    return 0
+  fi
+}
+
+is_major_minor_invalid() {
+  [[ "$VERSION_MAJOR" -ne 1 ]] && return 0
+  [[ "$VERSION_MINOR" -lt 6  ]] && return 0
```
zerobfd, line 780: 6 -> $((MINOR-1))

zerobfd (approved):
LGTM after a few more nits. Thanks!

scripts/asm-installer/install_asm
backward-compatible: SG

scripts/asm-installer/install_asm
backward-compatible: sg

scripts/asm-installer/install_asm
backward-compatible: sg

scripts/asm-installer/install_asm
backward-compatible: sg

scripts/asm-installer/install_asm
```diff
@@ -707,27 +713,121 @@ check_istio_not_deployed(){
...
+)"
+  if is_major_minor_invalid || is_minor_point_rev_invalid; then
+    return 1
+  else
+    return 0
+  fi
+}
+
+is_major_minor_invalid() {
+  [[ "$VERSION_MAJOR" -ne 1 ]] && return 0
```
zerobfd, line 785: It looks like the VERSION_* variables were marked local, does this work?

backward-compatible: It works with bash due to dynamic scoping.

PR #191, opened 2020-10-07
Aggregate vCPU count for all nodepools

### LINKED ISSUE #66: validate-asm errors and multiple nodepools
Hi,

I have a regional GKE cluster with multiple nodepools and while validating asm install using `kpt fn source ${BASE_DIR} | kpt fn run --image gcr.io/kustomize-functions/validate-asm:v0.1.0` produces the following errors/warnings:

```
2 warning(s) occurred:
        * Warning - spec.nodeCount is 2 in ContainerNodePool asm-node-pool (all.yaml [8]). ASM requires at least four nodes. If you need to add nodes, see https://bit.ly/2RnVL2T
        * Warning - spec.initialNodeCount missing in ContainerNodePool default-pool (all.yaml [9])

2 error(s) occurred:
        * Error - unsupported machine type: n1-standard-1 in ContainerCluster foo-cluster (all.yaml [7]). The minimum machine type is n1-standard-4, which has four vCPUs. If the machine type for your cluster doesn't have at least four vCPUs, change the machine type as described here https://bit.ly/2V0KPdu
        * Error - unsupported machine type: n1-standard-1 in ContainerNodePool default-pool (all.yaml [9]). The minimum machine type is n1-standard-4, which has four vCPUs. If the machine type for your cluster doesn't have at least four vCPUs, change the machine type as described here https://bit.ly/2V0KPdu
```

My `default-pool` has 0 nodes and `asm-node-pool` has 6 (2 per zone) of type `e2-standard-4`. Based on the documentation my understanding was that the requirement was 

- At least four nodes.
- The minimum machine type is e2-standard-4

Any insight into these errors or if I am missing something would be great!

nan-yu:
The current validate function checks each individual nodepool separately. We need to update the function to check the aggregated vcpu counts from multiple nodepools.

It also downgrades the machineType check failure from Error to Warning.

Fixes #66.

### DIFF
--- a/kustomize-functions/validate-asm/image/main.go
+++ b/kustomize-functions/validate-asm/image/main.go
@@ -34,6 +34,8 @@ const (
 	containerClusterKind   = "ContainerCluster"
 	containerNodePoolKind  = "ContainerNodePool"
 	apiGroup               = "container.cnrm.cloud.google.com"
+	minimumTotalVCPUs      = 8
+	minimumVCPUsPerNode    = 4
 )
 
 var supportedReleaseChannels = []string{"REGULAR", "RAPID", "STABLE"}
@@ -115,9 +117,10 @@ func (e ValidationErrors) Error() string {
 func (filter) Filter(in []*yaml.RNode) ([]*yaml.RNode, error) {
 	var errList []error
 	var warningList []error
+	var totalVCPUCount int
 
 	for _, r := range in {
-		if errs := validate(r); len(errs) != 0 {
+		if errs := validate(r, &totalVCPUCount); len(errs) != 0 {
 			for _, e := range errs {
 				if e.severity == Warning {
 					warningList = append(warningList, e)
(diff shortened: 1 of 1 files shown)


nan-yu:
FYI: @janetkuo

kustomize-functions/validate-asm/image/main.go
```diff
@@ -34,6 +34,8 @@ const (
 	containerClusterKind   = "ContainerCluster"
 	containerNodePoolKind  = "ContainerNodePool"
 	apiGroup               = "container.cnrm.cloud.google.com"
+	minimalTotalVCPUs      = 8
```
zerobfd, line 37: nit: this is normally "minimum", not "minimal"

kustomize-functions/validate-asm/image/main.go
```diff
@@ -253,23 +261,35 @@ func contains(s []string, e string) bool {
...
+	}
+	return nil
+}
 
-	} else if err != nil {
-		return err
+func validateMinimalVCPUCount(count int) error {
+	if count < minimalTotalVCPUs {
+		return fmt.Errorf("the aggregated vCPU count is %d. " +
+				"Anthos Service Mesh recommends at least %d vCPUs. If you need to add nodes, " +
```
zerobfd, line 280: We block on this in the installation script: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm#L633

I thought this was a requirement, not a recommendation. Did this change recently?

zerobfd (commented):
This seems generally useful, why is it just going into the release-1.5 branch?

kustomize-functions/validate-asm/image/main.go
nan-yu: Thanks, done!

nan-yu:
> This seems generally useful, why is it just going into the release-1.5 branch?

`release-1.5-asm` was the first branch that introduced the validation function. I was planning to cherry-pick into the other branches after this PR is merged.

zerobfd:
I think merging into master and cherry-picking onto the release branches makes more sense. Practically it's the same, but in general doing development directly on release branches isn't how most repos are organized.

nan-yu:
> I think merging into master and cherry-picking onto the release branches makes more sense. Practically it's the same, but in general doing development directly on release branches isn't how most repos are organized.

Sounds good to me. I will defer to Tao to merge this function into the master branch.

janetkuo:
IIUC GitHub doesn't allow changing target branch in a PR. For merging changes to the `master` branch you'd need to create a new PR. @nan-yu would you open a new PR targeting `master`?

nan-yu:
The PR should be created against the master branch. This one is superseded by #192.

PR #192, opened 2020-10-07
Aggregate vCPU count for all nodepools

### LINKED ISSUE #66: validate-asm errors and multiple nodepools

nan-yu:
The current validate function checks each individual nodepool separately. We need to update the function to check the aggregated vcpu counts from multiple nodepools.

It also downgrades the machineType check failure from Error to Warning.

Fixes #66.

### DIFF
--- a/kustomize-functions/validate-asm/image/main.go
+++ b/kustomize-functions/validate-asm/image/main.go
@@ -34,6 +34,8 @@ const (
 	containerClusterKind   = "ContainerCluster"
 	containerNodePoolKind  = "ContainerNodePool"
 	apiGroup               = "container.cnrm.cloud.google.com"
+	minimumTotalVCPUs      = 8
+	minimumVCPUsPerNode    = 4
 )
 
 var supportedReleaseChannels = []string{"REGULAR", "RAPID", "STABLE"}
@@ -115,9 +117,10 @@ func (e ValidationErrors) Error() string {
 func (filter) Filter(in []*yaml.RNode) ([]*yaml.RNode, error) {
 	var errList []error
 	var warningList []error
+	var totalVCPUCount int
 
 	for _, r := range in {
-		if errs := validate(r); len(errs) != 0 {
+		if errs := validate(r, &totalVCPUCount); len(errs) != 0 {
 			for _, e := range errs {
 				if e.severity == Warning {
 					warningList = append(warningList, e)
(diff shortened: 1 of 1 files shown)


nan-yu:
FYI: @janetkuo

This PR supersedes #191.

kustomize-functions/validate-asm/image/main.go
```diff
@@ -127,14 +130,17 @@ func (filter) Filter(in []*yaml.RNode) ([]*yaml.RNode, error) {
 			}
 		}
 	}
+	if err := validateMinimumVCPUCount(totalVCPUCount); err != nil {
+		warningList = append(warningList, ValidationError{Warning, err})
```
elfinhe, line 134: The ASM teams also have a couple discussions for a similar question - "should the auto installation generate errors on the vCPU requirements or should it just do a warning and continue?" 

Our answer for https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/tree/master/scripts/asm-installer is - the install should give errors and helpful messages and stop there.

So to keep consistent behavior, I'd suggest we change kpt validate function to use errors instead of warnings.

nan-yu: Okay. What about machine type check? From discussion in #66, we all agreed to return a warning for the node pool with unsatisfied machine type. But it seems to be a hard requirement from the [documentation](https://cloud.google.com/service-mesh/docs/gke-install-overview#requirements)
> Your GKE cluster must meet the following requirements: A machine type that has at least four vCPUs, such as e2-standard-4. If the machine type for your cluster doesn't have at least four vCPUs, change the machine type as described in Migrating workloads to different machine types.

zerobfd: From memory, the last conversation we had was something like this: machine type and CPU count should be errors for now since that's what we document, and we don't want to validate something we don't support. Pretty soon the requirement will go way down anyways, so this won't be an issue for long.

nan-yu: Fair enough. The validation was updated to return errors for both machine type and CPU count. PTAL.

elfinhe (approved):

zerobfd (approved):

PR #198, opened 2020-10-08
Add configuration for GCLB integration (#196)

zerobfd:
Cherry pick of ce6bd8a with the file moved to the pre-1.7 standard

### DIFF
--- a/asm/cluster/iap-operator.yaml
+++ b/asm/cluster/iap-operator.yaml
@@ -1,0 +1,44 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


elfinhe (changes requested):
this will break anthoscli workflow which is still available in ASM 1.6.

elfinhe:
> this will break anthoscli workflow which is still available in ASM 1.6.

we can copy asm/ folder into asm-iap/ and update asm-iap/cluster/istio-operator.yaml with the ingress gateway change.

Then users can use 

```
kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/asm-iap@release-1.6-asm asm
```

to get the iap compatible config.

PR #201, opened 2020-10-09
Add configuration for GCLB integration (#196)

zerobfd:
Cherry pick of ce6bd8a with the file moved to the pre-1.7 standard


asm-iap/cluster/iap-operator.yaml
```diff
@@ -0,0 +1,44 @@
...
+              name: http2
+            - port: 443
+              targetPort: 8443
+              name: https
+            - port: 15012
+              targetPort: 15012
+              name: tcp-istiod
+            - port: 15443
+              targetPort: 15443
+              name: tls
```
elfinhe, line 44: insert into istio-operator.yaml:26 should do the job.

zerobfd: Makes sense

elfinhe (approved):

PR #202, opened 2020-10-10
Support custom certificates for Citadel

zerobfd:
* Add 4 long options to install_asm for custom certificate files
 * Validates openssl CLI dependency if installing a custom cert
 * Validate files exist, and passing any option requires passing all of them
 * Validate that cusom ca arguments only work with citadel
 * Validate formats of root/CA/key
 * Validate key matches CA cert 
 * Validate that chain of trust is valid for root -> CA
 * Creates the cacerts k8s secret from given files
 * Added e2e test that generates certs, installs with them, them verifies the ones from istio-proxy are the same as the ones generated
 * Added e2e test to cloudbuild

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -60,6 +60,12 @@ SERVICE_ACCOUNT="${SERVICE_ACCOUNT:=}"
 KEY_FILE="${KEY_FILE:=}"
 OUTPUT_DIR="${OUTPUT_DIR:=}"
 
+CUSTOM_CA=0
+CA_CERT="${CA_CERT:=}"
+CA_KEY="${CA_KEY:=}"
+CA_ROOT="${CA_ROOT:=}"
+CA_CHAIN="${CA_CHAIN:=}"
+
 DRY_RUN="${DRY_RUN:=0}"
 ONLY_VALIDATE="${ONLY_VALIDATE:=0}"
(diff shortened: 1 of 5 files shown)


zerobfd:
Tested manually and CI test should pass now.

backward-compatible (commented):
LGTM

scripts/asm-installer/install_asm
```diff
@@ -472,6 +523,60 @@ EOF
...
+  CA_ROOT="$(readlink -f "${CA_ROOT}")"; readonly CA_ROOT;
+
+  info "Checking certificate files for consistency..."
+  if ! openssl rsa -in "${CA_KEY}" -check >/dev/null 2>/dev/null; then
+    fatal "${CA_KEY} failed an openssl consistency check."
+  fi
+  if ! openssl x509 -in "${CA_CERT}" -text -noout >/dev/null; then
+    fatal "${CA_CERT} failed an openssl consistency check."
+  fi
+  if ! openssl x509 -in "${CA_CHAIN}" -text -noout >/dev/null; then
```
myidpt, line 556: I think openssl x509 only checks a single one cert in the PEM file. But it's OK here because you have the cert chain verification later.

myidpt (commented):
lgtm

backward-compatible (approved):

PR #203, opened 2020-10-11
Support k8s 1.15

zerobfd:


scripts/asm-installer/tests/fake_kubectl
```diff
@@ -2,6 +2,35 @@
...
+    "gitCommit": "f5743093fd1c663cb0cbc89748f730662345d44d",
+    "gitTreeState": "clean",
+    "buildDate": "2020-09-16T13:41:02Z",
+    "goVersion": "go1.15",
+    "compiler": "gc",
+    "platform": "linux/amd64"
+  },
+  "serverVersion": {
+    "major": "1",
+    "minor": "16+",
```
elfinhe, line 21: Is this test just for k8s 1.16?

elfinhe (approved):

scripts/asm-installer/tests/fake_kubectl
zerobfd: Yea we'll probably have to update it next version. If you think there's a better way then lmk and we can open an issue or something.

PR #222, opened 2020-10-21
allow explicitly overriding http and status nodeports

therealmitchconnors:
This will allow users to upgrade with continuity.

Note: this doesn't currently work and I don't know why.  Running any kpt command results in the following:
`error: json: cannot unmarshal number into Go struct field setter.setter.value of type string`

### DIFF
--- a/asm-iap/Kptfile
+++ b/asm-iap/Kptfile
@@ -50,6 +50,20 @@ openAPI:
         setter:
           name: anthos.servicemesh.trustDomainAliases
           listValues:
+    io.k8s.cli.setters.gcloud.container.cluster.ingress.httpPort:
+      type: integer
+      x-k8s-cli:
+        setter:
+          name: gcloud.container.cluster.ingress.httpPort
+          value: 31224
+          isSet: true
+    io.k8s.cli.setters.gcloud.container.cluster.ingress.statusPort:
+      type: integer
+      x-k8s-cli:
+        setter:
+          name: gcloud.container.cluster.ingress.statusPort
+          value: 31223
+          isSet: true
     io.k8s.cli.setters.gcloud.container.cluster:
       type: string
       x-k8s-cli:
--- a/asm-iap/Kptfile
+++ b/asm-iap/Kptfile
@@ -55,14 +55,14 @@ openAPI:
       x-k8s-cli:
         setter:
           name: gcloud.container.cluster.ingress.httpPort
-          value: 31224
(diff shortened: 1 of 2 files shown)


asm-iap/Kptfile
```diff
@@ -50,6 +50,20 @@ openAPI:
         setter:
           name: anthos.servicemesh.trustDomainAliases
           listValues:
+    io.k8s.cli.setters.gcloud.container.cluster.ingress.httpPort:
+      type: integer
+      x-k8s-cli:
+        setter:
+          name: gcloud.container.cluster.ingress.httpPort
+          value: 31224
```
zerobfd, line 58: Try putting 31224 and 31223 in quotes.

therealmitchconnors: thx, that did the trick.  apparently ints are bad.

zerobfd: I think that very unintuitively, the values that kpt uses are all strings regardless of what type you set. Using integer is referring to the value of the yaml, not in the text file.

asm-iap/Kptfile
```diff
@@ -50,6 +50,20 @@ openAPI:
...
+          name: gcloud.container.cluster.ingress.httpPort
+          value: "31224"
+          isSet: true
+    io.k8s.cli.setters.gcloud.container.cluster.ingress.statusPort:
+      type: integer
+      x-k8s-cli:
+        setter:
+          name: gcloud.container.cluster.ingress.statusPort
+          value: "31223"
+          isSet: true
```
zerobfd, line 66: This is like the nittiest nit ever, but can you be consistent with the order of these? In istio-operator the statusPort comes first.

zerobfd (approved):
I trust that you know how to test this better than I can, but the code looks good.

asm-iap/Kptfile
therealmitchconnors, line 66:
```suggestion
    io.k8s.cli.setters.gcloud.container.cluster.ingress.statusPort:
      type: integer
      x-k8s-cli:
        setter:
          name: gcloud.container.cluster.ingress.statusPort
          value: "31223"
          isSet: true
    io.k8s.cli.setters.gcloud.container.cluster.ingress.httpPort:
      type: integer
      x-k8s-cli:
        setter:
          name: gcloud.container.cluster.ingress.httpPort
          value: "31224"
          isSet: true
```

PR #229, opened 2020-10-23
Allow remote overlays, add multicloud yaml

zerobfd:


scripts/asm-installer/install_asm
```diff
@@ -261,6 +263,13 @@ OPTIONS:
                                       omitted if not installing optional
                                       features. To add multiple files, specify
                                       them with multiple options one at a time.
+  -r|--remote_overlay    <FILE PATH>  The location of a YAML file in
```
jasonwzm, line 266: Based on what you are doing below, it is technically the file name of the overlay.

scripts/asm-installer/install_asm
```diff
@@ -210,6 +211,7 @@ OPTIONS:
   -c|--ca                <CA>
 
   -o|--operator_overlay  <FILE PATH>
+  -r|--remote_overlay    <FILE PATH>
```
jasonwzm, line 214: Can we actually call this asm_overlay? After seeing the name, I first thought it could accept a URL to an overlay file.

jasonwzm (changes requested):

scripts/asm-installer/install_asm
zerobfd: True, good point.

scripts/asm-installer/install_asm
zerobfd: Not sure asm is the right choice, but remote is confusing.

scripts/asm-installer/install_asm
```diff
@@ -718,6 +733,17 @@ EOF
     info "Downloading ASM kpt package..."
     retry 3 run kpt pkg get "${KPT_URL}" asm
   fi
+
+  while read -d ',' -r yaml_file; do
+    yaml_file="${OPTIONS_DIRECTORY}/${yaml_file}.yaml"
+    if [[ ! -f "${yaml_file}" ]]; then
+      fatal "Couldn't find remote yaml file ${yaml_file}."
```
jasonwzm, line 740: Maybe list the options users can choose here?

jasonwzm (changes requested):
Otherwise, LGTM.

elfinhe (approved):

scripts/asm-installer/install_asm
zerobfd: Good idea.

PR #231, opened 2020-10-26
Add istiod service to fix validation

zerobfd:


asm/istio/istiod-service.yaml
```diff
@@ -0,0 +1,45 @@
...
+# This YAML is intended to create an istiod service for validation to
+# work around istio/istio/issues/27501 in the case of no non-revision
+# installation.
+apiVersion: v1
+kind: Service
+metadata:
+  name: istiod
+  namespace: istio-system
+  labels:
+    istio.io/rev: asm-173-3 # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev"}
```
elfinhe, line 24: Any reason why we removed the operator labels?
```
    operator.istio.io/component: Pilot
    operator.istio.io/managed: Reconcile
    operator.istio.io/version: 1.7.3-asm.6
```

btw, operator labels now are used for pruning resources not used in a new version, which i think it is the right to not adding them here.

elfinhe (approved):

asm/Kptfile
```diff
@@ -99,6 +99,11 @@ openAPI:
         setter:
           name: anthos.servicemesh.tag
           value: 1.6.8-asm.9
+    io.k8s.cli.setters.anthos.servicemesh.rev:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.rev
+          value: asm-173-3
```
howardjohn, line 106: is value a default? if so, shouldn't it match `tag` if we are going to have version in it?

howardjohn (approved):
How does a user move `istiod` to the new revision on upgrade?

asm/Kptfile
zerobfd: Good point.

zerobfd:
I'm going to split this into 2 PRs, since otherwise tests will fail.

PR #249, opened 2020-11-03
Remove asm-patch used by anthoscli workflow; remove pkg used for sync…

elfinhe:
- Remove asm-patch used by anthoscli workflow; 
- Remove pkg used for syncing common configs;
- Bump asm tag, canonical service tag, rev defaults.


zerobfd (approved):
LGTM. I'd double check that kpt doesn't throw a syntax error somewhere just in case. In fact I think I'll add an issue to add a linter for that.

PR #253, opened 2020-11-09
README fix, add document explaining version tags

### LINKED ISSUE #170: Track releases with tags
In order for installs to be reproducible, we should use tags to represent patch releases in each minor branch.  Docs will also need to be updated to point to the tags, rather than the branch.

zerobfd:
Closes #170 

Also remove some README.md that isn't relevant anymore


backward-compatible (approved):

docs/versioning.md
```diff
@@ -0,0 +1,114 @@
...
+## Concepts
+
+* **Release branch**: a branch corresponding to a specific minor version of ASM
+that has the latest stable configuration for the latest minor release
+* **Staging branch**: a branch corresponding to a specific minor version of ASM
+that allows testing all of the configuration as one unit before merging into
+the release branch
+* **Tag**: an immutable pointer to a specific commit
+* **Major, minor, point**: their normal meaning in [SemVer](https://semver.org/)
+* **Revision**: on occasion, ASM will release an important fix without an
```
elfinhe, line 40: Will this mix up the concepts between `istioctl install --revision` and the image tag, and cause confusion? The part after the hyphen are called 'pre-release' https://semver.org/#spec-item-9 in versem.

But if we are sure the revision here is referring the same idea with what we append to `istioctl install --revision`, then the doc looks good to me.

elfinhe (approved):

docs/versioning.md
```diff
@@ -0,0 +1,114 @@
...
+* The administrator uses a signed, annotated git tag to tag the commit
+* (Alternatively they can use GitHub's feature for verified signatures)
+* If this is a new ASM release, the tag changes to that release and N resets to
+1
+* Otherwise, the same tag as previous release is used except increment N by one
+* The administrator pushes the tag to GitHub
+
+## Diagram
+
+![Diagram of the process](./diagram.png?raw=true)
```
elfinhe, line 103: minor: maybe we can update the right side of this diagram to be something more like a git tree.

docs/versioning.md
zerobfd: I agree, but I don't think there's a better way to handle it. Anything after the hyphen is prerelease, but the number after "asm." specifically is a revision number. That word has been used since long before Istio was a thing: see https://docs.microsoft.com/en-us/dotnet/api/system.version?redirectedfrom=MSDN&view=netcore-3.1#remarks for an example.

I'm up for suggestions, but I think we should wait for feedback that it's an actual problem before spending time on it.

docs/versioning.md
zerobfd: If someone better than me at diagrams wants to attempt that then I'd approve it immediately.

PR #264, opened 2020-11-16
Allow granular selection of permissions

### LINKED ISSUE #207: Don't bind IAM permissions by default
Right now the install script will always bind the current user/service user to the required IAM permissions. We shouldn't do that by default, and instead do the same thing we're doing with APIs--add a flag to enable it. There's been a lot of feedback about this from various customers.

zerobfd:
Previously the only action that could be turned off is enabling the
GCP APIs, and all of the other modifications were done all the time.
This commit adds several flags to control different modifications, and
adds an "--enable_all" flag that will allow users to keep the old
behavior. It also changes -e to --enable_all from --enable_apis.

Fixes #207

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -102,7 +102,7 @@ main() {
     auth_service_account
   fi
 
-  if should_validate || can_modify; then
+  if should_validate || can_modify_at_all; then
     validate_gcp_resources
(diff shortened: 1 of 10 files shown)


zerobfd:
@zabela This is the PR that splits up the permissions

elfinhe (approved):

PR #273, opened 2020-11-19
Add installation with Hub IDNS option support in install_asm script

jtrbs:


scripts/asm-installer/install_asm
```diff
@@ -84,6 +84,12 @@ ONLY_VALIDATE="${ONLY_VALIDATE:=0}"
 VERBOSE="${VERBOSE:=0}"
 PRINT_HELP=0
 
+# Environment variables used by Hub IDNS
+ENABLE_HUB_IDNS="${ENABLE_HUB_IDNS:=0}"
+ENVIRON_PROJECT_ID="${ENVIRON_PROJECT_ID:=}"
+HUB_MEMBERSHIP_ID="${HUB_MEMBERSHIP_ID:=}"
```
zerobfd, line 90: The := means that users could set these environment variables before running the script and they would be honored in the script. Looking below, you unconditionally overwrite these variables, so I think the intention is to not let users set them. In that case, I would initialize these just to 0 and "".

scripts/asm-installer/install_asm
```diff
@@ -692,6 +710,10 @@ EOF
     *) fatal "MODE must be one of 'install', 'migrate', 'upgrade'";;
   esac
 
+  if [[ "${MODE}" -ne "install" && "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
+    fatal "Hub is only supported in new installation"
```
zerobfd, line 710: Please specify hub idns, not just hub

scripts/asm-installer/install_asm
```diff
@@ -151,6 +157,15 @@ main() {
...
 
+  if [[ "${CA}" = "mesh_ca" ]] && can_modify_at_all; then
+    if [[ "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
+      get_hub_workload_identity
+      init_meshca_with_hub_idns
+    else
+      init_meshca
+    fi
+  fi
+
```
zerobfd: 1) I changed this section before you put up this PR, sorry! You'll have to rebase.

2) I think we can put this logic inside of init_meshconfig instead of here.

scripts/asm-installer/install_asm
```diff
@@ -1411,6 +1444,23 @@ is_workload_identity_enabled() {
   if [[ "${ENABLED}" = 'null' ]]; then false; fi
 }
 
+get_hub_workload_identity() {
+  local ID_NUMBER
+  ID_NUMBER=`kubectl get membership -o=json | jq '.items | length'`
+  if [[ ${ID_NUMBER} -gt 1 ]]; then
+    fatal "Found more than one Hub Membership. Please make sure one cluster is associated with one Hub Membership.";
```
zerobfd, line 1447: Is this possible?

scripts/asm-installer/install_asm
```diff
@@ -692,6 +710,10 @@ EOF
     *) fatal "MODE must be one of 'install', 'migrate', 'upgrade'";;
   esac
 
+  if [[ "${MODE}" -ne "install" && "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
```
zerobfd: -ne only works for integer comparisons, to compare strings you need to use !=. (shell is weird)

scripts/asm-installer/install_asm
```diff
@@ -1411,6 +1444,23 @@ is_workload_identity_enabled() {
...
 }
 
+get_hub_workload_identity() {
+  local ID_NUMBER
+  ID_NUMBER=`kubectl get membership -o=json | jq '.items | length'`
+  if [[ ${ID_NUMBER} -gt 1 ]]; then
+    fatal "Found more than one Hub Membership. Please make sure one cluster is associated with one Hub Membership.";
+  fi
+  local IDENTITY_PROVIDER
+  IDENTITY_PROVIDER=`kubectl get membership -o=json | jq .items[0].spec.identity_provider`
```
zerobfd, line 1450: It looks like you're calling `kubectl get membership` twice. Can we do something like this?

MEMBERSHIPS="$(kubectl get membership -o=json | jq '.items')"
NUM_MEMBERSHIPS="$(echo "${MEMBERSHIPS}" | jq 'length')"
if $NUM_MEMBERSHIPS -gt 1...
...
IDENDITY_PROVIDER = "$(echo "${MEMBERSHIPS}" | jq .[0].spec.identity_provider)"

scripts/asm-installer/install_asm
```diff
@@ -1411,6 +1444,23 @@ is_workload_identity_enabled() {
   if [[ "${ENABLED}" = 'null' ]]; then false; fi
 }
 
+get_hub_workload_identity() {
+  local ID_NUMBER
+  ID_NUMBER=`kubectl get membership -o=json | jq '.items | length'`
```
zerobfd, line 1445: Please use "$(...)" for subshells instead of `...`

scripts/asm-installer/install_asm
```diff
@@ -538,6 +553,9 @@ parse_args() {
       -o | --option)
         arg_required "${@}"
         OPTIONAL_OVERLAY="${2},${OPTIONAL_OVERLAY}"
+        if [[ ${2} == "hub-meshca" ]]; then
```
zerobfd, line 552: Please wrap "${2}" in double quotes

scripts/asm-installer/install_asm
```diff
@@ -1411,6 +1444,23 @@ is_workload_identity_enabled() {
   if [[ "${ENABLED}" = 'null' ]]; then false; fi
 }
 
+get_hub_workload_identity() {
+  local ID_NUMBER
+  ID_NUMBER=`kubectl get membership -o=json | jq '.items | length'`
+  if [[ ${ID_NUMBER} -gt 1 ]]; then
```
zerobfd, line 1446: Please wrap ${...} in double quotes

scripts/asm-installer/install_asm
```diff
@@ -1524,6 +1574,10 @@ configure_package() {
     run kpt cfg set asm anthos.servicemesh.hub "${_CI_ASM_IMAGE_LOCATION}"
     run kpt cfg set asm anthos.servicemesh.tag "${RELEASE}"
   fi
+  if [[ "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
+    run kpt cfg set asm gcloud.project.environProjectID ${ENVIRON_PROJECT_ID}
+    run kpt cfg set asm anthos.servicemesh.hubMembershipID ${HUB_MEMBERSHIP_ID}
```
zerobfd, line 1575: Please double quote variables

scripts/asm-installer/install_asm
```diff
@@ -1411,6 +1440,23 @@ is_workload_identity_enabled() {
...
+  local ID_NUMBER
+  ID_NUMBER=`kubectl get membership -o=json | jq '.items | length'`
+  if [[ ${ID_NUMBER} -gt 1 ]]; then
+    fatal "Found more than one Hub Membership. Please make sure one cluster is associated with one Hub Membership.";
+  fi
+  local IDENTITY_PROVIDER
+  IDENTITY_PROVIDER=`kubectl get membership -o=json | jq .items[0].spec.identity_provider`
+  if [[ ${IDENTITY_PROVIDER} =~ ^\"https:\/\/gkehub.googleapis.com\/projects\/(.+)\/locations\/global\/memberships\/(.+)\"$ ]]; then
+    ENVIRON_PROJECT_ID=${BASH_REMATCH[1]}
+    HUB_MEMBERSHIP_ID=${BASH_REMATCH[2]}
```
zerobfd, line 1453: It looks like CI isn't really working properly right now, but I don't think this is posh compatible. We try to avoid bash specific functionality in case someone wants to use busybox or something in automation.

You should be able to do something like

```
read -r ENVIRON_PROJECT_ID HUB_MEMBERSHIP_ID <(echo "${IDENTITY_PROVIDER}" | sed 's/^\"https:\/\/gkehub.googleapis.com\/projects\/(.+)\/locations\/global\/memberships\/(.+)\"$/\1 \2/g')
```

scripts/asm-installer/install_asm
```diff
@@ -135,11 +141,16 @@ main() {
...
+    if [[ "${CA}" = "mesh_ca" && "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
+      get_hub_workload_identity
+      init_meshca_with_hub_idns
+    else
+      enable_workload_identity
+      # As of now there's no way to check this without trying to initialize it,
+      # by this point if we're verifying then we'll already know the user has
+      # the proper IAM roles for the curl call.
+      init_meshconfig
+    fi
```
zerobfd, line 153: I think we can wrap all of this logic in `enable_workload_identity`

scripts/asm-installer/install_asm
```diff
@@ -1350,6 +1368,17 @@ Authorization: Bearer ${TOKEN}
...
+  info "Initializing Mesh CA with Hub..."
+  local TOKEN; TOKEN="$(retry 2 gcloud --project="${PROJECT_ID}" auth print-access-token)"
+  run curl --request POST \
+    --data '{"workloadIdentityPools":["'${ENVIRON_PROJECT_ID}'.hub.id.goog"]}' -o /dev/null \
+    "https://meshconfig.googleapis.com/v1alpha1/projects/${PROJECT_ID}:initialize" \
+    --header @- <<EOF
+Authorization: Bearer ${TOKEN}
+EOF
+}
+
```
zerobfd, line 1381: I think we can just have a `get_workload_pool` function which will return a different string based on whether we're using hub or not, then the initialize call logic should be the same.

zerobfd (changes requested):

scripts/asm-installer/install_asm
jtrbs: Hi @zerobfd, thanks a lot for reviewing! Yes I was to discuss about whether we want to give user option to set them. We can initialize these just to 0 and "" for 1.8 first release.

scripts/asm-installer/install_asm
jtrbs: I discussed this topic with GKE HUB team. `more than one Hub Membership`, if customer really wants, there are ways around it.  But it is not standard supported ways to register the cluster as there is exclusivity check enforced on client side. And I changed the command as they suggested. We do not need to worry about multiple Memberships case.

scripts/asm-installer/install_asm
```diff
@@ -1411,9 +1424,25 @@ is_workload_identity_enabled() {
...
+  read -r ENVIRON_PROJECT_ID HUB_MEMBERSHIP_ID <<EOF
+${IDENTITY}
+EOF
+  MESHAPI_POST_DATA='{"workloadIdentityPools":["'${ENVIRON_PROJECT_ID}'.hub.id.goog"]}'
+  info "Cluster is in the Hub of project ${ENVIRON_PROJECT_ID} with Membership ID ${HUB_MEMBERSHIP_ID}"
+}
+
 enable_workload_identity(){
+  # Enable workload identity is not needed for Hub IDNS as the cluster needs to be registered with Hub IDNS before the installation script
+  if [[ "${CA}" == "mesh_ca" && "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
```
zerobfd, line 1441: If Hub IDNS and Citadel isn't supported, let's check that combo in validate_args so we don't have to worry about it later.

scripts/asm-installer/install_asm
```diff
@@ -1411,9 +1424,25 @@ is_workload_identity_enabled() {
...
+    get_hub_workload_identity
+    return
+  fi
```
zerobfd, line 1444: Instead of putting logic here, how about in init_meshconfig:

```
init_meshconfig() {
get_token
if [[ use_hub_idns -eq 1 ]]; then
  POST_DATA="$(get_hub_workload_identity)"
  curl --post POST_DATA
else
  curl --post ''
fi
```

And then add an `echo "${MESHAPI_POST_DATA}"` at the end of `get_hub_workload_identity`. This would also let us get rid of the global MESHAPI_POST_DATA.

scripts/asm-installer/install_asm
```diff
@@ -1524,6 +1553,10 @@ configure_package() {
     run kpt cfg set asm anthos.servicemesh.hub "${_CI_ASM_IMAGE_LOCATION}"
     run kpt cfg set asm anthos.servicemesh.tag "${RELEASE}"
   fi
+  if [[ "${ENABLE_HUB_IDNS}" -eq 1 ]]; then
+    run kpt cfg set asm gcloud.project.environProjectID "${ENVIRON_PROJECT_ID}"
+    run kpt cfg set asm anthos.servicemesh.hubMembershipID "${HUB_MEMBERSHIP_ID}"
+  fi
```
zerobfd, line 1559: I'm not sure if it matters for Hub IDNS specifically, but if you call the script with --print-config then `ENVIRON_PROJECT_ID` and `HUB_MEMBERSHIP_ID` might not be set. If that matters then you should probably add a guard to set them, something like

```
if [[ -z ENVIRON || -z HUB_ID ]]; then
  get_hub_workload_identity > /dev/null
fi
```

zerobfd (changes requested):

zerobfd:
Also, can we change the `ENABLE_HUB_IDNS` variable to `USE_HUB_IDNS` instead? Right now it means something different from all of the other `ENABLE_` variables.

zerobfd (approved):
Thanks for all of the work!

PR #276, opened 2020-11-21
Fix cluster-name substitution used in multi-cluster

ruigulala:


asm/istio/options/multicluster.yaml
```diff
@@ -18,5 +18,5 @@ spec:
   values:
     global:
       multiCluster:
-        # Provided to ensure a human readable name rather than a UUID. gke_PROJECT_ID_us-central1-c_asm-cluster
-        clusterName: "cn-PROJECTID-us-central1-c-asm-cluster" # {"$ref":"#/definitions/io.k8s.cli.substitutions.cluster-name"}
\ No newline at end of file
+        # Provided to ensure a human readable name rather than a UUID. PROJECT_ID/CLUSTER_LOCATION/CLUSTER_NAME
+        clusterName: "PROJECT_ID/CLUSTER_LOCATION/CLUSTER_NAME" # {"$ref":"#/definitions/io.k8s.cli.substitutions.cluster-name"}
```
elfinhe, line 22: we need to wait for a consensus for the naming convention of the multiCluster.clusterName.

btw this may break the code in injection: https://github.com/istio/istio/blob/c1e7be13999d56cadab103dc79069f41ac4155ad/pkg/kube/inject/webhook.go#L748 where the cluster name is parsed by "/".

ruigulala: Where can I confirm that we pass clusterName to the webhook?
IIUC, cluster-name that is used in multi-cluster will not be passed to webhook.

elfinhe: hmm, that's a good question, if we want the cluster name to be "PROJECT_ID/CLUSTER_LOCATION/CLUSTER_NAME"

the injector may failed to recognize/generate the "ISTIO_META_CLUSTER_ID" env from the path
e.g., "/inject/cluster/PROJECT_ID/CLUSTER_LOCATION/CLUSTER_NAME".

I am not 100% sure it will break, but introducing "/" may not be a good idea.

ruigulala: Thanks for the info! We've fixed the issue on our side. So this PR is not needed anymore.

PR #285, opened 2020-11-24
Upgrade to ASM in Anthos Sample Deployment to 1.7.3-asm.6

katze120:
Upgrade to ASM in Anthos Sample Deployment to 1.7.3-asm.6.
* Sync-ed `IstioOperator` from https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/release-1.7-asm/asm/istio/istio-operator.yaml but kept the ASD custom resource request.
* Added Canonical Service from https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/release-1.7-asm/asm/canonical-service/controller.yaml with reduced CPU request.
* Sync-ed `Kptfile` from https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/release-1.7-asm/asm/Kptfile
* Remove patch files for [ISTIO-SECURITY-2020-007](https://istio.io/latest/news/security/istio-security-2020-007/), which is not applicable for Istio 1.7.


katze120:
@elfinhe I've reached out internally regarding the Validation Webhook & CanonicalService controller. If those are recommended or required for ASD, I can add them to this PR too. Thanks.

asm/istio/istio-operator.yaml
```diff
@@ -14,14 +14,10 @@
 
 apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
```
elfinhe, line 16: We've simplified the configs to reduce some input from customers. We can just copy the file below:
https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/release-1.7-asm/asm/istio/istio-operator.yaml

Then we can remove those 3 kpt commands below from your install script:
```
kpt cfg set xxx/asm gcloud.compute.location "$ZONE"
kpt cfg set xxx/asm gcloud.project.projectNumber "$PROJECT_NUMBER"
kpt cfg set xxx/asm gcloud.container.cluster "$CLUSTER"
```

elfinhe (approved):
LGTM, but need to verify it on staging before prod.

PR #286, opened 2020-11-24
remove duplicate trustdomain entry

richardwxn:
this is already set by the field meshConfig.defaultConfig.ProxyMetadata.TRUST_DOMAIN, not sure why we have two places here

and actually this is the reason why we saw `dulicate entry for env` error before when we use server-side apply


richardwxn:
@zerobfd @elfinhe

zerobfd (approved):

PR #290, opened 2020-11-24
Installation script changes to incorporate private ca

shankgan:


scripts/asm-installer/install_asm
```diff
@@ -450,7 +456,11 @@ OPTIONS:
                                       omitted if not installing optional
                                       features. To add multiple files, specify
                                       them with multiple options one at a time.
-
+  --ca_id|--ca-id         <CA_ID>     Required only if --ca option is private_ca.
```
pmansour, line 459: a note on terminology - the resource *ID* is just the last part of the full string, while the resource *name* is the `projects/foo/locations/bar/certificateAuthorities/baz` string you described below. In this case, I think they are flipped around.

more context: https://google.aip.dev/122

scripts/asm-installer/install_asm
```diff
@@ -1598,6 +1627,13 @@ install_secrets() {
     --from-file="${CA_CHAIN}"
 }
 
+init_private_ca() {
+  WORKLOAD_IDENTITY="$PROJECT_ID.svc.id.goog[istio-system/istiod-service-account]"
+  retry 3 run gcloud projects add-iam-policy-binding $PROJECT_ID \
```
pmansour, line 1632: (discussed offline yesterday) we can now add this binding on the CA itself

scripts/asm-installer/install_asm
shankgan: Thanks for the clarification! Will rename it to resource name

scripts/asm-installer/install_asm
myidpt: @pmansour what does that mean? We still using the IAM binding, and the GCP SA to call the CAS, right?

pmansour: We plan to use the 3p identity directly to call the CAS, but what I meant here is that we can grant the IAM role on the CA resource directly (vs. granting it on the parent project resource). This just means a more granular IAM binding which is always good for security :)

myidpt: Thanks for the clarification.

myidpt (approved):

shankgan:
@zerobfd - Could you review the changes?

zerobfd (approved):

PR #295, opened 2020-11-25
fix single quota in validate step

gustavodelima:
Without single quotes the RESULT variable returns with error


zerobfd:
I think changing those to single quotes will prevent the subshell from running the `gcloud` command. `$RESULT` would be set to the literal string, so the `-z` test will always return true.

What error are you getting? Did running it with `-v` give any extra information?

gustavodelima:
@googlebot  I signed it!

gustavodelima:
totalmente faz sentido @zerobfd , thank you

PR #301, opened 2020-11-30
Use -K with curl instead of @-

### LINKED ISSUE #296: install_asm fails to call meshconfig API
Hello! I'm using install_asm as downloaded from the storage bucket here:

```
https://storage.googleapis.com/csm-artifacts/asm/install_asm_1.7
```

I've got a very simple setup. Just trying to install ASM into a single GKE cluster in one project.

I'm using a service account created specifically for this purpose.

The validation succeeds (i.e. running install_asm with --only_validate).

However, when I run the script to actually install ASM, I get a "401 Unauthorized" error when the script tries to access the meshconfig API. Here's how I'm invoking the script:

```
    ./install_asm -v \
          --project_id myprojectid \
          --cluster_name myclustername \
          --cluster_location us-central1-a \
          --mode install \
          --enable_apis \
          --service_account anthos-installer@myprojectid.iam.gserviceaccount.com \
          --key_file ./sa.json
```

It gets all the way down to "Initializing Mesh CA", and then fails with a "401 Unauthorized" as below:

```
install_asm: Initializing Mesh CA...
install_asm: Running: 'curl --request POST --fail --data  -o /dev/null https://meshconfig.googleapis.com/v1alpha1/projects/myprojectid:initialize --header @-'
install_asm: -------------
curl: (22) The requested URL returned error: 401 Unauthorized
```

I notice that the script is granting the following roles to the service account:

```
editor
compute.admin
container.admin
resourcemanager.projectIamAdmin
iam.serviceAccountAdmin
iam.serviceAccountKeyAdmin
gkehub.admin
```

However, the script does not grant `meshconfig.admin`. I'm guessing that that might be required here?

Also interesting: I see no roles related to the `meshconfig` API in the Google Cloud web console for my project, not even when I search for the `meshconfig` string in the Roles section.

The `meshconfig` API is fully enabled for my project.

I'm at a bit of a loss here. Any help would be appreciated. Thank you!

zerobfd:
Fixes #296

### DIFF
--- a/scripts/asm-installer/cloudbuild.yaml
+++ b/scripts/asm-installer/cloudbuild.yaml
@@ -21,13 +21,16 @@ steps:
   - '${_SCRIPT_NAME}'
   waitFor:
   - 'publish-tester-image'
-- name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
-  dir: 'scripts/asm-installer'
-  id: 'check-posh-compatibility'
-  entrypoint: 'posh'
-  args: ['-n', '${_SCRIPT_NAME}']
-  waitFor:
-  - 'publish-tester-image'
+
+    # Disabled for now, see #296. We have more users with LTS Linux releases
+    # than non-Bourne shells. Please open an issue if you're affected.
+    #- name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
+    #  dir: 'scripts/asm-installer'
+    #  id: 'check-posh-compatibility'
+    #  entrypoint: 'posh'
+    #  args: ['-n', '${_SCRIPT_NAME}']
+    #  waitFor:
+    #  - 'publish-tester-image'
 
 - name: 'gcr.io/cloud-builders/gcloud'
   dir: 'scripts/asm-installer'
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/cloudbuild.yaml
```diff
@@ -21,13 +21,16 @@ steps:
...
-  - 'publish-tester-image'
+
+    # Disabled for now, see #296. We have more users with LTS Linux releases
+    # than non-Bourne shells. Please open an issue if you're affected.
+    #- name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
+    #  dir: 'scripts/asm-installer'
+    #  id: 'check-posh-compatibility'
+    #  entrypoint: 'posh'
+    #  args: ['-n', '${_SCRIPT_NAME}']
+    #  waitFor:
```
backward-compatible, line 32: How is the posh compatibility check related to the curl command update?

zerobfd: The <(command) construction is bash specific so it posh fails on it. I still want to have this at some point, but until we come up with a POSIX compatible way or we have no more users with curl versions < 7.55 then I think we have to give it up.

backward-compatible: got it, thanks!

backward-compatible (approved):

PR #311, opened 2020-12-02
Update installation script for Hub WIP to be compatible with both GKE and Hub WIP

jtrbs:
This is for the case that users have existing GKE on GCP clusters in their GCP project running mesh with GKE WIP and they want to try new mesh installation with script and Hub overlay in a new cluster of the same project. By specify both GKE and Hub WIP, we will not break existing mesh with GKE WIP.

Per discussion, we will hold off the multiproject setup for Hub at the master branch. We have a PR for master-prow branch to check the Hub multiproject setup https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/310.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1371,7 +1371,8 @@ init_meshconfig() {
 ${IDENTITY}
 EOF
     info "Cluster is in the Hub of project ${ENVIRON_PROJECT_ID} with Membership ID ${HUB_MEMBERSHIP_ID}"
-    local POST_DATA; POST_DATA='{"workloadIdentityPools":["'${ENVIRON_PROJECT_ID}'.hub.id.goog"]}'
+    # Support both Hub Workload Identity Pool and GKE Workload Identity Pool.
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/install_asm
```diff
@@ -1371,7 +1371,8 @@ init_meshconfig() {
 ${IDENTITY}
 EOF
     info "Cluster is in the Hub of project ${ENVIRON_PROJECT_ID} with Membership ID ${HUB_MEMBERSHIP_ID}"
-    local POST_DATA; POST_DATA='{"workloadIdentityPools":["'${ENVIRON_PROJECT_ID}'.hub.id.goog"]}'
+    # Support both Hub Workload Identity Pool and GKE Workload Identity Pool.
```
zerobfd, line 1374: Can you make the comment a bit clearer that this call replaces the pools and doesn't just initialize it? Maybe

```
# :initialize replaces the existing pools, so we need to support both Hub and GKE Workload Identity Pools
```

zerobfd (approved):

PR #314, opened 2020-12-04
Add third party manifests to asm

backward-compatible:


zerobfd (approved):
Can you add the source for these in the commit message? e.g. "pulled from https://X"

elfinhe (approved):

PR #316, opened 2020-12-04
Add functionality for managed control plane

zerobfd:
This shouldn't be merged until cloudbuild.yaml is fixed back

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -15,15 +15,15 @@ fi
 
 ### Internal variables ###
 MAJOR="${MAJOR:=1}"
-MINOR="${MINOR:=7}"
-POINT="${POINT:=3}"
-REV="${REV:=6}"
+MINOR="${MINOR:=8}"
+POINT="${POINT:=1}"
+REV="${REV:=2}"
 RELEASE="${MAJOR}.${MINOR}.${POINT}-asm.${REV}"
 RELEASE_LINE="${MAJOR}.${MINOR}."
 PREVIOUS_RELEASE_LINE="${MAJOR}.$(( MINOR - 1 ))."
 REVISION_LABEL="asm-${MAJOR}${MINOR}${POINT}-${REV}"
 KPT_BRANCH="${_CI_ASM_KPT_BRANCH:=release-${MAJOR}.${MINOR}-asm}"
-readonly RELEASE; readonly RELEASE_LINE; readonly REVISION_LABEL; readonly KPT_BRANCH;
+readonly RELEASE; readonly RELEASE_LINE; readonly KPT_BRANCH;
 K8S_MINOR=0
 
 ### These are hooks for Cloud Build to be able to use debug/staging images
@@ -31,6 +31,8 @@ K8S_MINOR=0
 ### in CI/CD.
 _CI_ASM_IMAGE_LOCATION="${_CI_ASM_IMAGE_LOCATION:=}"
 _CI_ASM_PKG_LOCATION="${_CI_ASM_PKG_LOCATION:=}"
+_CI_CLOUDRUN_IMAGE_HUB="${_CI_CLOUDRUN_IMAGE_HUB:=}"
+_CI_CLOUDRUN_IMAGE_TAG="${_CI_CLOUDRUN_IMAGE_TAG:=}"
 
 ### File related constants ###
 ISTIO_FOLDER_NAME="istio-${RELEASE}"; readonly ISTIO_FOLDER_NAME;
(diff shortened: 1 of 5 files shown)


scripts/asm-installer/tests/run_basic_suite_managed
```diff
@@ -0,0 +1,110 @@
...
+      -n "${CLUSTER_NAME}" \
+      -p "${PROJECT_ID}" \
+      -m install \
+      --managed \
+      -c mesh_ca -v -e
+  fi
+
+  sleep 5
+  echo "Installing Istio manifests for demo app..."
+  install_demo_app_istio_manifests "${NAMESPACE}"
```
ruigulala, line 77: Why do we need to re-install the app in the same namespace? Simply re-label the namespace is not enough?

scripts/asm-installer/tests/run_basic_suite_managed
```diff
@@ -0,0 +1,110 @@
...
+
+  echo "Performing a rolling restart of the demo app..."
+  roll "${NAMESPACE}"
```
ruigulala, line 80: Where did we re-label the namespace to be "istio.io/rev=asm-managed"? It seems that the namespace label is still "istio-injection=enabled".

scripts/asm-installer/install_asm
```diff
@@ -127,8 +132,29 @@ main() {
...
+      enable_stackdriver_kubernetes
+    else
+      exit_if_no_workload_identity
+      exit_if_stackdriver_not_enabled
+    fi
+    if can_modify_gcp_iam_roles; then
+      bind_user_to_iam_policy "roles/meshconfig.admin" "$(local_iam_user)"
+    fi
+    local INIT_ERROR; INIT_ERROR=0;
+    MANAGED_SERVICE_ACCOUNT="$(init_meshconfig_managed | jq -r .serviceAccount)" || INIT_ERROR=1
```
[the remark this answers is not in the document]
zerobfd: Whoops, forgot to remove that. Thanks.

scripts/asm-installer/install_asm
```diff
@@ -127,8 +132,29 @@ main() {
...
+    info "Got ${MANAGED_SERVICE_ACCOUNT} from mesh initialization"
+    if [[ "${INIT_ERROR}" -eq 1 ]]; then
```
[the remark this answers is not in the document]
zerobfd: Yep, I was doing this trying to debug and forgot to change it back.

scripts/asm-installer/tests/run_basic_suite_managed
zerobfd: We don't here, there aren't any checks for sidecars for this test.

scripts/asm-installer/tests/run_basic_suite_managed
zerobfd: This doesn't reinstall the app, just the Istio objects to configure ingress etc.

scripts/asm-installer/tests/run_basic_suite_managed
ruigulala: Ah, so this test is to make sure all the existing application still works with the new gateway and MCP.

ruigulala: Btw, why don't we install the app with asm-managed label in a new namespace? Is it because it's not working or we want to add it later?

zerobfd: >> still works

Yep, that's how most of these tests work.

>> why don't we

Mostly because of time. If you have a good idea for a test or a way to refactor them (or make them faster!) I'm happy for ideas or PRs.

ruigulala: SG. I'll add one.

ruigulala (approved):

scripts/asm-installer/install_asm
```diff
@@ -341,13 +372,15 @@ can_modify_gcp_apis() {
...
+  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_GCP_IAM_ROLES}" -eq 0 ]] && \
+    ! is_managed ||
+    ! can_modify_at_all; then false; fi
 }
 
 can_modify_gcp_components() {
-  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_GCP_COMPONENTS}" -eq 0 ]] \
-    || ! can_modify_at_all; then false; fi
+  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_GCP_COMPONENTS}" -eq 0 ]] &&
+    ! is_managed || \
```
cagataygurturk, line 382: @zerobfd I am running into issues with this line. Please check #419.

PR #319, opened 2020-12-09
Add identityprovider

JimmyCYJ:


zerobfd (approved):
lgtm. Could you change the filename for the crd to be `identityprovider-crd.yaml` so it's more obvious that it's not normal config like the other files in this repo?

JimmyCYJ:
> lgtm. Could you change the filename for the crd to be `identityprovider-crd.yaml` so it's more obvious that it's not normal config like the other files in this repo?

Good point. I have renamed the CRD file. Thanks!

PR #321, opened 2020-12-10
Use more compatible shebang

zerobfd:
Specifically for those running on a Mac, many people install an updated
version of bash in a different location to avoid stomping on the system
bash. This will help default to the correct shell in more environments.


jtrbs (approved):

jtrbs:
I have tested it on Mac. Works well. Thanks! 👍

PR #323, opened 2020-12-11
ILB - add correct targetPort for ports <1024

kenthua:
With the current IstioOperator CR, the installation fails.

```
Error: failed to install manifests: generated config failed semantic validation: port http2/80 in gateway istio-ingressgateway invalid: targetPort is set to 0, which requires root. Set targetPort to be greater than 1024 or configure values.gateways.istio-ingressgateway.runAsRoot=true, port https/443 in gateway istio-ingressgateway invalid: targetPort is set to 0, which requires root. Set targetPort to be greater than 1024 or configure values.gateways.istio-ingressgateway.runAsRoot=true
```

proxyv2 identified ports:
```
Containers:
  istio-proxy:
    ...
    Ports:         15020/TCP, 8080/TCP, 8443/TCP, 15090/TCP
```

### DIFF
--- a/asm/istio/options/internal-load-balancer.yaml
+++ b/asm/istio/options/internal-load-balancer.yaml
@@ -29,6 +29,8 @@ spec:
             port: 15020
           - name: http2
             port: 80
+            targetPort: 8080
           - name: https
             port: 443
+            targetPort: 8443
 # [END docs]


elfinhe (approved):

kenthua:
@elfinhe The PR looks stuck on a check.

```
Summary
Changed files did not match file filter
```

elfinhe:
The CI test hasn't exercised this config file. We can merge it.

kenthua:
Thanks.  We'll probably need to cherry pick this into `stage-1.8-asm` and `master` ?

zerobfd:
If this is generally applicable, it should be merged into master first and then cherry-picked where necessary, not committed directly to a release branch.

kenthua:
Oops, then i may have had the PR into the wrong branch then.  It definitely affects 1.7 right now, but likely hood it will be the same in 1.8.

zerobfd:
I put it in the right place, just wanted to note that for next time. Thanks for putting the fix in!

PR #325, opened 2020-12-11
Add initial version of release script for install_asm

zerobfd:


elfinhe (commented):
Can we add a line into Readme to explain how and who will be running this?

elfinhe (approved):

zerobfd:
> Can we add a line into Readme to explain how and who will be running this?

Good idea, I'll add that before merging.

PR #327, opened 2020-12-14
Add expansion config

zerobfd:


scripts/gen-eastwest-gateway.sh
```diff
@@ -0,0 +1,129 @@
...
+done
+# single-cluster installations may need this gateway to allow VMs to get discovery
+# for non-single cluster, we add additional topology information
+SINGLE_CLUSTER="${SINGLE_CLUSTER:-0}"
+if [[ "${SINGLE_CLUSTER}" -eq 0 ]]; then
+  if [[ -z "${CLUSTER:-}" ]] || [[ -z "${NETWORK:-}" ]] || [[ -z "${MESH:-}" ]]; then
+    echo "Must specify either --single-cluster or --mesh, --cluster, and --network."
+    exit 1
+  fi
+fi
```
jtrbs, line 52: Do we need to warn if user happens to specify all --single-cluster, --mesh, --cluster, and --network?

jtrbs (approved):

scripts/gen-eastwest-gateway.sh
zerobfd: Hopefully we'll have a better method to configure this soon, so I don't want to put too much time into error handling. (Good catch though.)

backward-compatible (changes requested):
Can you pull the latest version from https://github.com/istio/istio/blob/master/samples/multicluster/gen-eastwest-gateway.sh which also has the new _--revision _ flag

zerobfd:
This is the one from the 1.8 release branch, are we sure that master is compatible?

backward-compatible (approved):

jasonwzm (approved):

PR #330, opened 2020-12-15
Private CA: Install script integration changes

shankgan:
Installation script changes to integrate ASM with Google Certificate Authority Service - https://cloud.google.com/certificate-authority-service

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -81,6 +81,8 @@ CA_CERT="${CA_CERT:=}"
 CA_KEY="${CA_KEY:=}"
 CA_ROOT="${CA_ROOT:=}"
 CA_CHAIN="${CA_CHAIN:=}"
+CA_ID="${CA_ID:=}"
+
 
 DRY_RUN="${DRY_RUN:=0}"
 ONLY_VALIDATE="${ONLY_VALIDATE:=0}"
@@ -169,6 +171,9 @@ main() {
     if can_modify_gcp_components; then
       enable_workload_identity
       init_meshconfig
+      if [[ "${CA}" = "private_ca" ]]; then
+        init_private_ca
+      fi
       enable_stackdriver_kubernetes
     elif should_validate; then
       exit_if_no_workload_identity
@@ -422,6 +427,7 @@ OPTIONS:
   --ca_key               <FILE PATH>
   --root_cert            <FILE PATH>
   --cert_chain           <FILE PATH>
+  --ca_id                <CA-ID>
 
 FLAGS:
   -e|--enable_all
@@ -468,7 +474,7 @@ OPTIONS:
                                       used. Defaults to "mesh_ca" for --mode
                                       install. Specifying the CA is required
                                       for --mode migrate.  Allowed values for
-                                      <CA> are {citadel|mesh_ca}.
+                                      <CA> are {citadel|mesh_ca|private_ca}.
   -o|--option            <FILE NAME>  The name of a YAML file in the kpt pkg to
                                       apply. For options, see the
                                       anthos-service-mesh-package GitHub
@@ -496,7 +502,11 @@ OPTIONS:
                                       omitted if not installing optional
                                       features. To add multiple files, specify
                                       them with multiple options one at a time.
-
+  --ca_id|--ca-id         <CA_ID>     Required only if --ca option is private_ca.
+                                      Name of the ca in the GCP CAS service used to
+                                      sign certificates in the format
+                                      'projects/$(project_name)/locations/ \
+                                      $(ca_region)/certificateAuthorities/$(ca_name)'
   The following four options must be passed together and are only necessary
   for using a custom certificate for Citadel. Users that aren't sure whether
   they need this probably don't.
@@ -596,6 +606,11 @@ parse_args() {
       -c | --ca)
         arg_required "${@}"
         CA="$(echo "${2}" | tr '[:upper:]' '[:lower:]')"
+        shift 2
+        ;;
+      --ca_id | --ca-id)
+        arg_required "${@}"
+        CA_ID="$(echo "${2}" | tr '[:upper:]' '[:lower:]')"
         shift 2
         ;;
       -o | --option)
@@ -754,12 +769,12 @@ EOF
 
   if [[ "${MODE}" != "upgrade" ]]; then
     case "${CA}" in
-      citadel | mesh_ca);;
+      citadel | mesh_ca | private_ca);;
       "")
         MISSING_ARGS=1
         warn "Missing value for CA"
         ;;
-      *) fatal "CA must be one of 'citadel', 'mesh_ca'";;
+      *) fatal "CA must be one of 'citadel', 'mesh_ca', 'private_ca'";;
     esac
   fi
 
@@ -787,6 +802,10 @@ EOF
 
   if [[ "${CUSTOM_CA}" -eq 1 ]]; then
     validate_certs
+  fi
+
+  if [[ "${CA}" = "private_ca" ]]; then
+    validate_private_ca
   fi
 
   while read -r FLAG; do
@@ -844,6 +863,14 @@ EOF
 
   set_kpt_package_url
   WORKLOAD_POOL="${PROJECT_ID}.svc.id.goog"
+}
+
+validate_private_ca() {
+  if [[ -n ${CA_ID}]]; then
+    fatal "A CA_ID must be provided for integration with private ca."
+  fi
+  #TODO: Validate that the CA_ID must be of the form 
+  # 'projects/$(project_name)/locations/$(ca_region)/certificateAuthorities/$(ca_name)'
 }
 
 validate_certs() {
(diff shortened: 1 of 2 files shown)


jtrbs (approved):

scripts/asm-installer/install_asm
```diff
@@ -1705,6 +1745,20 @@ install_secrets() {
     --from-file="${CA_CHAIN}"
 }
 
+init_private_ca() {
+  WORKLOAD_IDENTITY="$PROJECT_ID.svc.id.goog[istio-system/istiod-service-account]"
```
jtrbs, line 1749: Will "init_private_ca" work with Hub workload identity in the future?

shankgan: yes, but NOT for preview. HUB workload identity is something we will integrate with for GA

scripts/asm-installer/install_asm
```diff
@@ -496,7 +502,11 @@ OPTIONS:
                                       omitted if not installing optional
                                       features. To add multiple files, specify
                                       them with multiple options one at a time.
-
+  --ca_name|--ca-name    <CA NAME>    Required only if --ca option is private_ca.
```
zerobfd, line 505: The canonical form is just the snake_case version, don't list the hyphenated-case version as well.

scripts/asm-installer/install_asm
```diff
@@ -1288,33 +1326,32 @@ is_rev_invalid() {
...
-  ! is_meshca_installed && [[ "${CA}" = "mesh_ca" ]] && return 0
-}
-
-is_ca_meshca_to_citadel() {
-  is_meshca_installed && [[ "${CA}" = "citadel" ]] && return 0
-}
-
 is_meshca_installed() {
+  local INSTALLED_CA; INSTALLED_CA="$(kubectl -n istio-system get pod -l istio=ingressgateway \
+    -o jsonpath='{.items[].spec.containers[].env[?(@.name=="CA_ADDR")].value}')"
```
zerobfd, line 1345: Could you separate the "getting the installed CA" logic from the "check what CA it is" logic? something like

```
get_installed_ca() {
  kubectl -n istio-system get pod \
    -l istio=ingressgateway \
    -o jsonpath='{.items[].spec.containers[].env[?(@.name=="CA_ADDR")].value}'
}

is_privateca_installed() {
  if [[ "$(get_installed_ca)" != "ISTIOD_RA_CAS_API" ]]; then false; fi
}
```
And put "ISTIOD_RA_CAS_API" in a readonly variable at the top of the file.

scripts/asm-installer/install_asm
```diff
@@ -1693,6 +1730,9 @@ configure_package() {
     run kpt cfg set asm gcloud.project.environProjectID "${ENVIRON_PROJECT_ID}"
     run kpt cfg set asm anthos.servicemesh.hubMembershipID "${HUB_MEMBERSHIP_ID}"
   fi
+  if [[ -n "${CA_NAME}" ]]; then
+    run kpt cfg set asm anthos.servicemesh.external_ca.ca_name "${CA_NAME}"
+  fi
```
zerobfd, line 1735: Unless we have plans for other CAs using the same setter, I'd rather gate this on CA == private_ca as well.

scripts/asm-installer/install_asm
```diff
@@ -1715,6 +1755,20 @@ install_secrets() {
     --from-file="${CA_CHAIN}"
 }
 
+init_private_ca() {
+  WORKLOAD_IDENTITY="$PROJECT_ID.svc.id.goog[istio-system/istiod-service-account]"
```
zerobfd, line 1759: I don't think this needs to be a variable. I'd prefer it be a function:

```
workload_identity() {
  echo "${PROJECT_ID}.svc.id.goog[istio-system/istiod-service-account]"
}
```

If it needs to be a variable for some reason, please use local.

scripts/asm-installer/install_asm
```diff
@@ -1715,6 +1755,20 @@ install_secrets() {
...
+
+  local NAME; NAME=$(echo "${CA_NAME}" | cut -f6 -d/)
+  local CA_LOCATION; CA_LOCATION=$(echo "${CA_NAME}" | cut -f4 -d/)
+  local CA_PROJECT; CA_PROJECT=$(echo "${CA_NAME}" | cut -f2 -d/)
+
+  retry 3 run gcloud beta privateca subordinates add-iam-policy-binding "${NAME}" \
+    --location "${CA_LOCATION}" \
+    --project "${CA_PROJECT}" \
+    --member "serviceAccount:$WORKLOAD_IDENTITY" \
```
zerobfd, line 1768: Use ${} if this remains a variable

scripts/asm-installer/install_asm
```diff
@@ -496,7 +502,11 @@ OPTIONS:
                                       omitted if not installing optional
                                       features. To add multiple files, specify
                                       them with multiple options one at a time.
-
+  --ca_name|--ca-name    <CA NAME>    Required only if --ca option is private_ca.
+                                      Name of the ca in the GCP CAS service used to
```
zerobfd, line 506: Please capitalize 'ca'

scripts/asm-installer/install_asm
```diff
@@ -496,7 +502,11 @@ OPTIONS:
...
+                                      sign certificates in the format
+                                      'projects/project_name/locations/ \
+                                       ca_region/certificateAuthorities/ca_name'
```
zerobfd, line 509: Why does this entire thing need to be entered by the user? We should know $PROJECT_ID and $CA_NAME already. Does Private CA support clusters in different regions?

scripts/asm-installer/install_asm
```diff
@@ -496,7 +502,11 @@ OPTIONS:
                                       omitted if not installing optional
                                       features. To add multiple files, specify
                                       them with multiple options one at a time.
-
+  --ca_name|--ca-name    <CA NAME>    Required only if --ca option is private_ca.
+                                      Name of the ca in the GCP CAS service used to
+                                      sign certificates in the format
```
zerobfd, line 507: I think `sign certificates. The format should be` is clearer, since it removes the possibility that someone reads this as "the format that GCP CAS signs certificates with is 'projects/...'"

scripts/asm-installer/install_asm
```diff
@@ -733,6 +748,10 @@ validate_args() {
       fatal "Citadel is not supported with managed control plane."
     fi
 
+    if [[ "${CA}" == "private_ca" ]]; then
+      fatal "Google Certificate Authority Service integration is not supported with managed control plane."
+    fi
+
     if [[ "${CUSTOM_CA}" -eq 1 ]]; then
       fatal "Specifying a custom CA with managed control plane is not supported."
     fi
```
zerobfd, line 757: Is private CA not a custom CA?

scripts/asm-installer/install_asm
```diff
@@ -733,6 +748,10 @@ validate_args() {
       fatal "Citadel is not supported with managed control plane."
     fi
 
+    if [[ "${CA}" == "private_ca" ]]; then
+      fatal "Google Certificate Authority Service integration is not supported with managed control plane."
```
zerobfd, line 752: Can we be more consistent with naming? The argument is `private_ca`, in the usage prompt it's `GCP CAS`, and here it's different and doesn't match GCP CAS.

scripts/asm-installer/install_asm
```diff
@@ -839,13 +862,28 @@ ${CUSTOM_OVERLAY}
...
   CUSTOM_OVERLAY="${ABS_OVERLAYS}"
 
   set_kpt_package_url
   WORKLOAD_POOL="${PROJECT_ID}.svc.id.goog"
 }
 
+validate_private_ca() {
+  local CA_NAME_TEMPLATE
+  CA_NAME_TEMPLATE="projects/project_name/locations/ca_region/certificateAuthorities/ca_name"
+
```
zerobfd, line 877: Please add `readonly CA_NAME_TEMPLATE` after the declaration

scripts/asm-installer/install_asm
```diff
@@ -839,13 +862,28 @@ ${CUSTOM_OVERLAY}
...
+  elif [[ "${CA}" = "private_ca" ]]; then
+    ABS_OVERLAYS="${OPTIONS_DIRECTORY}/private-ca.yaml,${ABS_OVERLAYS}"
   fi
   CUSTOM_OVERLAY="${ABS_OVERLAYS}"
 
   set_kpt_package_url
   WORKLOAD_POOL="${PROJECT_ID}.svc.id.goog"
 }
 
+validate_private_ca() {
```
zerobfd, line 874: If these are just validating the arguments, this should be `validate_private_ca_args` and be called in `validate_args` instead of further down the stack. That way we can fail as early as possible if we know that the input is malformed.

scripts/asm-installer/install_asm
```diff
@@ -1715,6 +1755,20 @@ install_secrets() {
...
 }
 
+init_private_ca() {
+  WORKLOAD_IDENTITY="$PROJECT_ID.svc.id.goog[istio-system/istiod-service-account]"
+
+  local NAME; NAME=$(echo "${CA_NAME}" | cut -f6 -d/)
+  local CA_LOCATION; CA_LOCATION=$(echo "${CA_NAME}" | cut -f4 -d/)
+  local CA_PROJECT; CA_PROJECT=$(echo "${CA_NAME}" | cut -f2 -d/)
+
+  retry 3 run gcloud beta privateca subordinates add-iam-policy-binding "${NAME}" \
```
zerobfd, line 1765: I'm not sure what this is doing, but it looks like this should only be required per project and not per cluster, right? Can we make the function name a bit more descriptive? Is there a way to check if this is already done before attempting to add the policy binding? Otherwise the caller requires permissions to modify Private CA even if it's already set up--this is the reason for all of the individual `--enable...` flags.

zerobfd (changes requested):
I really apologize for the delay, it's been a busy winter.

scripts/asm-installer/install_asm
shankgan: As of now, CA configuration is not a first class API - there is no common way routine to figure out which CA is installed. Each CA has its own method of detection. This will change in the near future, but that is why there are two separate functions for the different CA's

scripts/asm-installer/install_asm
shankgan: It isn't. Custom CA relies on Citadel's CA logic to sign certificates by using custom key/cert's uploaded by the user. Private CA is a completely different gRPC CA service that signs certificates.

scripts/asm-installer/install_asm
shankgan: I agree that the caller needs permission to modify Private CA IAM. Will add that as well. 

However, this script (as of now) is being called per cluster. In our docs, we mention that we want a sub-ordinate CA per cluster. Therefore, the expectation is that the IAM policy binding command will need to be called each time this script is invoked, even if from the same project

scripts/asm-installer/install_asm
shankgan: Prefer it remains a local function. May be built from more local variables.

scripts/asm-installer/install_asm
shankgan: fixed

scripts/asm-installer/install_asm
shankgan: fixed

scripts/asm-installer/install_asm
shankgan: fixed

scripts/asm-installer/install_asm
zerobfd: You're right, I thought that the kubectl commands were the same but now I realize that the environment variables are different. My bad!

scripts/asm-installer/install_asm
zerobfd: How does this command know what cluster it's attached to?

shankgan: It doesn't, but the CA_NAME is expected to be unique per cluster

scripts/asm-installer/install_asm
```diff
@@ -169,6 +171,9 @@ main() {
     if can_modify_gcp_components; then
       enable_workload_identity
       init_meshconfig
+      if [[ "${CA}" = "gcp_cas" ]]; then
+        init_gcp_cas
```
zerobfd, line 175: Can we have a comment here saying that there's no associated check if you don't have --enable_gcp_components set, so it's possible to have a cluster configured to use privateca with no CSR privs

scripts/asm-installer/install_asm
zerobfd: Sounds good, thanks!

scripts/asm-installer/install_asm
zerobfd: Will the setup fail? Or will it install properly and not be able to request certs?

shankgan: the ASM installation will technically fail. The XDS will never be able to come up if the certificates are not obtained from the CA server...

zerobfd: Ok, gtg as is then.

zerobfd (approved):
There are a few open questions that I'd like resolved, but since this is a preview feature it looks good.

PR #335, opened 2020-12-17
Update install_asm

frots:
Add custom Istio ingress gateway compatbility to validate_ca_consistency() by changing the label selector.


zerobfd (approved):
Thank you for the patch!

PR #348, opened 2021-01-09
Using placeholder for rev and tag

elfinhe:
because install_asm will always set them just like PROJECT_ID, also reducing the work of maintaining them in sync from various places with our release version.

https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm#L17-L20 will be the source of truth of the revision and version.

### DIFF
--- a/asm/istio/istiod-service.yaml
+++ b/asm/istio/istiod-service.yaml
@@ -21,7 +21,7 @@ metadata:
   name: istiod
   namespace: istio-system
   labels:
-    istio.io/rev: asm-173-6 # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev"}
+    istio.io/rev: ASM_REV # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev"}
     app: istiod
     istio: pilot
     release: istio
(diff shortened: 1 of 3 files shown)


zerobfd (approved):
Before this makes it to a release branch we should double check that none of the manual install instructions rely on this, but we should go ahead and merge.

asm/istio/istiod-service.yaml
```diff
@@ -42,4 +42,4 @@ spec:
     protocol: TCP
   selector:
     app: istiod
-    istio.io/rev: asm-173-6 # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev"}
+    istio.io/rev: ASM_REV # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev"}
```
elfinhe, line 45: one thing I am not sure is if somewhere in our doc uses kubectl apply asm/istio/istiod-service.yaml without kpt cfg set. then the placeholder may break that, but now it is still set to asn-173-6, so i guess we are not using it directly without install_asm. 

Anyway the istiod service fix should be transparent. So let's make this happen.

PR #356, opened 2021-01-14
Allow asm installer to prepare VM setup

jasonwzm:
* Add `vm` option and its validation
* Register cluster to Anthos environ
* Install eastwest gateway
* Expose Istiod
* Register Google Identity Provider

Testing verifies cluster is registered and identity provider is registered, and makes sure the demo app still works with the VM setup.

/cc @maasen


scripts/asm-installer/install_asm
```diff
@@ -215,6 +225,10 @@ EOF
     return 0
   fi
 
+  if [[ "${VM_EXPANSION}" -eq 1 ]]; then
```
zerobfd, line 228: Please gate this on enable-gcp-components and otherwise fail if they're not installed

scripts/asm-installer/install_asm
```diff
@@ -539,6 +559,8 @@ FLAGS:
                                       IAM permissions
      --enable_gcp_components          Allow the script to enable required GCP
                                       managed services and components
+     --enable_registration            Allow the script to register the cluster
+                                      as an environ memeber
```
zerobfd, line 563: I think "register a cluster to an environ" is a bit clearer

scripts/asm-installer/install_asm
```diff
@@ -96,6 +97,9 @@ USE_HUB_WIP=0
 ENVIRON_PROJECT_ID=""
 HUB_MEMBERSHIP_ID=""
 
+# Environment variable used by VM
+VM_EXPANSION=0
```
zerobfd, line 101: Can you name it USE_ to be the same as Hub WIP? USE_VMS maybe

scripts/asm-installer/install_asm
zerobfd: In fact, why don't you move this to the block above with init_meshconfig, init_gcp_cas, etc. and then it will be taken care of

scripts/asm-installer/install_asm
```diff
@@ -811,6 +844,10 @@ EOF
     fatal "Hub Workload Identity Pool is only supported for Mesh CA"
   fi
 
+  if [[ "${VM_EXPANSION}" -eq 1 && "${USE_HUB_WIP}" -eq 0 ]]; then
+    fatal "Hub Workload Identity Pool is required to add VM workloads. Run the script with the -o hub-meshca option."
+  fi
```
zerobfd, line 849: Is there a reason we can't just automatically do this instead of making the user re-run the script?

scripts/asm-installer/install_asm
```diff
@@ -1830,6 +1906,11 @@ EOF
     info "...done!"
   fi
 
+  if [[ "${VM_EXPANSION}" -eq 1 ]]; then
+    info "Exposing the control plane for VM workloads..."
+    retry 3 run kubectl apply -f asm/istio/expansion/expose-istiod.yaml
```
zerobfd, line 1911: Please use "${PACKAGE_DIRECTORY}/expansion/expose-istiod.yaml"

scripts/asm-installer/tests/run_basic_suite_vm
```diff
@@ -0,0 +1,133 @@
...
+
+  sleep 5
+
+  echo "Verifying east-west gateway is installed and running..."
+  kubectl wait --for=condition=available --timeout=600s \
+        deployment/istio-eastwestgateway -n istio-system
+
+  echo "Verifying cluster is registered..."
+  if ! is_cluster_registered "${PROJECT_ID}" "${CLUSTER_NAME}" "${CLUSTER_LOCATION}"; then
+    fatal "Cluster is not registered."
```
zerobfd, line 87: Specify "registered to environ"

zerobfd (changes requested):

scripts/asm-installer/install_asm
```diff
@@ -1597,6 +1635,38 @@ get_cluster_labels() {
   echo "${LABELS}"
 }
 
+is_cluster_registered() {
+  local RETVAL; RETVAL=0;
+  kubectl get memberships \
```
maasen, line 1640: Why are we not using the gcloud cmd for this?

scripts/asm-installer/install_asm
```diff
@@ -1597,6 +1635,38 @@ get_cluster_labels() {
...
+    --request-timeout='20s' \
+    membership 1>/dev/null 2>/dev/null || RETVAL=$?
+  return "${RETVAL}"
```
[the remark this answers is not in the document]
jasonwzm: Done. Changed the CRD name to be more specific (also did the same to the other membership check below).

scripts/asm-installer/install_asm
```diff
@@ -1597,6 +1635,38 @@ get_cluster_labels() {
...
+
+  local MEMBERSHIP_NAME
+  local GKE_CLUSTER
+  MEMBERSHIP_NAME="gke-${PROJECT_ID}-${CLUSTER_LOCATION}-${CLUSTER_NAME}"
+  GKE_CLUSTER="${CLUSTER_LOCATION}/${CLUSTER_NAME}"
+  info "Registering the cluster as ${MEMBERSHIP_NAME}..."
+  retry 2 run gcloud beta container hub memberships register "${MEMBERSHIP_NAME}" \
+    --project="${PROJECT_ID}" \
+    --gke-cluster="${GKE_CLUSTER}" \
+    --enable-workload-identity
```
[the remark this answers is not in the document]
jasonwzm: This enables WI on the environ. This is not at cluster level and is gated by --enable-registration.

scripts/asm-installer/install_asm
```diff
@@ -1597,6 +1635,38 @@ get_cluster_labels() {
...
+  return "${RETVAL}"
+}
+
+register_cluster() {
+  if is_cluster_registered; then return; fi
+
+  local MEMBERSHIP_NAME
+  local GKE_CLUSTER
+  MEMBERSHIP_NAME="gke-${PROJECT_ID}-${CLUSTER_LOCATION}-${CLUSTER_NAME}"
+  GKE_CLUSTER="${CLUSTER_LOCATION}/${CLUSTER_NAME}"
```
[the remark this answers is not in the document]
jasonwzm: I made GKE_CLUSTER to use GKE_URI and put it at the beginning of the script. I don't want to put MEMBERSHIP_NAME in a separate function. The other functions that need to use MEMBERSHIP_NAME should use the result from get hub workload identity function.

scripts/asm-installer/install_asm
jasonwzm: This is not enabling a GCP component. It is just registering a CR.

scripts/asm-installer/install_asm
jasonwzm: Done.

scripts/asm-installer/install_asm
jasonwzm: Done. Renamed to USE_VM. VMS sometimes has other representation.

scripts/asm-installer/install_asm
jasonwzm: Done.

scripts/asm-installer/tests/run_basic_suite_vm
jasonwzm: Done.

scripts/asm-installer/install_asm
jasonwzm: We need to know the membership id of the cluster in order to retrieve it deterministically.

zerobfd (commented):
Looks good, after these two issues are resolved I'll approve as long as you tested all of the cases locally to double check.

scripts/asm-installer/install_asm
```diff
@@ -539,6 +561,8 @@ FLAGS:
                                       IAM permissions
      --enable_gcp_components          Allow the script to enable required GCP
                                       managed services and components
+     --enable_registration            Allow the script to register the cluster
```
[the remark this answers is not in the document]
jasonwzm: Done.

jasonwzm:
> How do you handle the case where someone has already registered the cluster to an Environ and not enabled workload identity?

@zerobfd That should be handled by hub-meshca option. Currently we don't validate that case and we basically assume users to have WI. I can create a separate PR to improve that.

zerobfd (approved):
We can merge when tests pass.

jasonwzm:
The test pipeline keeps failing due to cluster creation. VM test succeeded: https://console.cloud.google.com/cloud-build/builds;region=global/8bf40d5e-64dd-4577-8e0a-61d14dd3cd6a;step=9?project=asm-scriptaro-oss

@zerobfd Can we just merge?

zerobfd:
The latest run has a legitimate error in step 8.

jasonwzm:
@zerobfd I added function to generate membership name to truncate the name and avoid conflict. PTAL.

jasonwzm:
Also clean up memberships at the end of the tests. Need to manually clean the memberships before tests can pass.

jasonwzm:
The vm test and a non-vm test passed locally.

PR #364, opened 2021-01-23
Revamp 2

zerobfd:
Add a flag to install_asm that only enables everything without installing anything
Adds script to create a long-term testing cluster if it doesn't already exist, and puts that script
into the cloudbuild pipeline for master and staging
Adds a _CI_REVISION_PREFIX env variable to allow specifying rev name for testing

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -13,6 +13,15 @@ else
   set -u
 fi
 
+### These are hooks for Cloud Build to be able to use debug/staging images
+### when necessary. Don't set these environment variables unless you're testing
+### in CI/CD.
+_CI_ASM_IMAGE_LOCATION="${_CI_ASM_IMAGE_LOCATION:=}"
+_CI_ASM_PKG_LOCATION="${_CI_ASM_PKG_LOCATION:=}"
+_CI_CLOUDRUN_IMAGE_HUB="${_CI_CLOUDRUN_IMAGE_HUB:=}"
+_CI_CLOUDRUN_IMAGE_TAG="${_CI_CLOUDRUN_IMAGE_TAG:=}"
+_CI_REVISION_PREFIX="${_CI_REVISION_PREFIX:=}"
+
 ### Internal variables ###
 MAJOR="${MAJOR:=1}"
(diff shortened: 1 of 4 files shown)


scripts/asm-installer/tests/setup_longterm_cluster
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+set -CeEu
+set -o pipefail
+
+CLUSTER_NAME="long-term-test-cluster"
```
elfinhe, line 6: should we create long term cluster for each branch? Istio revision may share some base resources within the same clusters, and a new minor version istiod may stop working on old base resources.

scripts/asm-installer/install_asm
```diff
@@ -13,6 +13,15 @@ else
...
 fi
 
+### These are hooks for Cloud Build to be able to use debug/staging images
+### when necessary. Don't set these environment variables unless you're testing
+### in CI/CD.
+_CI_ASM_IMAGE_LOCATION="${_CI_ASM_IMAGE_LOCATION:=}"
+_CI_ASM_PKG_LOCATION="${_CI_ASM_PKG_LOCATION:=}"
+_CI_CLOUDRUN_IMAGE_HUB="${_CI_CLOUDRUN_IMAGE_HUB:=}"
+_CI_CLOUDRUN_IMAGE_TAG="${_CI_CLOUDRUN_IMAGE_TAG:=}"
+_CI_REVISION_PREFIX="${_CI_REVISION_PREFIX:=}"
```
elfinhe, line 23: Do we use _CI_REVISION_PREFIX to isolate different test runs, and reuse the cluster?

scripts/asm-installer/tests/setup_longterm_cluster
zerobfd: That's a good catch. Although now I wonder if we should just keep one cluster in order to catch the case where that happens, since I bet we don't have a great idea when it does.

scripts/asm-installer/install_asm
zerobfd: That's the idea.

elfinhe (approved):

PR #370, opened 2021-01-26
Change one test to use long term cluster

zerobfd:


elfinhe (approved):

elfinhe:
I haven't seen the subnet quota issue in Guitar E2E tests. Do we know which operation will create the subnetwork?

zerobfd:
Creating the cluster, and sometimes creating the ingress can cause some networking resources to also get left over and hit quota. It shouldn't happen anymore once all of the tests get converted to use the same cluster.

PR #371, opened 2021-01-26
Add support for MCP logging buckets

douglas-reid:
This PR adds support for creating logs buckets in the consumer project for receiving control plane logs in managed scenarios.

A few notes:
- The logs bucket is created in the same location as the target cluster
- A default name of `asm-control-plane-logs` is used for the bucket. If customization is needed, coordination with the managed service will be needed.

/hold 

This update has been validated in standalone installs, but the managed control plane has not yet been updated to support sending logs to the bucket, so this should not be merged until validated e2e.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -104,6 +104,9 @@ HUB_MEMBERSHIP_ID=""
 # Environment variable used by VM
 USE_VM=0
 
+# Managed Control Plane variables
+LOGS_BUCKET="${LOGS_BUCKET:=asm-control-plane-logs}"
+
 main() {
   init
   parse_args "${@}"
@@ -231,9 +234,11 @@ EOF
   fi
 
   if [[ "${MANAGED}" -eq 1 ]]; then
+    establish_managed_logging_bucket
     start_managed_control_plane
     configure_package
     install_managed_components
+    grant_managed_logging_permissions
     outro
     return 0
   fi
@@ -1975,6 +1980,16 @@ EOF
   fi
 
   outro
+}
+
+establish_managed_logging_bucket() {
+  gcloud beta logging buckets create ${LOGS_BUCKET} --project=${PROJECT_ID} --location=${CLUSTER_LOCATION}
+}
+
+grant_managed_logging_permissions() {
+  gcloud projects add-iam-policy-binding ${PROJECT_ID} \
+    --member=serviceAccount:${MANAGED_SERVICE_ACCOUNT} \
+    --role=roles/logging.bucketWriter
 }
 
 start_managed_control_plane() {
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1982,18 +1982,50 @@ EOF
   outro
 }
 
+pref_len() {
+  local LEN; LEN="${#1}";
+  if ((${#2} > $LEN)); then
+    LEN="${#2}"
+  fi
+
+  for ((i=0; i<$LEN; ++i)); do
+    if [[ "${1:i:1}" != "${2:i:1}" ]]; then
+      echo "${i}"
+      return
+    fi
+  done
+
+  echo "${LEN}"
+}
+
+find_best_region() {
+  REGION="global"
+  BEST=0
+  for location in $(gcloud beta logging locations list --format='value(location_id)'); do
+    SCORE="$(pref_len "${CLUSTER_LOCATION}" "${location}")"
+    if (( $SCORE > $BEST )); then
+      BEST="${SCORE}"
+      REGION="${location}"
+    fi
+  done
+
+  echo ${REGION}
+}
+
+managed_logging_bucket_exists() {
+  BUCKET_LOCATION="$(find_best_region)"
+  BUCKET_COUNT=$(gcloud beta logging buckets list --project="$PROJECT_ID" --location="$BUCKET_LOCATION" --filter="$LOGS_BUCKET" --format=json | jq length)
+
+  if [[ -z "${BUCKET_COUNT}" ]] || [[ "${BUCKET_COUNT}" = '0' ]]; then
+    false
+  fi
+}
+
 establish_managed_logging_bucket() {
-  # The only supported regions are: global, us-central1, us-east1, us-west1, asia-east1, europe-west1.
-  BUCKET_LOCATION="global"
-  case "$CLUSTER_LOCATION" in
-    us-central*    ) BUCKET_LOCATION="us-central1";;
-    us-east*       ) BUCKET_LOCATION="us-east1";;
-    us-west*       ) BUCKET_LOCATION="us-west1";;
-    asia*          ) BUCKET_LOCATION="asia-east1";;
-    europe*        ) BUCKET_LOCATION="europe-west1";;
-    *              ) BUCKET_LOCATION="global";;
-  esac
-  gcloud beta logging buckets create "${LOGS_BUCKET}" --project="${PROJECT_ID}" --location="${BUCKET_LOCATION}"
+  if managed_logging_bucket_exists; then return; fi
+
+  info "creating new logs bucket for managed control plane"
+  gcloud beta logging buckets create "${LOGS_BUCKET}" --project="${PROJECT_ID}" --location="$(find_best_region)"
 }
 
 grant_managed_logging_permissions() {
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1611,7 +1611,7 @@ init_meshconfig_managed() {
   info "Initializing meshconfig managed API..."
   FULL_LOGS_BUCKET_ID="projects/${PROJECT_ID}/locations/$(find_best_region)/buckets/${LOGS_BUCKET}"
   run curl --request POST --fail \
-    --data '{"prepare_istiod": true,"logs_destination":"'${FULL_LOGS_BUCKET_ID}'"}' \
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/install_asm
```diff
@@ -1977,6 +1982,16 @@ EOF
   outro
 }
 
+establish_managed_logging_bucket() {
+  gcloud beta logging buckets create ${LOGS_BUCKET} --project=${PROJECT_ID} --location=${CLUSTER_LOCATION}
+}
+
+grant_managed_logging_permissions() {
+  gcloud projects add-iam-policy-binding ${PROJECT_ID} \
+    --member=serviceAccount:${MANAGED_SERVICE_ACCOUNT} \
```
qfel, line 1991: This is likely the SA used internally by MCP. What you want is the logging service account, which this scripts doesn't know about. We either need to add it to the response from initialize, or infer from MANAGED_SERVICE_ACCOUNT(if possible, seems hacky).

scripts/asm-installer/install_asm
```diff
@@ -1977,6 +1982,16 @@ EOF
   outro
 }
 
+establish_managed_logging_bucket() {
+  gcloud beta logging buckets create ${LOGS_BUCKET} --project=${PROJECT_ID} --location=${CLUSTER_LOCATION}
```
zerobfd, line 1986: Does this require specific IAM permissions to do? What happens if you run this command and the bucket already exists?

zerobfd (changes requested):
In case you can't see the cloud build results, the linter doesn't like that you didn't surround all of your ${VARIABLES} with double quotes "${LIKE_THIS}".

scripts/asm-installer/install_asm
douglas-reid: I don't have access to the MCP projects, so I can't say for sure. Do you use a different account for logging?

Here's an example dump of the perms for that account with these changes:

```
istiod@<REDACTED>-tp.iam.gserviceaccount.com
Compute Admin
Kubernetes Engine Admin
GKE Hub Admin
Service Account Admin
Service Account Key Admin
*Logs Bucket Writer*
Mesh Config Owner
Monitoring Metric Writer
CA Service Admin
Project IAM Admin
Service Management Administrator
Service Usage Admin
Service Usage Consumer
```

It makes _some_ sense that the metrics writer and logs writer be the same, but if there is indeed a different SA, we will need to switch.

scripts/asm-installer/install_asm
douglas-reid: It requires: `logging.buckets.create`.

If you run the command again, you get an error: `ERROR: (gcloud.beta.logging.buckets.create) ALREADY_EXISTS: Bucket asm-control-plane-logs already exists`.

scripts/asm-installer/install_asm
qfel: My reading of the API docs is that this service account will be provided by the logging API, used only to copy logs to the bucket, and has nothing to do with the service account used to write original logs. In other words, I don't think we have any control over what this service account is. Best we just return it from initialize along with the istiod SA.

scripts/asm-installer/install_asm
douglas-reid: What is the preferred way to handle non-idempotent operations in the install script?

zerobfd: Check first and exit if it exists already. See the register_cluster() function for an example.

Could you the appropriate role to the required_iam_roles() function? Ref. https://cloud.google.com/iam/docs/understanding-roles#logging-roles

scripts/asm-installer/install_asm
```diff
@@ -1977,6 +1982,26 @@ EOF
...
+  # The only supported regions are: global, us-central1, us-east1, us-west1, asia-east1, europe-west1.
+  BUCKET_LOCATION="global"
+  case "$CLUSTER_LOCATION" in
+    us-central*    ) BUCKET_LOCATION="us-central1";;
+    us-east*       ) BUCKET_LOCATION="us-east1";;
+    us-west*       ) BUCKET_LOCATION="us-west1";;
+    asia*          ) BUCKET_LOCATION="asia-east1";;
+    europe*        ) BUCKET_LOCATION="europe-west1";;
+    *              ) BUCKET_LOCATION="global";;
+  esac
```
zerobfd, line 1995: This case statement bugged me a bit and I was thinking of a way to avoid it, and then I ended up implementing it to prove to myself that it worked:

```
pref_len() {
  local LEN; LEN="${#1}";
  if ((${#2} > $LEN)); then
    LEN="${#2}"
  fi

  for ((i=0; i<$LEN; ++i)); do
    if [[ "${1:i:1}" != "${2:i:1}" ]]; then
      echo "${i}"
      return
    fi
  done

  echo "${LEN}"
}

find_best_region() {
  REGION="global"
  BEST=0
  for location in $(gcloud beta logging locations list --format='value(location_id)'); do
    SCORE="$(pref_len "${CLUSTER_LOCATION}" "${location}")"
    if (( $SCORE > $BEST )); then
      BEST="${SCORE}"
      REGION="${location}"
    fi
  done

  echo ${REGION}
}
```

Then we can just do `BUCKET_LOCATION="$(find_best_region)"` and drop the case statement, and it also avoids hard coding strings. (And hopefully this script doesn't live that long, but in the case that GCS adds another region we're future proof.)

scripts/asm-installer/install_asm
douglas-reid: OK. I'll await the API change with the appropriate SA before submitting.

scripts/asm-installer/install_asm
douglas-reid: ok. as long as i don't have to maintain that bit of bash-fu :). updated.

scripts/asm-installer/install_asm
douglas-reid: Added a check.

scripts/asm-installer/install_asm
```diff
@@ -1977,6 +1982,58 @@ EOF
...
+  for location in $(gcloud beta logging locations list --format='value(location_id)'); do
+    SCORE="$(pref_len "${CLUSTER_LOCATION}" "${location}")"
+    if (( $SCORE > $BEST )); then
+      BEST="${SCORE}"
+      REGION="${location}"
+    fi
+  done
+
+  echo ${REGION}
+}
```
zerobfd, line 2013: Well now I know I can't write pretty bash:

```
In install_asm line 1987:
  if ((${#2} > $LEN)); then
               ^-- SC2004: $/${} is unnecessary on arithmetic variables.


In install_asm line 1991:
  for ((i=0; i<$LEN; ++i)); do
               ^-- SC2004: $/${} is unnecessary on arithmetic variables.


In install_asm line 2006:
    if (( $SCORE > $BEST )); then
          ^-- SC2004: $/${} is unnecessary on arithmetic variables.
                   ^-- SC2004: $/${} is unnecessary on arithmetic variables.


In install_asm line 2012:
  echo ${REGION}
       ^-- SC2086: Double quote to prevent globbing and word splitting.

```

scripts/asm-installer/install_asm
zerobfd: It looks less scary like this:

```python
def pref_len(a, b):
  len = len(b) len(a) > len(b) else len(a)
  for i in range(len):
    if a[i] != b[i]:
      return i

def find_best_region(target):
  region="global"
  best=0
  for location in locations:
    cur_val = pref_len(location, target)
    if cur_val > best:
      best=cur_val
      region=location

  return location
```

scripts/asm-installer/install_asm
douglas-reid: Updated per these recommendations.

scripts/asm-installer/install_asm
douglas-reid: Updated required roles, validated check.

scripts/asm-installer/install_asm
douglas-reid: Updated logic to match forthcoming API changes.  Will hold to test once endpoint is live.

scripts/asm-installer/install_asm
douglas-reid: case statement replaced.

scripts/asm-installer/install_asm
```diff
@@ -1977,6 +1987,60 @@ EOF
...
+  if [[ -z "${BUCKET_COUNT}" ]] || [[ "${BUCKET_COUNT}" = '0' ]]; then
+    false
+  fi
+}
+
+establish_managed_logging_bucket() {
+  if managed_logging_bucket_exists; then return; fi
+
+  info "creating new logs bucket for managed control plane..."
+  run gcloud beta logging buckets create "${LOGS_BUCKET}" --project="${PROJECT_ID}" --location="$(find_best_region)"
```
cagataygurturk, line 2034: This requires `logging.admin` role which I think should be added to required_iam_roles list

scripts/asm-installer/install_asm
```diff
@@ -234,6 +241,7 @@ EOF
     start_managed_control_plane
     configure_package
     install_managed_components
+    grant_managed_logging_permissions
```
cagataygurturk, line 244: This does not respect the lack of `--enable-gcp-iam-roles` option

```suggestion
    if can_modify_gcp_iam_roles; then
    grant_managed_logging_permissions
    fi
```
probably a warning should be also given to indicate that users should add the permission themselves

douglas-reid:
Thanks for all the reviews. Project shifted direction a bit. Replaced by #555 .

PR #372, opened 2021-01-27
Avoid VM env var override

jasonwzm:
To make sure autoregistration env var is not missed when adding hub-meshca overlay.

This env var will not be needed and will be removed in 1.9 release.


jasonwzm:
Tests were run locally to verify hub-meshca setup does not break when it is used.

elfinhe (commented):
LGTM, need a review from @jtrbs

jtrbs (approved):

elfinhe (approved):

PR #378, opened 2021-01-28
Rebase master-prow on master

elfinhe:


scripts/asm-installer/install_asm
```diff
@@ -681,6 +688,16 @@ parse_args() {
...
+      --trusted_gcp_project | --trusted-gcp-project)
+        arg_required "${@}"
+        TRUSTED_GCP_PROJECTS="${2},${TRUSTED_GCP_PROJECTS}"
+        shift 2
+        ;;
+      --environ_project_number | --environ-project-number)
+        arg_required "${@}"
+        ENVIRON_PROJECT_NUMBER="$(echo "${2}" | tr '[:upper:]' '[:lower:]')"
+        shift 2
+        ;;
```
zerobfd, line 700: Is there any reason these can't also be environment variables? I'd rather not add options if we can help it. We can go back and add trusted projects later when it's supported, and ostensibly we should be able to look up the project number via the project ID.

elfinhe: yep, that makes sense. will add a patch for it. Also changing this parameters will require more changes in prow tests code. But it is worth doing.

elfinhe: ENVIRON_PROJECT_NUMBER is actually required from input to configure multi-project tests in prow. I've moved them into _CI environments to not expose premature configs.

scripts/asm-installer/install_asm
```diff
@@ -1899,13 +1913,34 @@ configure_package() {
...
+${_CI_TRUSTED_GCP_PROJECTS}
+EOF
+    # kpt treats words in quotes as a single param, while kpt need ${TRUST_DOMAIN_ALIASES} to be splitting params for a list. If we remove quotes, the lint will complain.
+    # eval will translate the quoted TRUST_DOMAIN_ALIASES into params to workaround both.
+    run eval kpt cfg set asm anthos.servicemesh.trustDomainAliases "${TRUST_DOMAIN_ALIASES}"
+  fi
+
+  if [[ -n "${_CI_ENVIRON_PROJECT_NUMBER}" ]]; then
+    run kpt cfg set asm gcloud.project.environProjectNumber "${_CI_ENVIRON_PROJECT_NUMBER}"
+  fi
```
zerobfd, line 1943: Can you move this up where the other environProjectNumber setter is so that it's more obvious what gets overwritten?

zerobfd (approved):
One change then lgtm, good to merge as soon as tests pass w/the change

Issue #377, opened 2021-01-28
clusterrolebinding name should be unique


zerobfd:
"cluster-admin-binding" is generic and could overwrite something unintentionally. install_asm should prefix it with the username.

ShrillShrestha:
Hi @zerobfd, I am new to open source and am learning to contribute. I think this can be a good issue for me to work on. If you can guide me and provide a bit more info regarding the issue, I might be able to work on it.

zerobfd:
Sure! In scripts/asm_installer/install_asm there are several functions that handle dependencies. One of them, `bind_user_to_cluster_admin`, creates a clusterrolebinding in the cluster to make the current user an admin so they can install properly.

Right now it always uses the name `cluster-admin-binding` for the k8s object, which means that if you run `install_asm` several times in a row with different users, the clusterrolebinding will get overwritten every time. To prevent this we can prefix the name of the resource with some information about the user. The user is of the form "something@domain", so splitting that on the @ and taking the left side to get the username part of it seems like a reasonable prefix, and then we can use `${USER}-cluster-admin-binding` instead.

ShrillShrestha:
> The user is of the form "something@domain", so splitting that on the @ and taking the left side to get the username part of it seems like a reasonable prefix, 

Are we passing users as a CLI argument? And saving the left part in the `USER` variable.

> and then we can use `${USER}-cluster-admin-binding` instead.

How do we implement it? do I just rename 
https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/ef60c226a13a90442051e1f676b4b78bf8decbb8/scripts/asm-installer/install_asm#L1823

to
`${USER}-cluster-admin-binding`

Please forgive my novice question, I haven't used a bash script much.

zerobfd:
The user information we pull from `gcloud config get-value core/account` at the top of the function. Implementing it can be done using `cut`, something like `PREFIX="$(echo "${GCLOUD_USER}" | cut -f 2 -d @)"` and then change the kubectl arguments to use `clusterrolebinding "${PREFIX}-cluster-admin-binding" \`.

zerobfd:
Thank you!

PR #383, opened 2021-01-29
unique name for clusterrolebinding

ShrillShrestha:
Ref: #377
@zerobfd, I have made some changes. Please let me know if I need to change anything.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1848,8 +1848,9 @@ bind_user_to_cluster_admin(){
   info "Querying for core/account..."
   local GCLOUD_USER; GCLOUD_USER="$(gcloud config get-value core/account)"
   info "Binding ${GCLOUD_USER} to cluster admin role..."
+  PREFIX="$(echo "${GCLOUD_USER}" | cut -f 1 -d @)"
   local YAML; YAML="$(retry 5 kubectl create \
-    clusterrolebinding cluster-admin-binding \
+    clusterrolebinding "${PREFIX}-cluster-admin-binding" \
     --clusterrole=cluster-admin \
     --user="${GCLOUD_USER}" \
     --dry-run -o yaml)"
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/install_asm
```diff
@@ -1848,8 +1848,9 @@ bind_user_to_cluster_admin(){
   info "Querying for core/account..."
   local GCLOUD_USER; GCLOUD_USER="$(gcloud config get-value core/account)"
   info "Binding ${GCLOUD_USER} to cluster admin role..."
+  PREFIX="$(echo "${GCLOUD_USER}" | cut -f 1 -d @)"
```
zerobfd, line 1851: Can you add a `local PREFIX; ` here so that the variable is scoped to this function?

zerobfd (approved):
Looks good! One small change, I'll merge it after it's done and tests pass.

scripts/asm-installer/install_asm
ShrillShrestha: Sure, I have added the `local` keyword. It should be fine now. Thanks, @zerobfd, for helping me out.

scripts/asm-installer/install_asm
```diff
@@ -1848,8 +1848,9 @@ bind_user_to_cluster_admin(){
   info "Querying for core/account..."
   local GCLOUD_USER; GCLOUD_USER="$(gcloud config get-value core/account)"
   info "Binding ${GCLOUD_USER} to cluster admin role..."
+  local PREFIX="$(echo "${GCLOUD_USER}" | cut -f 1 -d @)"
```
zerobfd, line 1851: Please split the `local` declaration and the assignment. e.g. `local VAR; VAR=value`. Otherwise tests won't pass since the linter will block it.

ShrillShrestha: Sorry, I missed that. I made the change.

zerobfd:
No worries! Thanks for the PR, I appreciate it.

PR #388, opened 2021-02-02
Revision based istio-ingressgateway

elfinhe:
I tested with istioctl and kubectl on ingressgateways for a while with different combinations of config structures, and think this is the optimal approach for revision based gateways. Comparing to the original proposal:

- We don't separate the gateways from their original yamls because that introduces one more jump/mapping from an install option to its gateways.
- We still use a single istioctl install command to install a new revision of ingressgateway, istiod, and other resources. But now each revision ingressgateway will get a separate deployment, naming  istio-ingressgateway-ASM_REV.
- We will put the command of switching ingressgateway revisions into our upgrade guide, and let user decide when to change revision to be used. And in the rollback process, user will get a *symmetric* command to change it back.
- We will recommend users to use the new revision of ingressgateway before re-inject workloads. After all workloads get re-injected, there won't be any long lived connections through the old ingressgateway, so it will be safe to cleanup all old revision resources including ingressgateway deployment without disrupting traffic. 

```
kubectl patch service -n istio-system istio-ingressgateway --type='json' -p='[{"op": "replace", "path": "/spec/selector/service.istio.io~1canonical-revision", "value": "<ASM_REV>"}]'
```

If this looks good, I will create followup CRs (or notify the owners) to apply this method to other gateways in options.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -2020,7 +2020,7 @@ EOF
 
   info "Installing ASM control plane..."
   # shellcheck disable=SC2086
-  retry 5 run "$(istioctl_path)" install --skip-confirmation $PARAMS
+  retry 5 run "$(istioctl_path)" install $PARAMS
 
   # Prevent the stderr buffer from ^ messing up the terminal output below
   sleep 1
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -317,3 +317,11 @@ openAPI:
           name: anthos.servicemesh.external_ca.ca_name
           value: ""
           isSet: true
+    io.k8s.cli.substitutions.istio-ingressgateway-deploy-name:
+      x-k8s-cli:
+        substitution:
+          name: istio-ingressgateway-deploy-name
+          pattern: istio-ingressgateway-ASM_REV
+          values:
+          - marker: ASM_REV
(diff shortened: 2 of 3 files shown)


scripts/asm-installer/install_asm
```diff
@@ -2020,7 +2020,7 @@ EOF
 
   info "Installing ASM control plane..."
   # shellcheck disable=SC2086
-  retry 5 run "$(istioctl_path)" install --skip-confirmation $PARAMS
```
zerobfd, line 2023: Was this on purpose? If so can you note why?

zerobfd (changes requested):
Can we make this an overlay itself instead of putting it in the operator? We can still have install_asm default to using it if we want, but I'd rather have the option to toggle it with just config.

scripts/asm-installer/install_asm
elfinhe: func istioctl_path will decide whether or not to use the _CI_ISTIOCTL_REL_PATH, we can leave this change into a separate PR, so we can discuss more there.

zerobfd: I meant getting rid of the `--skip-confirmation` flag. I agree with the variable -> function change.

zerobfd (approved):
Thanks!

PR #389, opened 2021-02-02
Install base client-side for MCP

zerobfd:
Also adds default tag in istio-operator for manual installation


elfinhe (commented):
Kptfile also needs to be updated.

The command can catch all occurrences: 
`kpt cfg set asm anthos.servicemesh.tag 1.8.2-asm.2`

zerobfd:
That's pretty clever, I never would have thought to use kpt to change the config in the repo. Thanks!

elfinhe (approved):

PR #390, opened 2021-02-02
1.8.2

zerobfd:


asm/istio/istio-operator.yaml
```diff
@@ -11,16 +11,56 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+#
 apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
 spec:
-  profile: asm-gcp
+  profile: empty
```
richardwxn, line 18: Did we remove the asm-gcp profile?

richardwxn (approved):

asm/istio/istio-operator.yaml
zerobfd: It's not removed, it's still in the binary and will still work. This operator isn't going to use it anymore. We had too many cases of the profile setting some option that was unexpected so we're explicitly setting everything ourselves.

elfinhe (approved):

asm/istio/istio-operator.yaml
```diff
@@ -11,16 +11,56 @@
...
+    egressGateways:
+    - enabled: false
+      name: istio-egressgateway
+      k8s:
+        hpaSpec:
+          maxReplicas: 5
+          minReplicas: 2
+    ingressGateways:
+    - enabled: true
+      name: istio-ingressgateway
```
elfinhe, line 48: it is quite tricky. But the default profile has one more port than the charts only in istio-ingressgateway.

https://github.com/istio/istio/blob/release-1.8/manifests/profiles/default.yaml#L70-L72, which is used for mesh expansion IIRC. We'd better still have them in the yaml:

```

    - name: istio-ingressgateway-181-5
      enabled: true
      k8s:
        env:
          - name: ISTIO_META_ROUTER_MODE
            value: "standard"
        service:
          selector:
            app: istio-ingressgateway
            istio: ingressgateway
            service.istio.io/canonical-revision: asm-181-5
          ports:
            - port: 15021
              targetPort: 15021
              name: status-port
              protocol: TCP
            - port: 80
              targetPort: 8080
              name: http2
              protocol: TCP
            - port: 443
              targetPort: 8443
              name: https
              protocol: TCP
            - port: 15012
              targetPort: 15012
              name: tcp-istiod
              protocol: TCP
            - port: 15443
              targetPort: 15443
              name: tls
              protocol: TCP
```

elfinhe: or we can still use asm-gcp profile in 1.8, changing to empty profile is too much for a patch version.

elfinhe (changes requested):
We need think twice before changing to empty profile.

zerobfd:
`istioctl profile dump` generates identical output for the `istio-operator.yaml` file regardless of whether the profile is `asm-gcp` or `empty`. (Tested with

```
/tmp/tmp.Hzgono5vED/istio-1.8.1-asm.5/bin/istioctl profile dump -f ~/repos/anthos-service-mesh-packages/asm/istio/istio-operator.yaml > empty.yaml
sed -i 's/empty/asm-gcp/g' ~/repos/anthos-service-mesh-packages/asm/istio/istio-operator.yaml
/tmp/tmp.Hzgono5vED/istio-1.8.1-asm.5/bin/istioctl profile dump -f ~/repos/anthos-service-mesh-packages/asm/istio/istio-operator.yaml > gcp.yaml
diff empty.yaml gcp.yaml # no ouput
```

)

Is there something that could still affect the behavior?

elfinhe (approved):

elfinhe:
> `istioctl profile dump` generates identical output for the `istio-operator.yaml` file regardless of whether the profile is `asm-gcp` or `empty`. (Tested with
> 
> ```
> /tmp/tmp.Hzgono5vED/istio-1.8.1-asm.5/bin/istioctl profile dump -f ~/repos/anthos-service-mesh-packages/asm/istio/istio-operator.yaml > empty.yaml
> sed -i 's/empty/asm-gcp/g' ~/repos/anthos-service-mesh-packages/asm/istio/istio-operator.yaml
> /tmp/tmp.Hzgono5vED/istio-1.8.1-asm.5/bin/istioctl profile dump -f ~/repos/anthos-service-mesh-packages/asm/istio/istio-operator.yaml > gcp.yaml
> diff empty.yaml gcp.yaml # no ouput
> ```
> 
> )
> 
> Is there something that could still affect the behavior?

nope, looks good! thanks Clay.

asm/istio/istio-operator.yaml
elfinhe: this one additional port and ISTIO_META_ROUTER_MODE are inside the charts and only applies to igw with name `istio-ingressgateway`, so won't affect this change.

jasonwzm (approved):

PR #393, opened 2021-02-03
Add CLOUDRUN_ADDR for managed control plane

ruigulala:
For Gateway, a new env (ISTIO_META_CLOUDRUN_ADDR) is needed.
The value of this env is obtained from mutatingwebhook after
MCP installation.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -2089,9 +2089,12 @@ install_managed_components() {
   local URL
   URL="$(kubectl get mutatingwebhookconfiguration istiod-asm-managed -ojson | jq .webhooks[0].clientConfig.url -r)"
   # shellcheck disable=SC2001
-  URL="$(echo "${URL}" | sed 's/inject.*$/validate/g')"
+  VALIDATION_URL="$(echo "${URL}" | sed 's/inject.*$/validate/g')"
+  # shellcheck disable=SC2001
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/install_asm
```diff
@@ -2084,8 +2084,10 @@ install_managed_components() {
   URL="$(kubectl get mutatingwebhookconfiguration istiod-asm-managed -ojson | jq .webhooks[0].clientConfig.url -r)"
   # shellcheck disable=SC2001
   URL="$(echo "${URL}" | sed 's/inject.*$/validate/g')"
+  CLOUDRUN_ADDR=$(kubectl get mutatingwebhookconfigurations istiod-asm-managed -ojson | jq ".webhooks[0].clientConfig.url" | cut -d'/' -f3)
 
   run kpt cfg set asm anthos.servicemesh.controlplane.validation-url "${URL}"
+  run kpt cfg set asm anthos.servicemesh.managed-controlplane.cloudrun-addr "${CLOUDRUN_ADDR}"
```
zerobfd, line 2090: The first $URL already has this, change the second URL to VALIDATION_URL and then have CLOUDRUN_ADDR use $URL instead of querying the cluster again

zerobfd (changes requested):

scripts/asm-installer/install_asm
ruigulala: Done.

zerobfd (approved):

PR #402, opened 2021-02-03
[VM Script] Update create instance template logic

jasonwzm:
The PR has the following updates:
- Cluster registration validation
- WorkloadGroup validation
  - ASM revision is labeled correctly
  - Cluster has the correct identity provider
  - Workload is using an existing service account
- Retrieve eastwest gateway IP
- Retrieve cluster network
- Retrieve identity provider

Adding one basic test to test the instance template creation.

To reduce review burden, more tests will be added in following PRs.


jasonwzm:
@zerobfd WDYT putting VM tests at the same directory level as other tests to better use the common library? Or do you prefer in different directory.

scripts/asm-installer/tests/common.sh
```diff
@@ -665,3 +670,55 @@ run_basic_test() {
 
   return "$SUCCESS"
 }
+
+create_service_account() {
+  WORKLOAD_SERVICE_ACCOUNT="vm-${LT_NAMESPACE}@${PROJECT_ID}.iam.gserviceaccount.com"
```
zerobfd, line 675: Don't put the vm- prefix here, do it when you call the function. Also there's nothing specific to WORKLOAD here, just call this SERVICE_ACCOUNT (or SA)

scripts/asm-installer/tests/common.sh
```diff
@@ -665,3 +670,55 @@ run_basic_test() {
...
+}
+
+verify_instance_template() {
+  # TODO(jasonwzm): include more sophisticated test for proxy config.
+  local VAL
+  VAL="$(gcloud compute instance-templates list \
+    --filter="name=${INSTANCE_TEMPLATE_NAME}" --format="value(name)")"
+  if [[ -z "${VAL}" ]]; then
+    fail "Cannot find instance template ${INSTANCE_TEMPLATE_NAME} in the project."
+    return 1
```
zerobfd, line 710: Change `return 1` to `false`

scripts/asm-installer/tests/run_basic_instance_template_creation
```diff
@@ -0,0 +1,51 @@
...
+# shellcheck source=common.sh
+. "${SDIR}/common.sh"
+
+cd "${SDIR}"
+
+main() {
+  # CLI setup
+  parse_args "$@"
+
+  run_basic_test "install" "mesh_ca" "-o vm -o hub-meshca --enable-registration --enable-gcp-components"
```
zerobfd, line 18: Why are these --enable flags necessary?

asm/vm/asm_vm
```diff
@@ -207,53 +222,57 @@ All options can also be passed via environment variables by using the ALL_CAPS
...
-                                                    passed.
+  -l|--cluster_location         <LOCATION>            The GCP location of the
+                                                      target cluster.
+  -n|--cluster_name             <NAME>                The name of the target
+                                                      cluster.
+  -p|--project_id               <ID>                  The GCP project ID.
+  -w|--workload_name            <WORKLOAD_NAME>       The name of the
+                                                      WorkloadGroup for GCE VM
+                                                      workloads.
+  -ns|--workload_namespace      <WORKLOAD_NAMESPACE>  (optional)The namespace of
```
zerobfd, line 237: It's very non-standard to have a two character flag after a single dash. I don't have a good suggestion for what it should be, but if you can't think of one just remove the single-dash option and make it `--workload_namespace` only.

asm/vm/asm_vm
```diff
@@ -518,62 +536,203 @@ EOF
   fi
 }
 
-validate_asm_installation() {
+validate_retrieve_asm_installation() {
```
zerobfd, line 539: What does this function name mean?

asm/vm/asm_vm
```diff
@@ -518,62 +536,203 @@ EOF
...
+retrieve_workload_asm_revision() {
+  info "Retrieving the ASM revision for the WorkloadGroup..."
+  WORKLOAD_ASM_REVISION="$(retry 2 kubectl get namespace "${WORKLOAD_NAMESPACE}" -ojson \
+    | jq -r '.metadata.labels["'${ASM_REVISION_LABEL_KEY}'"]')"
+  if [[ -z "${WORKLOAD_ASM_REVISION}" ]]; then
+    { read -r -d '' MSG; fatal "${MSG}"; } <<EOF || true
+Unable to find an ASM revision for the namespace of the WorkloadGroup. To apply
+a revision label on the namespace, replace <REVISION> with an available revision
+name and run:
+  kubectl label namespace "${WORKLOAD_NAMESPACE}" istio-injection- \
```
zerobfd, line 675: These double quotes will be printed literally since you're in a heredoc.

asm/vm/asm_vm
```diff
@@ -677,27 +836,62 @@ retrieve_cluster_root_cert() {
...
+        MEMBERSHIP_UID="$(retry 2 run gcloud container hub memberships describe \
+          "${MEMBERSHIP}" --project "${PROJECT_ID}" --format="value(uniqueId)")"
+      fi
+    fi
+  done <<EOF
+${MEMBERSHIPS}
+EOF
+}
+
 # Generate GCE service proxy definition for the instance template
```
zerobfd, line 895: You don't break after you find what you're looking for. It looks like you might find multiple UIDs but only set the latest one. Can you return early? Also, is there a reason not to just set CREDENTIAL_IDENTITY_PROVIDER in the same function?

scripts/asm-installer/cloudbuild-master.yaml
```diff
@@ -144,6 +144,24 @@ steps:
   - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
   timeout: 600s
 
+- name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
```
zerobfd, line 147: Can you update the lint rule to also look at asm/vm/asm_vm? Also, since you're now creating Cloud resources, you need to have some sort of logic to clean up ones that leak.

zerobfd (changes requested):

scripts/asm-installer/tests/common.sh
jasonwzm: > Don't put the vm- prefix here, do it when you call the function. 
I don't understand the benefit here. I don't want to only add it in one place but forget to add it to other places.
> Also there's nothing specific to WORKLOAD here, just call this SERVICE_ACCOUNT (or SA)
This is very specific to workload. Renamed to workload_service_account.

scripts/asm-installer/tests/common.sh
jasonwzm: Done.

scripts/asm-installer/cloudbuild-master.yaml
jasonwzm: Done.

asm/vm/asm_vm
jasonwzm: Good point. break is added. But it will not have more than one UIDs returned, break is making it more efficient.

> Also, is there a reason not to just set CREDENTIAL_IDENTITY_PROVIDER in the same function?

Not really, just separation of logic. But combined.

asm/vm/asm_vm
jasonwzm: Removed.

asm/vm/asm_vm
jasonwzm: It also retrieves the asm revisions. But renamed.

asm/vm/asm_vm
jasonwzm: Done. Removed.

scripts/asm-installer/tests/run_basic_instance_template_creation
jasonwzm: Yea, i see it is not needed. Removed from VM control plane test as well.

scripts/asm-installer/cloudbuild-master.yaml
jasonwzm: > Also, since you're now creating Cloud resources, you need to have some sort of logic to clean up ones that leak.

The test script cleans up leaked objects. Is there somewhere else that also cleans up stuff?

scripts/asm-installer/tests/common.sh
zerobfd: I mean that there's nothing specific to workloads with the gcloud command you're calling.

If you want to do it like this, I'd rather you have a generic create_service_account() function, and then you can wrap that with create_workload_service_account() where you add the vm- and whatever else. At some point we're going to have to test permissions and create service accounts, and whoever adds that logic is just going to have to pull the logic apart anyways.

I don't care enough to block the PR on it if you don't want to put it in this PR though.

scripts/asm-installer/cloudbuild-master.yaml
zerobfd: You're creating service accounts and instance templates in the test and the script might error and close before it reaches the end.

zerobfd: Sorry, I misread your comment at first. It's in the setup_longterm_cluster script. That way things that leak will get cleaned up eventually on the next test run.

scripts/asm-installer/tests/run_basic_instance_template_creation
zerobfd: Thanks!

asm/vm/asm_vm
```diff
@@ -16,27 +16,47 @@ fi
...
 INGRESS_IP=""
-MESH_AGENT_BUCKET=\
-"gs://gce-mesh/service-proxy-agent/releases/service-proxy-agent-0.2-asm.tgz"
-INSTANCE_IPS=""
+ASM_REVISIONS=""
+NETWORK=""
+MEMBERSHIP_UID=""
+GKE_URL=""
+ENVIRON_PROJECT_ID=""
+HUB_MEMBERSHIP_ID=""
```
maasen, line 49: Can we shift to referencing Environ instead of Hub?

asm/vm/asm_vm
```diff
@@ -207,53 +224,57 @@ All options can also be passed via environment variables by using the ALL_CAPS
...
-                                                    passed.
+  -l|--cluster_location         <LOCATION>            The GCP location of the
+                                                      target cluster.
+  -n|--cluster_name             <NAME>                The name of the target
+                                                      cluster.
+  -p|--project_id               <ID>                  The GCP project ID.
+  -w|--workload_name            <WORKLOAD_NAME>       The name of the
+                                                      WorkloadGroup for GCE VM
+                                                      workloads.
+  --workload_namespace          <WORKLOAD_NAMESPACE>  (optional)The namespace of
```
maasen, line 239: nit - space after '(optional)' and possibly ident to match other flags

scripts/asm-installer/tests/common.sh
jasonwzm: sgtm. I extract create_service_account out for now.

scripts/asm-installer/cloudbuild-master.yaml
jasonwzm: Added cleanup all vm instance template and cleanup all workload service accounts to the build step.

asm/vm/asm_vm
jasonwzm: Done.

asm/vm/asm_vm
jasonwzm: Done.

asm/vm/asm_vm
zerobfd: Agree about the rename, but I'd rather not do it here. It needs to be changed in install_asm (and possible other places) as well, and I'd rather have it all in one PR that doesn't change functionality. There's a lot in this PR already.

asm/vm/asm_vm
```diff
@@ -16,27 +16,47 @@ fi
...
 INGRESS_IP=""
-MESH_AGENT_BUCKET=\
-"gs://gce-mesh/service-proxy-agent/releases/service-proxy-agent-0.2-asm.tgz"
-INSTANCE_IPS=""
+ASM_REVISIONS=""
+NETWORK=""
+MEMBERSHIP_UID=""
+GKE_URL=""
+ENVIRON_PROJECT_ID=""
+ENVIRON_MEMBERSHIP_ID=""
```
zerobfd, line 49: Please change back to HUB for this PR re: other comment thread

zerobfd (approved):
LGTM as soon as tests pass if the only thing left are test fixes. Re-request review if the logic materially changes.

asm/vm/asm_vm
jasonwzm: I am removing this membership id variable. This is not needed by VM script.

PR #405, opened 2021-02-03
Work around istioctl bug with multi component yaml files

zerobfd:
If `csplit` generates more than one file when splitting on the yaml document separator, then remove the yaml file from the list of files to pass to `istioctl` and instead add the split yaml files. Verified that the new test fails before adding the fix.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1116,7 +1116,9 @@ handle_multi_yaml_bug() {
   done <<EOF
 ${CUSTOM_OVERLAY}
 EOF
-  CUSTOM_OVERLAY="${CSPLIT_OUTPUT:1:${#CSPLIT_OUTPUT}},"
+  if [[ -n "${CSPLIT_OUTPUT}" ]]; then
+    CUSTOM_OVERLAY="${CSPLIT_OUTPUT:1:${#CSPLIT_OUTPUT}},"
+  fi
 }
 
 validate_cli_dependencies() {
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/install_asm
```diff
@@ -1094,9 +1094,31 @@ EOF
...
 
+# This is a workaround for https://github.com/istio/istio/issues/30632
+# which doesn't handle files with multiple operator specs correctly.
+# This will split all of the files based off of the yaml document separator.
+handle_multi_yaml_bug() {
+  local CSPLIT_OUTPUT; CSPLIT_OUTPUT="";
+  local BASE_NAME
+  while read -d ',' -r yaml_file; do
+    BASE_NAME="$(basename "${yaml_file}")"
+    if [[ "$(csplit -z -f "overlay-${BASE_NAME}" -z "${yaml_file}" '/^---$/' '{*}' | wc -l)" -eq 1 ]]; then
```
elfinhe, line 1109: not a blocker, but csplit is not in alpine.

```
$ docker run -ti alpine csplit
Wed 03 Feb 2021 01:34:40 PM PST: executing "docker run -ti alpine csplit"
docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "csplit": executable file not found in $PATH: unknown.
ERRO[0000] error waiting for container: context canceled
```

elfinhe (approved):

scripts/asm-installer/install_asm
zerobfd: Good point. `csplit` is in coreutils and we already have dependencies there so I don't think this is a regression, but I added it to the dependency list to make it clearer.

PR #413, opened 2021-02-05
Allow cluster to be registered in a different project

jasonwzm:
/cc @chizhg

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1750,21 +1750,15 @@ is_cluster_registered() {
 
   populate_environ_info
 
-  if [[ "${PROJECT_ID}" == "${ENVIRON_PROJECT_ID}" ]]; then
-    local WANT
-    WANT="//container.googleapis.com/projects/${PROJECT_ID}/locations/${CLUSTER_LOCATION}/clusters/${CLUSTER_NAME}"
-    local LIST
-    LIST="$(gcloud container hub memberships list --project "${PROJECT_ID}" \
-      --format=json | grep "${WANT}")"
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/install_asm
```diff
@@ -1735,28 +1735,38 @@ get_cluster_labels() {
...
-  if [[ -z "${IDENTITY_PROVIDER}" ]] || [[ "${IDENTITY_PROVIDER}" = 'null' ]]; then
+  if [[ -z "${IDENTITY_PROVIDER}" ]] || [[ "${IDENTITY_PROVIDER}" == 'null' ]]; then
     false
   fi
+
+  populate_environ_info
+
+  if [[ "${PROJECT_ID}" == "${ENVIRON_PROJECT_ID}" ]]; then
+    local WANT
+    WANT="//container.googleapis.com/projects/${PROJECT_ID}/locations/${CLUSTER_LOCATION}/clusters/${CLUSTER_NAME}"
```
zerobfd, line 1755: Can you just replace PROJECT_ID with ENVIRON_PROJECT_ID and remove the `--project "${PROJECT_ID}"` flag?

jasonwzm: How about permission check? Try to check membership, if permission issue, then warn?

zerobfd: Makes sense to me. We can just make it a general "Failed to validate environ registration" warning.

jasonwzm: done.

zerobfd (approved):

PR #416, opened 2021-02-05
Wrap commands better

zerobfd:
Instead of having to remember to put `run` everywhere, just override the commands that could have side effects and wrap them with whatever safety mechanisms we want. This fixes a bug in dry-run that incorrectly causes clusters to be considered unreachable. Also adds a small qol change to testing to allow easier local tests without editing the file,

### DIFF
--- a/scripts/asm-installer/cloudbuild-master.yaml
+++ b/scripts/asm-installer/cloudbuild-master.yaml
@@ -10,13 +10,14 @@ steps:
   dir: 'scripts/asm-installer'
   id: 'publish-tester-image'
   args: ['push', 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}']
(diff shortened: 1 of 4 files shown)


elfinhe (approved):
lint complaint: 
`-- SC2230: which is non-standard. Use builtin 'command -v' instead.`

which i kinda disagree, but we will have to follow what is suggested there.

ruigulala (approved):

zerobfd:
That's weird, my local shellcheck (v0.7.1) doesn't say that. We seem to use an old version (0.5.0-3) in the docker container. https://github.com/koalaman/shellcheck/wiki/SC2230 lists is as opt-in for new versions, so I'm going to ignore it :) I'll fix the cloudbuild yaml to not do that from now on.

PR #421, opened 2021-02-06
MCP improvements

### LINKED ISSUE #419: When --managed is enabled, can_modify_gcp_iam_roles always returns true
https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/3b87e4e176790f1eafdb033726b435a953ea5e7f/scripts/asm-installer/install_asm#L427-L431

This makes impossible to run the script if the user prefers setting IAM rules on her own.

Modifying the function as follows solves the issue.

```
can_modify_gcp_iam_roles() {
  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_GCP_IAM_ROLES}" -eq 0 ]] || \
    ! can_modify_at_all; then false; fi
}
```

### LINKED ISSUE #420: Too many roles are assigned on istiod service account
The following line

https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/8337ba1608557b30a8e5b549c9f6089a08b74acb/scripts/asm-installer/install_asm#L190

attempts to add the following IAM roles on the managed istiod service account:

```
roles/servicemanagement.admin
roles/serviceusage.serviceUsageAdmin
roles/meshconfig.admin
roles/compute.admin
roles/container.admin
roles/resourcemanager.projectIamAdmin
roles/iam.serviceAccountAdmin
roles/iam.serviceAccountKeyAdmin
roles/gkehub.admin
roles/privateca.admin
```

These roles are the same roles that the local IAM user needs to do many things. I highly doubt the managed istiod SA requires such extensive roles on user's project like `resourcemanager.projectIamAdmin` or `compute.admin`. In fact I was able to verify that without at least `resourcemanager.projectIamAdmin`, things work fine. Can these roles be reviewed?

cagataygurturk:
- Assign or validate only the required set of roles for Istiod SA (Fixes #420)
- Address and issue with disabling IAM modifications option upon managed control plane installation (Fixes #419)

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -187,7 +187,7 @@ main() {
   fi
 
   if can_modify_gcp_iam_roles; then
-    bind_user_to_iam_policy "$(required_iam_roles)" "$(iam_user)"
+    bind_user_to_iam_policy "$(required_iam_roles_istiod_service_account)" "$(iam_user)"
   elif should_validate; then
     exit_if_out_of_iam_policy
   fi
@@ -491,8 +491,7 @@ can_modify_gcp_apis() {
 }
 
 can_modify_gcp_iam_roles() {
-  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_GCP_IAM_ROLES}" -eq 0 ]] && \
-    ! is_managed ||
+  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_GCP_IAM_ROLES}" -eq 0 ]] ||
     ! can_modify_at_all; then false; fi
 }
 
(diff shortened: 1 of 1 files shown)


zerobfd:
Thank you for the PR! In general, I don't think the set of permissions needed by the service account used by managed control plane and the user installing the script is the same in all cases. I realize the script is less than straightforward and I don't want to make you go through several iterations to get the PR approved. Do you mind if I close this PR and put up another one to fix the issues? If you'd rather do it yourself, I'm glad to review as many times as necessary.

cagataygurturk:
Yes feel free to close this or any other PRs I open and implement it properly. I don‘t feel super competent with Bash 😊

zerobfd:
There was nothing wrong with your bash :) it's us, not you.

cagataygurturk:
Let‘s implement all this in anthoscli in Go 😉

PR #422, opened 2021-02-07
Remove cluster admin binding

cagataygurturk:
The script is requiring `container.admin` role which automatically grants the user full API access in GKE. If I am not wrong with this assumption, making the user cluster admin again seems unnecessary. In my trials without this clusterrolebinding the installation went through without any problem.

So this PR is removing that piece if you want to simply the script a bit.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -215,12 +215,6 @@ main() {
     fi
   fi
 
-  if can_modify_cluster_roles; then
-    bind_user_to_cluster_admin
-  elif should_validate; then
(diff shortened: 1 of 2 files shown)


zerobfd:
Sorry, somehow missed responding to this one. Two pieces: one, there's a larger effort trying to reduce the IAM roles that will help for the GKE case. Two, we can't remove it outright (yet) because for off-GKE clusters we'll need to use cluster roles instead of IAM.

Feel free to open an issue if you want to track a specific piece of this. Or for anything else you come up with :)

PR #424, opened 2021-02-08
[VM Script] Add more tests

jasonwzm:
Added more tests to
- verify the fields in the instance template
- verify instance template can be created based on an existing template


zerobfd (approved):
LGTM as long as tests pass.

PR #426, opened 2021-02-08
Don't register with the --enable-all flag

zerobfd:


jasonwzm (changes requested):
The description of --enable_all needs to be change to exclude --enable_registration

zerobfd:
I'll update the PR after the tests finish so I don't need to wait for them to run again. Here's the proposed change:

```
-  every necessary action, pass the -e|--enable_all flag. All of these flags
-  are incompatible with --only_validate.
+  every necessary action, pass the -e|--enable_all flag. --enable_registration
+  is not included in --enable_all and must be passed separately.
+  All of these flags are incompatible with --only_validate.
 
   -e|--enable_all                     Allow the script to perform all of the
                                       individual enable actions below.
@@ -586,8 +587,9 @@ FLAGS:
                                       IAM permissions
      --enable_gcp_components          Allow the script to enable required GCP
                                       managed services and components
+
      --enable_registration            Allow the script to register the cluster
-                                      to an environ
+                                      to an environ (experimental)
```

jasonwzm (approved):
Thanks!

PR #429, opened 2021-02-08
Add support for custom ingressgateway

elfinhe:
Add support for custom ingressgateway

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -331,3 +331,26 @@ openAPI:
           values:
           - marker: ASM_REV
             ref: '#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev'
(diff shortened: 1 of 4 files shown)


zerobfd (approved):
Consider renaming to custom-revisioned-ingressgateway just in case? Also it dawns on me that we don't have a great way to allow specifying multiple of a single config...

elfinhe:
> Consider renaming to custom-revisioned-ingressgateway just in case? Also it dawns on me that we don't have a great way to allow specifying multiple of a single config...

yes, the current API structure of only supporting a list of ingressgateway makes it somehow not easy to manage multiple ingressgateway. I was thinking let users reinstall per each gateway (after use kpt set the name and namespace).

PR #434, opened 2021-02-09
Fix option flags being ignored issue when reinstalling on an existing folder

elfinhe:
When reinstalling with an existing folder, options are not converted into yamls in CUSTOM_OVERLAY. This PR should fix it.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -146,7 +146,10 @@ main() {
 
   if needs_asm && ! necessary_files_exist; then
     download_asm
-    validate_kpt_files
+  fi
+
+  if needs_asm ; then
+    organize_kpt_files
   fi
 
   if should_validate && ! is_managed; then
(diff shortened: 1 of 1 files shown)


elfinhe:
Need to cherry-pick this into 1.8 for both staging and release.

scripts/asm-installer/install_asm
```diff
@@ -146,7 +146,10 @@ main() {
 
   if needs_asm && ! necessary_files_exist; then
     download_asm
-    validate_kpt_files
+  fi
+
+  if needs_asm ; then
```
zerobfd, line 151: Can you remove the space between asm and the ; ?

scripts/asm-installer/install_asm
```diff
@@ -1140,13 +1143,13 @@ validate_dependencies() {
   fi
 }
 
-validate_kpt_files () {
+organize_kpt_files () {
```
zerobfd, line 1146: Not sure how it got there, but can you remove the space between files and ( ?

zerobfd (approved):
Nice catch. If you `git diff` after the nits and make sure that the two spaces are the only changes you can just push straight to master, no point in waiting for the tests again.

PR #440, opened 2021-02-09
Setup different cluster for environ-based installations

jasonwzm:


scripts/asm-installer/cloudbuild-master.yaml
```diff
@@ -51,6 +51,24 @@ steps:
...
+- name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
+  dir: 'scripts/asm-installer'
+  id: 'setup-long-term-cluster-for-environ-test'
+  entrypoint: '/bin/bash'
+  args:
+  - '-c'
+  - >
+    ./tests/setup_longterm_cluster
+  waitFor:
+  - 'fetch-secrets'
```
zerobfd, line 63: This file doesn't use waitFor, all of the tests are sequential.

scripts/asm-installer/cloudbuild-master.yaml
```diff
@@ -146,6 +164,47 @@ steps:
...
+  id: 'run-basic-suite-vm'
+  entrypoint: '/bin/bash'
+  args:
+  - '-c'
+  - >
+    ./tests/run_basic_suite_vm
+    --PROJECT_ID "${PROJECT_ID}"
+    --BUILD_ID "${BUILD_ID}"
+  waitFor:
+  - 'setup-long-term-cluster-for-environ-test'
```
zerobfd, line 178: This file doesn't use waitFor, all of the tests are sequential.

zerobfd (changes requested):
Instead of doing it this way, can you pull the current logic into a function and keep it in the same  script? That will keep the cloudbuild.yaml files from getting larger and also lower the chance that it might get left out somewhere else.

jasonwzm:
> Instead of doing it this way, can you pull the current logic into a function and keep it in the same script?

Can you be more specific? Do you mean setting up two clusters in the same script and just call once from cloud build file?

scripts/asm-installer/cloudbuild-master.yaml
jasonwzm: What is the reason for no waitFor if they are setting up/running different clusters?

zerobfd: The reason why we changed to running on one cluster in the first place is because cluster creation/IAM operations would interfere with each other or time out.

zerobfd:
Yea, setting them both up in the same script and having one cloudbuild entry

scripts/asm-installer/cloudbuild-master.yaml
jasonwzm: I don't think with two clusters running tests separately, there should be operations conflicting each other. But for now, I remove that as it is not in scope of what this PR is trying to do.

jasonwzm:
> Yea, setting them both up in the same script and having one cloudbuild entry

Done. PTAL.

scripts/asm-installer/tests/setup_longterm_cluster
```diff
@@ -9,71 +9,78 @@ SDIR="$(dirname "${SPATH}")"
...
-    --no-enable-master-authorized-networks \
-    --addons HorizontalPodAutoscaling,HttpLoadBalancing \
-    --enable-autoupgrade \
-    --enable-autorepair \
-    --max-surge-upgrade 1 \
-    --max-unavailable-upgrade 0 \
-    --workload-pool "${LT_PROJECT_ID}.svc.id.goog" \
-    --enable-shielded-nodes
-fi
+  # TODO: Not depend on a cluster to retrieve ASM package.
```
zerobfd, line 25: Please remove this.

scripts/asm-installer/tests/setup_longterm_cluster
```diff
@@ -9,71 +9,78 @@ SDIR="$(dirname "${SPATH}")"
 . "${SDIR}/common.sh"
 cd "${SDIR}"
 
-LT_CLUSTER_NAME="long-term-test-cluster"
-LT_CLUSTER_LOCATION="us-central1-c"
-LT_PROJECT_ID="asm-scriptaro-oss"
+EXPECTED_CLUSTER_NAMES="$1"
```
zerobfd, line 12: I think you're making this too complicated. Just make an LT_CLUSTER_NAME and LT_ENVIRON_CLUSTER_NAME variable, put everything in this script into a function by wrapping it with `setup_cluster() {}`, and then put

```
setup_cluster "${LT_CLUSTER_NAME}"
setup_cluster "${LT_ENVIRON_CLUSTER_NAME}"
```

There's no need to start accepting input or do validation in a test setup script.

zerobfd (changes requested):

scripts/asm-installer/tests/setup_longterm_cluster
jasonwzm: I did that just to help our testing pipeline run a little faster and also it is easier to know from the cloudbuild file clusters to be set up.
But I am also okay with what you are suggesting here.

zerobfd (approved):
Thanks! I appreciate it.

Also, whatever text editor you're using doesn't put a newline at the end of the file sometimes. Some systems don't handle this properly since it's not in the POSIX spec. Can you add one to setup_longterm_cluster before merging?

PR #441, opened 2021-02-10
Allow revisioned istiod to be exposed for discovery

jasonwzm:
Re-using the existing file so that multiproject multicluster setup can still use the same file.

Manually asking @nmittler for review. Nathan, can you also take a look at the change?

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -120,6 +120,24 @@ openAPI:
           name: anthos.servicemesh.rev
           value: ASM_REV
           isSet: true
+    io.k8s.cli.setters.anthos.servicemesh.istiodHost:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.istiodHost
+          value: istiod.istio-system.svc.cluster.local
+          isSet: true
+    io.k8s.cli.setters.anthos.servicemesh.istiod-vs-name:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.istiod-vs-name
+          value: istiod-vs
+          isSet: true
+    io.k8s.cli.setters.anthos.servicemesh.istiod-dr-name:
+      x-k8s-cli:
+        setter:
+          name: anthos.servicemesh.istiod-dr-name
+          value: istiod-dr
+          isSet: true
     io.k8s.cli.setters.anthos.servicemesh.canonicalServiceHub:
       x-k8s-cli:
         setter:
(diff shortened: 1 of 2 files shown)


zerobfd (commented):
When you tested this what did the kpt command look like? I thought if it were done like this you'd need to specify the entire name, instead of e.g. how we do the Service name where all you set via the cli is the revision tag and the value gets generated. See gke-hub-metadata for an example.

I'm not a `kpt` expert so this could be right right way.

jasonwzm:
@zerobfd 
Here I am using entire value replacement. `kpt cfg set anthos.servicemesh.istiodHost istiod-asm-182-2.istio-system.svc.cluster.local` The reason is I still want multiproject setup to use the raw yaml, therefore I cannot use substitution. With substitution, that value on the yaml would look like `istiod-REV.istio-system.svc.cluster.local`.

@elfinhe Any thoughts on this? since you worked with kpt before.

elfinhe:
> @zerobfd
> Here I am using entire value replacement. `kpt cfg set anthos.servicemesh.istiodHost istiod-asm-182-2.istio-system.svc.cluster.local` I still want multiproject setup to use the raw yaml, therefore I cannot use substitution. With substitution, that value on the yaml would look like `istiod-REV.istio-system.svc.cluster.local`.
> 
> @elfinhe Any thoughts on this? since you worked with kpt before.

what are the other alternative values of anthos.servicemesh.istiodHost? And what use cases are this istiod discovery used for, for connecting sidecars to istiod?

jasonwzm:
> what are the other alternative values of anthos.servicemesh.istiodHost, need to support MCP? 

Yes and any remote or in-cluster istiod address.

> And what use cases are this istiod discovery used for, for connecting sidecars to istiod?

Yes and also service proxy agent and sds agent connecting to istiod.

elfinhe:
> > what are the other alternative values of anthos.servicemesh.istiodHost, need to support MCP?
> 
> Yes and any remote or in-cluster istiod address.
> 
> > And what use cases are this istiod discovery used for, for connecting sidecars to istiod?
> 
> Yes and also service proxy agent and sds agent connecting to istiod.

LGTM. Thanks Jason.

I remembered that the istiod VS can only work for connections via the gateways, not other sidecars in the cluster (connecting to istiod during sidecar bootstrapping before envoy route rules take effect). I guess that is the case for vm. So looks good.

jasonwzm:
@elfinhe Yes and you can see the gateway on the VS.

Can you also approve the PR? Adding you as a reviewer.

elfinhe (approved):

nmittler (approved):

PR #443, opened 2021-02-10
Enable new auth for managed control plane gateway

ruigulala:
Use gcr.io/wlhe-cr/asm repo to install MCP gateway.
This will let us test the new image.


zerobfd (approved):
Let's try and remember to not leave these in before this gets merged into staging

costinm (approved):
Will it work ? The hub is not public. I think we still need to copy the 1.9-alpha to wlhe or some other public gcr repo.

ruigulala:
> Will it work ? The hub is not public. I think we still need to copy the 1.9-alpha to wlhe or some other public gcr repo.

Done. PTAL.

PR #445, opened 2021-02-10
Enable new auth for managed control plane gateway

ruigulala:
Use gcr.io/wlhe-cr/asm repo to install MCP gateway.
This will let us test the new image.


zerobfd:
Merged manually.

PR #449, opened 2021-02-11
Add BSD csplit compatibility

zerobfd:
*Some* platforms use BSD instead of GNU, so we check for empty files manually instead of handling it with csplit.


elfinhe (approved):
Cool

PR #452, opened 2021-02-12
Check existence of membership CR

jasonwzm:


asm/vm/asm_vm
```diff
@@ -644,6 +644,11 @@ is_membership_crd_installed() {
   if ! kubectl api-resources --api-group=hub.gke.io | grep -q memberships; then
     false
```
zerobfd, line 645: return under this false

scripts/asm-installer/install_asm
```diff
@@ -1895,6 +1895,11 @@ is_membership_crd_installed() {
   if ! kubectl api-resources --api-group=hub.gke.io | grep -q memberships; then
     false
```
zerobfd, line 1896: return under this false

zerobfd (approved):

PR #453, opened 2021-02-12
Remove symlink, move flaky test to separate trigger

zerobfd:
This will unblock development for known flaky tests that aren't for core functionality.


jasonwzm (approved):

jasonwzm:
It is probably better to also do setup_long_term_cluster for flaky test. Otherwise, they may be always flaky because resources are not cleaned up.

ruigulala (approved):

elfinhe (approved):

PR #457, opened 2021-02-12
enable cross cluster workload entry reading by default

stevenctl:


jasonwzm (changes requested):
Also add to hub-meshca file. https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/asm/istio/options/hub-meshca.yaml

jasonwzm (approved):
Thanks!

stevenctl:
Is there something similar to `/retest` for this repo? I don't see anything in the logs for the one failure, and one says it needs a collaborator to trigger.

PR #462, opened 2021-02-15
bump asm 1.8 version

richardwxn:


zerobfd:
There were a lot of commits that needed to be rebased onto staging to fix various test issues. I rolled the version bump into a staging merge with more cherry-picks in #465 so if that sounds OK we can close this PR.

PR #464, opened 2021-02-16
Update service proxy image

jasonwzm:


zerobfd:
Can you template the default string somehow and then rely on the environment variable for specific versions that differ from the normal pattern?

jasonwzm:
> Can you template the default string somehow and then rely on the environment variable for specific versions that differ from the normal pattern?

Not yet. This is a staging version of the image. We can do that when official release is out.

zerobfd:
What are we using this for that we don't have control over the environment variable to set the image? I don't understand why this requires a code change.

PR #467, opened 2021-02-16
[VM Script] Add default istiod service name resolution

jasonwzm:
/cc @stevenctl


zerobfd (approved):
lgtm if VM people say OK

stevenctl (approved):

PR #472, opened 2021-02-17
dont set default revision in meta

stevenctl:
It may be better for this logic to exist in the agent, but I'm testing against this workaround. 

similar logic: 

https://github.com/istio/istio/blob/386a2dbd7f71c16c0f47820d4f4cc3322ef87e2c/manifests/charts/istio-control/istio-discovery/templates/configmap.yaml#L60

https://github.com/istio/istio/blob/90e498421d5ccdb1c4631c63d8e30dc5d913308f/istioctl/cmd/workload.go#L511-L519

### DIFF
--- a/asm/vm/asm_vm
+++ b/asm/vm/asm_vm
@@ -1001,6 +1001,12 @@ EOF
 # Generate GCE service proxy definition for the instance template
 gen_gce_service_proxy() {
   local PROXY
+
+  local ASM_META_REVISION="${WORKLOAD_ASM_REVISION}"
+  if [ "${ASM_META_REVISION}" == "default" ]; then
+    ASM_META_REVISION=""
+  fi
+
   PROXY=$(jq -n --arg ingress_ip "${INGRESS_IP}" \
     --arg project_id "${PROJECT_ID}" \
     --arg workload_name "${WORKLOAD_NAME}" \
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/install_asm
```diff
@@ -2068,7 +2068,7 @@ configure_package() {
     kpt cfg set asm anthos.servicemesh.external_ca.ca_name "${CA_NAME}"
   fi
 
-  if [[ "${USE_VM}" -eq 1 ]]; then
+  if [[ "${USE_VM}" -eq 1 ]] && [[ _CI_NO_REVISION -eq 0 ]]; then
```
jasonwzm, line 2071: should quote variable "${_CI_NO_REVISION}"

asm/vm/asm_vm
```diff
@@ -1001,6 +1001,12 @@ EOF
 # Generate GCE service proxy definition for the instance template
 gen_gce_service_proxy() {
   local PROXY
+
+  local ASM_META_REVISION="${WORKLOAD_ASM_REVISION}"
+  if [ "${ASM_META_REVISION}" == "default" ]; then
```
jasonwzm, line 1006: What if the user has a revision called default? Maybe we should have a _USE_CI toggle for that.

jasonwzm (changes requested):

asm/vm/asm_vm
stevenctl: If their revision is called `default` there are bigger issues... and they probably want this to get hit?

jasonwzm (approved):

jasonwzm:
@stevectl can you rebase? it seems the flaky tests are still in master suite.

PR #478, opened 2021-02-17
handle default revision when fetching config map

stevenctl:
same as https://github.com/istio/istio/blob/90e498421d5ccdb1c4631c63d8e30dc5d913308f/istioctl/cmd/workload.go#L391

although the template doesn't have `if rev == default`, it assumes the revision isn't passed at all when it is `default`. 

Without this, an ASM installation with `_CI_NO_REVISION` will make the vm script error with `asm_vm: [ERROR]: The environ workload identity is not used in any of your ASM installation in the cluster`


jasonwzm (approved):
good catch!

PR #482, opened 2021-02-19
Don't register with -e unless using a feature that requires it

zerobfd:
Added/updated tests, also I had to refactor the `can_*` functions because the negative logic was too difficult to read.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -687,7 +687,9 @@ FLAGS:
   are incompatible with --only_validate.
 
   -e|--enable_all                     Allow the script to perform all of the
-                                      individual enable actions below.
+                                      individual enable actions below. (Environ
+                                      registration won't happen unless necessary
+                                      for a selected option.)
      --enable_cluster_roles           Allow the script to attempt to set
                                       the necessary cluster roles.
      --enable_cluster_labels          Allow the script to attempt to set
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -488,41 +488,74 @@ can_modify_at_all() {
 }
 
 can_modify_cluster_roles() {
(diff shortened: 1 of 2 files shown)


elfinhe (approved):

scripts/asm-installer/install_asm
```diff
@@ -488,41 +488,74 @@ can_modify_at_all() {
...
+    false
+  fi
 }
 
 can_register_cluster() {
-  if [[ "${ENABLE_ALL}" -eq 0 && "${ENABLE_REGISTRATION}" -eq 0 ]] \
-    || ! can_modify_at_all; then false; fi
+  if ! can_modify_at_all; then false; return; fi
+
+  if [[ "${ENABLE_ALL}" -eq 1 && ("${USE_VM}" -eq 1 || "${USE_HUB_WIP}" -eq 1) ]] \
```
jasonwzm, line 543: USE_HUB_WIP should be enough here. 

Im okay with this approach. Like previous release changes, we should also change the description of --enable-all to reflect this.

jasonwzm (changes requested):

scripts/asm-installer/install_asm
zerobfd: USE_VM is there because HUB_WIP isn't the only environ dependency.

jasonwzm: We fail early if VM is not specified with hub-meshca. https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm#L960

zerobfd: I'm aware. Even if VMs didn't need HUB_WIP, it would still need environ registration. Removing it serves no purpose, and leaving it in makes that dependency more explicit.

jasonwzm: Good point. LGTM.

jasonwzm (approved):

PR #487, opened 2021-02-19
Print flashy message when using tmp directory

zerobfd:
A lot of users miss this in the outro, so this makes the up-front notification more obvious.


elfinhe (approved):
LGTM

PR #491, opened 2021-02-23
Update istiod routing config

jasonwzm:
Update based on https://github.com/istio/istio/blob/master/samples/multicluster/expose-istiod.yaml to use TLS routing for correct revision selection.

/cc @nmittler

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -124,6 +124,12 @@ openAPI:
       x-k8s-cli:
         setter:
           name: anthos.servicemesh.istiodHost
+          value: istiod.istio-system.svc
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
lgtm

elfinhe (approved):

JimmyCYJ:
Can we run a multi-cluster test to verify that this change does not break anything?

jasonwzm:
> Can we run a multi-cluster test to verify that this change does not break anything?

None of the ASM multicluster instructions actually uses expost-istiod file currently.

PR #498, opened 2021-02-24
Don't install the gateway on ASM installations with MCP

zerobfd:


scripts/asm-installer/install_asm
```diff
@@ -2352,12 +2352,7 @@ install_managed_components() {
   kubectl apply -f "${BASE_REL_PATH}"
   kubectl replace -f "${MANAGED_WEBHOOKS}"
 
-  PARAMS="-f ${MANAGED_MANIFEST}"
-  PARAMS="${PARAMS} -c ${KUBECONFIG}"
-  PARAMS="${PARAMS} --skip-confirmation"
-  PARAMS="${PARAMS} --set revision=${REVISION_LABEL}"
-
-  info "Installing ASM managed control plane components..."
+  info "Configuring ASM managed control plane components..."
```
costinm, line 2355: That doesn't fix the problem - istioctl install is still there.

Also: it would be great to print "To install gateway, run .... "

scripts/asm-installer/install_asm
```diff
@@ -2352,12 +2352,7 @@ install_managed_components() {
...
   # shellcheck disable=SC2086
   retry 5 istioctl install $PARAMS
```
costinm, line 2357: That's the line to delete.

zerobfd: Thanks, trying to do too many things at once

scripts/asm-installer/install_asm
zerobfd: Removed the install. If we're going to print anything I'd rather print a link to the docs like the other examples in the outro.

costinm: That's fine too.

costinm (approved):

ruigulala (approved):

PR #501, opened 2021-02-25
Don't patch validating webhook when provisioning managed control plane

zerobfd:


costinm (commented):
This is not sufficient unfortunately. We need a 'base' that doesn't have the validatingwebhookconfig ( i.e. global.configValidation=false ).

I'm starting a PR to not include the validation webhook into the generated file - but now I'm getting more paranoid. Can you use the same code that istio-gcp.sh is using ( or similar ):
```
kubectl get ns istio-system
# shellcheck disable=SC2181
if [[ "$?" != "0" ]]; then
  echo "Initializing istio-system and CRDs, fresh cluster"
  kubectl create ns istio-system
  kubectl apply -f /var/lib/istio/config/gen-istio-cluster.yaml 
fi
```

The 'base' has a bunch of other customizable params, and if a previous install exists we don't want to override it. We will need to document how to upgrade the base - for public preview we officially only support upgrades from ASM-1.9, but even ASM-1.9 install is allowed to use overlay, so there is no guarantee that it was installed in the same config. 

For example user may enable remote istiod or other options, and we would wipe them when provisioning managed.

costinm:
https://gke-internal-review.googlesource.com/c/istio/istio/+/232004

scripts/asm-installer/install_asm
```diff
@@ -2394,6 +2393,10 @@ outro() {
     info "${OUTPUT_DIR}/${RAW_YAML}"
     info "The full, expanded set of kubernetes resources can be found at:"
     info "${OUTPUT_DIR}/${EXPANDED_YAML}"
+  else
+    info "You can find the validatingwebhook and gateway configuration files at:"
+    info "${OUTPUT_DIR}/${MANAGED_WEBHOOKS}"
+    info "${OUTPUT_DIR}/managed_control_plane_gateway.yaml"
```
costinm, line 2399: We still need to modify the name of the webhook in this file - otherwise when user applies it will still be patched by istiod and stop working.

I guess if we do modify the name - we can as well apply it...

costinm (commented):
To clarify: we must also edit asm/istio/options/managed-control-plane-webhooks.yaml  and change the name of the resource. 

I am ok with not applying it. 
I am ok with applying it as well - and potentially having 2 validators.

But the name must be changed, otherwise it won't work.

costinm (approved):

PR #502, opened 2021-02-25
Save generated kubeconfig, create OUTPUT_DIR when safe to do so

zerobfd:


costinm (approved):
Seems reasonable - if user specifies an out dir, we should not use tmp dir.

PR #506, opened 2021-02-26
Add a version check of scripts

jasonwzm:


scripts/asm-installer/tests/script_version_check.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -eu
+
+while read -r KEYWORD; do
+  INSTALL_ASM_LINE="$(grep "^${KEYWORD}=" "${INSTALL_ASM_SCRIPT}")"
```
zerobfd, line 5: I'd set `INSTALL_ASM_SCRIPT=install_asm; readonly INSTALL_ASM_SCRIPT;` in this file instead of passing them in by env. They're extremely unlikely to change, and if they do then the tests are going to fail anyways, and it would be easier to find the cause of the error.

scripts/asm-installer/cloudbuild-master.yaml
```diff
@@ -22,6 +22,18 @@ steps:
...
+    id: 'script-version-check'
+    entrypoint: '/bin/bash'
+    args:
+      - '-c'
+      - >
+        ./tests/script_version_check.sh
+    env:
+      - 'INSTALL_ASM_SCRIPT=${_SCRIPT_NAME}'
+      - 'ASM_VM_SCRIPT=${_ASM_VM_SCRIPT_NAME}'
+
```
zerobfd, line 36: Don't parameterize these, just hardcode them in the script.

scripts/asm-installer/tests/script_version_check.sh
```diff
@@ -0,0 +1,24 @@
...
+  if [[ "${INSTALL_ASM_LINE}" != "${ASM_VM_LINE}" ]]; then
+    echo "${KEYWORD} version line does not match in install_asm and asm_vm. Please make the change."
+    exit 1
+  fi
+done <<EOF
+MAJOR
+MINOR
+POINT
+REV
+EOF
```
zerobfd, line 24: echo "Success" or something at the end so that something shows up in the logs

zerobfd (approved):
lgtm after a couple of changes, thanks!

scripts/asm-installer/tests/script_version_check.sh
jasonwzm: Done.

scripts/asm-installer/cloudbuild-master.yaml
jasonwzm: Done.

scripts/asm-installer/tests/script_version_check.sh
jasonwzm: Done.

PR #508, opened 2021-02-26
Add GKE WP enablement condition for unmanaged ASM

jasonwzm:
Avoid GKE WP validation error when Environ WP is being used.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1998,7 +1998,7 @@ should_enable_gke_workload_identity() {
 
 enable_gke_workload_identity(){
   if is_gke_workload_identity_enabled; then return; fi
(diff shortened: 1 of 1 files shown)


zerobfd (changes requested):
"GKE workload identity" isn't a product, and I'm pretty sure the plan is to move everything under the same WI anyways. The logic change is fine, but don't re-name anything.

jasonwzm:
Good point. I removed the rename in the messages. But I'd still like to change the method names to make it clear for internal dev.

zerobfd:
In that case use "svc" to match the workload pool name.

jasonwzm:
Now that I think about it more, we may not need to differentiate WI then. Will keep it as it was.

zerobfd (approved):

PR #513, opened 2021-03-01
Symlink istioctl binary, clarify MCP outro

zerobfd:


jasonwzm:
What is the intent behind using symlink for istioctl?

jasonwzm (approved):

PR #515, opened 2021-03-02
Include directions for upgrating validation fix on upgrade/migration

zerobfd:
Right now we don't give any guidance for upgrades/migrations if the existing installation uses a custom istiod service as a fix for the OSS validatingwebhook bug. This commit copies the config to a known location and adds steps to the outro.

### DIFF
--- a/scripts/asm-installer/tests/run_migration_suite_meshca
+++ b/scripts/asm-installer/tests/run_migration_suite_meshca
@@ -67,6 +67,7 @@ main() {
       -p ${PROJECT_ID} \
       -m migrate \
       -c mesh_ca \
+      -D mesh_migration \
       -s ${SERVICE_ACCOUNT} \
       -k ${KEY_FILE} -v -e"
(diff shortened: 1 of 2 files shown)


elfinhe (approved):

scripts/asm-installer/tests/run_migration_suite_meshca
```diff
@@ -123,6 +127,13 @@ main() {
     set -e
   fi
 
+  echo "Checking for validation service yaml..."
+
+  if [[ ! -f mesh_migration/istiod-service.yaml ]]; then
```
elfinhe, line 132: np: use variable to represent the filename? Not a must, but depending on how to manage all filenames

zerobfd: I always waffle for tests. One one hand changing the filename will require a change, on the other hand if the file accidentally get changed it would be nice for the test to catch it.

PR #516, opened 2021-03-03
Modifying asm installer script to add support for custom revision tag. 

shankgan:
Also adding asm option file for meshCA trustAnchor


scripts/asm-installer/install_asm
```diff
@@ -682,7 +684,11 @@ OPTIONS:
                                       Name of the ca in the GCP CAS service used to
                                       sign certificates in the format
                                       'projects/project_name/locations/ \
-                                       ca_region/certificateAuthorities/ca_name'
+                                      ca_region/certificateAuthorities/ca_name'
+  --revision       <CUSTOM_REVISION>  Custom revision label. Label needs to follow DNS
+                                      label formats (re: RFC 1123). Not supported if
+                                      control plane is managed. Recommended format:
+                                      asm-<major>.<minor>.<point>-revisionTag
```
zerobfd, line 691: This is RFC 1123, compliant, but the revision also needs to be a valid Kubernetes label, which doesn't allow periods: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

scripts/asm-installer/install_asm
```diff
@@ -1079,6 +1099,20 @@ EOF
   WORKLOAD_POOL="${PROJECT_ID}.svc.id.goog"
 }
 
+validate_revision_label() {
+  # Matches DNS label formats of RFC 1123
+  local DNS1123LabelMaxLength; DNS1123LabelMaxLength=63
+  local DNS1123LabelFmt; DNS1123LabelFmt="[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?"
+  
+  if [[ ${#REVISION_LABEL} -gt $DNS1123LabelMaxLength ]]; then
```
zerobfd, line 1107: Please wrap ${} in double quotes.

scripts/asm-installer/install_asm
```diff
@@ -1079,6 +1099,20 @@ EOF
   WORKLOAD_POOL="${PROJECT_ID}.svc.id.goog"
 }
 
+validate_revision_label() {
+  # Matches DNS label formats of RFC 1123
+  local DNS1123LabelMaxLength; DNS1123LabelMaxLength=63
+  local DNS1123LabelFmt; DNS1123LabelFmt="[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?"
```
zerobfd, line 1105: Let's use UPPER_SNAKE_CASE to follow convention, and add `readonly VAR_NAME` for these after setting the value.

asm/istio/options/trustanchor-meshca.yaml
```diff
@@ -0,0 +1,59 @@
+# Copyright 2020 Google LLC
```
zerobfd, line 1: I don't think this file is supposed to be in the commit, is it?

scripts/asm-installer/install_asm
```diff
@@ -602,6 +603,7 @@ OPTIONS:
   --root_cert            <FILE PATH>
   --cert_chain           <FILE PATH>
   --ca_name              <CA NAME>
+  --revision             <CUSTOM_REVISION>
```
zerobfd, line 606: Let's use `--revision_name`, change `CUSTOM_REVISION` to `REVISION_NAME`, and add a short flag `-r`. The first two are to align with later usage in other places in ASM, the last one because I imagine this is a flag that will get specified a lot.

scripts/asm-installer/install_asm
```diff
@@ -904,6 +910,12 @@ parse_args() {
         CUSTOM_CA=1
         shift 2
         ;;
+      --revision)
```
zerobfd, line 913: When this is changed, accept `revision_name` and `revision-name`.

scripts/asm-installer/install_asm
```diff
@@ -1079,6 +1099,20 @@ EOF
...
+  # Matches DNS label formats of RFC 1123
+  local DNS1123LabelMaxLength; DNS1123LabelMaxLength=63
+  local DNS1123LabelFmt; DNS1123LabelFmt="[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?"
+  
+  if [[ ${#REVISION_LABEL} -gt $DNS1123LabelMaxLength ]]; then
+    fatal "Custom revision label cannot be longer than $DNS1123LabelMaxLength."
+  fi
+
+  if ! [[ "$REVISION_LABEL" =~ $DNS1123LabelFmt ]]; then
+    fatal "Custom revision label does not follow RFC 1123 formatting convention"
```
zerobfd, line 1112: Remove "Custom", capitalize "revision", add period to the end.

scripts/asm-installer/install_asm
```diff
@@ -953,6 +969,10 @@ validate_args() {
     if [[ "${USE_VM}" -eq 1 ]]; then
       fatal "Adding VM workloads with managed control plane is not supported."
     fi
+
+    if [[ "${CUSTOM_REVISION}" -eq 1 ]]; then
+      fatal "Adding custom revision label with managed control plane is not supported"
```
zerobfd, line 974: I think "Specifying a revision label" is a bit closer to what's happening, instead of "Adding custom revision label".

zerobfd (changes requested):
Logic looks perfect, some wording/style issues.

scripts/asm-installer/install_asm
shankgan: corrected regex to not allow this.

asm/istio/options/trustanchor-meshca.yaml
shankgan: I wanted to add the option file for Mesh CA trustanchor in the same commit. Is that okay?

zerobfd: Can you explain a bit how it's related to revision labels? Also do you want it in the same PR, or the same commit? I'd rather the commits be separate even if it's the same PR just to make commit management easier later.

scripts/asm-installer/install_asm
```diff
@@ -682,7 +684,11 @@ OPTIONS:
                                       Name of the ca in the GCP CAS service used to
                                       sign certificates in the format
                                       'projects/project_name/locations/ \
-                                       ca_region/certificateAuthorities/ca_name'
+                                      ca_region/certificateAuthorities/ca_name'
+  -r|--revision_name <REVISION NAME>  Custom revision label. Label needs to follow DNS
+                                      label formats (re: RFC 1123). Not supported if
+                                      control plane is managed. Recommended format:
+                                      asm-revisionTag
```
zerobfd, line 691: I think maybe "Prefixing the revision name with 'asm' is recommended" is clearer than "Recommended format: asm-revisionTag". (plus tags aren't a thing yet)

zerobfd (approved):
1 wording update and I want to finish the thread about the trust anchor but code looks good, I see no reason to not approve.

asm/istio/options/trustanchor-meshca.yaml
shankgan: I wanted to add it as part of the same PR, not the same commit. This installation option installs the meshCA trust Anchor in the ASM mesh so that  workloads signed by Istiod can communicate with ASM mesh CA workloads.

shankgan: Pushed into two separate commits

shankgan: I can move this out into a separate PR if you think we need to discuss this in more detail

zerobfd: If they're in separate commits that's enough for me :)

PR #519, opened 2021-03-04
Add release-1.9-asm into release script.

elfinhe:


scripts/asm-installer/release_asm_installer
```diff
@@ -19,6 +19,7 @@ BUCKET_PATH="csm-artifacts/asm"; readonly BUCKET_PATH
 CURRENT_RELEASE="release-1.8-asm"; readonly CURRENT_RELEASE
```
jasonwzm, line 19: release-1.9-asm

jasonwzm (changes requested):

jasonwzm (approved):

scripts/asm-installer/release_asm_installer
elfinhe: nice catch, thanks, updated.

PR #520, opened 2021-03-04
set ISTIO_META_CLUSTER_ID in metadata

stevenctl:
istiod uses this in many areas; one thing in particular this breaks (using the latest changes to master post 1.9) is that we won't send DNS for headless services unless this is set. 

I've manually tested this by editing instance metadata/resetting. It fixes DNS resolution of headless services. This change should probably be backported to 1.9 as well.

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -1246,6 +1246,7 @@ gen_gce_service_proxy() {
     --arg asm_revision "${ASM_META_REVISION}"\
     --arg credential_identity_provider "${CREDENTIAL_IDENTITY_PROVIDER}" \
     --arg network "${NETWORK}" \
+    --arg clusterID "${CLUSTER_NAME}" \
     --arg asm_meta_version "${ASM_META_VERSION}" '{
   "mode": "ON",
   "proxy-spec": {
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

scripts/asm-installer/asm_vm
```diff
@@ -1246,6 +1246,7 @@ gen_gce_service_proxy() {
     --arg asm_revision "${ASM_META_REVISION}"\
     --arg credential_identity_provider "${CREDENTIAL_IDENTITY_PROVIDER}" \
     --arg network "${NETWORK}" \
+    --arg clusterID "${CLUSTER_NAME}" \
```
jasonwzm, line 1249: Does this need to be the same as cluster name in multicluster? https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/956568b25dd2318fec585cfe17539671fa90841b/asm/istio/options/multicluster.yaml#L23

stevenctl: Oh, right those are different. Yes it does.

jasonwzm: So should it be cn-${PROJECT_ID}-${CLUSTER_LOCATION}-${CLUSTER_NAME} here?

stevenctl: Fixed.. I don't want to hardcode that templating here since it has variations in other environments (such as onprem). I used something a bit more future proof: grabbing it from the injector values.

stevenctl: Just manually tested with these changes

nmittler (approved):

jasonwzm (approved):

jasonwzm:
@stevenctl It seems you changed the permission of the file to 644. Can you change it back to 755?

jasonwzm:
@stevenctl You probably need to do a `chmod +x asm_vm` on your local file and then push.

PR #526, opened 2021-03-05
vm eastwest iop should not have name

stevenctl:
Without this, the IstioOperator will overwrite other parts of the installation. 

the istiod deployment/service, mutating webhook configmap/config, validating webhook config and probably a ton of other things end up with `install.operator.istio.io/owning-resource: eastwest` which is incorrect.

### DIFF
--- a/asm/istio/options/vm.yaml
+++ b/asm/istio/options/vm.yaml
@@ -15,8 +15,6 @@
 ---
 apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
-metadata:
-  name: eastwest
 spec:
   components:
     ingressGateways:


stevenctl:
We had this here, not sure why it wasn't copied over:

https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/b7cff407f77d74c4b7fb750b4cd853059cac80eb/asm/istio/expansion/vm-eastwest-gateway.yaml#L21

stevenctl:
AFAIK ASM doesn't use the istio operator controller at all, but if it did, this would probably cause issues.

jasonwzm:
These are two different use cases. The vm.yaml file is for install_asm when we add it as an overlay.

stevenctl:
If that's the case, we probably don't want to set `name` in `vm.yaml`?

jasonwzm:
Oh I did not notice that. Yes, that makes sense. We can remove metadata field from this yaml.

jasonwzm:
It does not seem your new change is pushed.

asm/istio/options/vm.yaml
```diff
@@ -34,7 +32,7 @@ spec:
             ports:
               - name: status-port
                 port: 15021
-                targetPort: 15021
+                targetPort: 15021+
```
jasonwzm, line 35: Is this a typo?

jasonwzm (changes requested):

asm/istio/options/vm.yaml
stevenctl: yep...

jasonwzm (approved):

PR #527, opened 2021-03-07
add image_family parameter

williamaronli:
add image_family parameter when asm_vm installation

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -97,6 +97,7 @@ WORKLOAD_NAMESPACE="${WORKLOAD_NAMESPACE:=}"
 
 SERVICE_ACCOUNT="${SERVICE_ACCOUNT:=}"
 KEY_FILE="${KEY_FILE:=}"
+IMAGE_FAMILY="${IMAGE_FAMILY:=}"
 
 DRY_RUN="${DRY_RUN:=0}"
 ONLY_VALIDATE="${ONLY_VALIDATE:=0}"
@@ -341,6 +342,15 @@ OPTIONS:
                                                       Required if
                                                       --service_account is
                                                       passed.
+  -m|--image_family           <IMAGE_FAMILY>          (optional) The image family
+                                                      for the operating system
+                                                      Only valid when the
+                                                      --source_instance_template
+                                                      is not specified. Default
+                                                      to the "debian-10" if not
+                                                      specified. Available os
+                                                      image-families are:
+                                                      [debian-10|debian-9|centos-7|centos-8]
 
 FLAGS:
 
@@ -501,6 +511,11 @@ parse_args_create_gce_instance_template() {
         KEY_FILE="${2}"
         shift 2
         ;;
+      -m | --image_family | --image-family)
+        arg_required "${@}"
+        IMAGE_FAMILY="${2}"
+        shift 2
+        ;;
       --dry_run | --dry-run)
         DRY_RUN=1
         shift 1
@@ -630,6 +645,10 @@ EOF
     readonly KEY_FILE
   elif [[ -n "${KEY_FILE}" ]]; then
     fatal "Couldn't find key file ${KEY_FILE}."
+  fi
+
+  if [[ ! -z "${SOURCE_INSTANCE_TEMPLATE}" ]] && [[ ! -z "${IMAGE_FAMILY}" ]]; then
+      warn "source instance template has been specified and the image family parameter will not be used"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/asm_vm
```diff
@@ -341,6 +342,15 @@ OPTIONS:
                                                       Required if
                                                       --service_account is
                                                       passed.
+  -m|--image_family           <IMAGE_FAMILY>          (optional) The image family
+                                                      for the operating system
+                                                      Only valid when the
+                                                      --source_instance_template
+                                                      is not specified. Default
```
JimmyCYJ, line 349: Why is this only valid when --source_instance_template is not specified? 

For example, when creating a source instance template for bookinfo, that template only specifies the logic for bootstrapping the mysql app. Maybe a better way is to allow source_instance_template and overwrite the image type in the source instance template? @jasonwzm WDYT?

JimmyCYJ: IIUC, this command should work. Otherwise it breaks CUJ
```
./asm_vm create_gce_instance_template \
asm-mysql-instance-template \
--project_id PROJECT_ID \
--cluster_location CLUSTER_LOCATION \
--cluster_name CLUSTER_NAME \
--workload_name mysql \
--image_family debian-10 \
--source_instance_template mysql-instance-template
```

williamaronli: source template already include a linux distros https://cloud.google.com/sdk/gcloud/reference/compute/instance-templates/create

jasonwzm: I think we need to be clear on what this flag really entails. Does it mean the OS the service proxy agent supports or the OS that will be created/overridden in the source instance template? What if the user uses a custom image which is built on top of centos and they would like a service proxy agent for centos?

scripts/asm-installer/asm_vm
```diff
@@ -631,6 +646,10 @@ EOF
   elif [[ -n "${KEY_FILE}" ]]; then
     fatal "Couldn't find key file ${KEY_FILE}."
   fi
+
+  if [[ ! -z "${SOURCE_INSTANCE_TEMPLATE}" ]] && [[ ! -z "${IMAGE_FAMILY}" ]]; then
```
jasonwzm, line 650: replace with `-n`

scripts/asm-installer/asm_vm
```diff
@@ -631,6 +646,10 @@ EOF
...
+      warn "source instance template has been specified and the image family parameter will not be used"
```
jasonwzm, line 651: The description says having two flags specified together is not valid. If it is a validation error, we should error out.

scripts/asm-installer/asm_vm
```diff
@@ -1384,11 +1403,22 @@ gen_gce_instance_labels() {
...
 # Generate a default instance template.
 gen_default_instance_template() {
+  local source_image="projects/debian-cloud/global/images/family/debian-10"
+  if [[ ! -z "${IMAGE_FAMILY}" ]]; then
+    if [[ "${IMAGE_FAMILY}" == "debian-10" ]] || [[ "${IMAGE_FAMILY}" == "debian-9" ]]; then
+      source_image="projects/debian-cloud/global/images/family/${IMAGE_FAMILY}"
+    elif [[ "${IMAGE_FAMILY}" == "centos-7" ]] || [[ "${IMAGE_FAMILY}" == "centos-8" ]]; then
+      source_image="projects/centos-cloud/global/images/family/${IMAGE_FAMILY}"
+    else
+      fatal "${IMAGE_FAMILY} is not supported."
```
jasonwzm, line 1413: This validation should happening at the beginning of the script execution.

jasonwzm (changes requested):
What is the scope of this PR?

It seems it is only supporting changing the default template image family.

scripts/asm-installer/asm_vm
JimmyCYJ: What will happen if customer run command like this? 
```
./asm_vm create_gce_instance_template \ asm-mysql-instance-template \ --project_id PROJECT_ID \ --cluster_location CLUSTER_LOCATION \ --cluster_name CLUSTER_NAME \ --workload_name mysql \ --image_family debian-10 \ --source_instance_template mysql-instance-template
```

By looking at the script, it generates a message at `warn` level, what else does the script handle this situation?

scripts/asm-installer/asm_vm
williamaronli: As there will only be one mesh agent for both debian and centos which can automatically detect the os system and use correct envoy path

jasonwzm: Then why do we need --image-family flag at all? User can simply use source instance template to choose their intended OS.

scripts/asm-installer/asm_vm
williamaronli: yes it will warn the user and the final instance-template linux distro will use the distro from mysql-instance-template

scripts/asm-installer/asm_vm
williamaronli: but there is one cases that the user doesn't have any source instance template. They want directly create a new template in one step and I think that is the user's requirement.

scripts/asm-installer/asm_vm
williamaronli: ok I will change it to a fatal

scripts/asm-installer/asm_vm
williamaronli: ok will add that

scripts/asm-installer/asm_vm
williamaronli: I just add a limitation: At most one of these parameters may be specified

williamaronli:
close this one as our latest design says that the image-family parameter is never required

PR #529, opened 2021-03-08
change size of init vm for centos to 20gb

williamaronli:
centos distros disk size needs to be at least 20 gb to work.

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -1409,7 +1409,7 @@ gen_default_instance_template() {
         "initializeParams": {
           "sourceImage": "projects/debian-cloud/global/images/family/debian-10",
           "diskType": "pd-standard",
-          "diskSizeGb": "10"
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/asm_vm
```diff
@@ -1409,7 +1409,7 @@ gen_default_instance_template() {
         "initializeParams": {
           "sourceImage": "projects/debian-cloud/global/images/family/debian-10",
           "diskType": "pd-standard",
-          "diskSizeGb": "10"
+          "diskSizeGb": "20"
```
jasonwzm, line 1412: I think we should try to follow the GCE defaults for different image's disk size.

jasonwzm (changes requested):

williamaronli:
close the PR and add this change to this PR: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/527

PR #533, opened 2021-03-10
allow no revision label in CI

stevenctl:
This makes testing easier. Currently we have to set both `istio-injection=enabled` and `istio.io/rev=default` on the namespace.

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -19,6 +19,7 @@ fi
 _CI_ASM_IMAGE_TAG="${_CI_ASM_IMAGE_TAG:=}"
 _CI_ASM_PKG_LOCATION="${_CI_ASM_PKG_LOCATION:=}"
 _CI_ISTIOCTL_REL_PATH="${_CI_ISTIOCTL_REL_PATH:=}"
+_CI_NO_REVISION="${_CI_NO_REVISION:=0}"
 
 ### Internal variables ###
 MAJOR="${MAJOR:=1}"; readonly MAJOR;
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/asm_vm
```diff
@@ -987,13 +988,17 @@ retrieve_workload_asm_revision() {
   WORKLOAD_ASM_REVISION="$(retry 2 kubectl get namespace "${WORKLOAD_NAMESPACE}" -ojson \
     | jq -r '.metadata.labels["'${ASM_REVISION_LABEL_KEY}'"]')"
   if [[ -z "${WORKLOAD_ASM_REVISION}" ]]; then
-    { read -r -d '' MSG; fatal "${MSG}"; } <<EOF || true
+      WORKLOAD_ASM_REVISION=default
+    if [[ "${_CI_NO_REVISION}" -eq 1]]; then
```
jasonwzm, line 992: just check `"${_CI_NO_REVISION}" -eq 0`

scripts/asm-installer/asm_vm
```diff
@@ -987,13 +988,17 @@ retrieve_workload_asm_revision() {
   WORKLOAD_ASM_REVISION="$(retry 2 kubectl get namespace "${WORKLOAD_NAMESPACE}" -ojson \
     | jq -r '.metadata.labels["'${ASM_REVISION_LABEL_KEY}'"]')"
   if [[ -z "${WORKLOAD_ASM_REVISION}" ]]; then
-    { read -r -d '' MSG; fatal "${MSG}"; } <<EOF || true
+      WORKLOAD_ASM_REVISION=default
```
jasonwzm, line 991: Do we have `istio-default` svc in the CI cluster? If not, we would either need to change the script to allow the `WORKLOAD_ASM_REVISION` be empty string or change service proxy agent code to have a different logic for `default` revision.

jasonwzm (changes requested):

PR #536, opened 2021-03-11
Add check to verify the source instance template os linux distro

williamaronli:
Add check to verify the source instance template os linux distro

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -1120,8 +1120,7 @@ EOF
 
     # check whether it is a  public gcp image
     if [[ "$source_image_path" == *projects* && "$source_image_path" == *global/images/* ]]; then
-      if [[ "$source_image_path" != *centos-7* ]] && [[ "$source_image_path" != *centos-8* ]] \
-        && [[ "$source_image_path" != *debian-9* ]] && [[ "$source_image_path" != *debian-10* ]]; then
+      if [[ "$source_image_path" != *debian-9* ]] && [[ "$source_image_path" != *debian-10* ]]; then
         fatal "found not support image, the image path is $source_image_path"
       fi
     else
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -1110,6 +1110,23 @@ this project, run:
   --project="${PROJECT_ID}"
 EOF
     fi
+
+    local INSTANCE_TEMPLATE_REQ=$(retry 2 run gcloud compute instance-templates \
+      describe "${SOURCE_INSTANCE_TEMPLATE}" \
+      --format=json \
+      --project="${PROJECT_ID}")
+    local source_image_path="$(echo "${INSTANCE_TEMPLATE_REQ}" \
+    | jq '.properties.disks[].initializeParams.sourceImage')"
+
+    # check whether it is a  public gcp image
+    if [[ "$source_image_path" == *projects* && "$source_image_path" == *global/images/* ]]; then
+      if [[ "$source_image_path" != *centos-7* ]] && [[ "$source_image_path" != *centos-8* ]] \
+        && [[ "$source_image_path" != *debian-9* ]] && [[ "$source_image_path" != *debian-10* ]]; then
+        fatal "found not support image, the image path is $source_image_path"
+      fi
+    else
+      info "custom image found, skip checking"
+    fi
   fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/asm_vm
```diff
@@ -1110,6 +1110,22 @@ this project, run:
   --project="${PROJECT_ID}"
 EOF
     fi
+
+    local INSTANCE_TEMPLATE_REQ=$(retry 2 run gcloud compute instance-templates \
+      describe "${SOURCE_INSTANCE_TEMPLATE}" \
```
JimmyCYJ, line 1115: What if user does not specify SOURCE_INSTANCE_TEMPLATE? For example, user just want to create debian VM with the default setup.

`INSTANCE_TEMPLATE_REQ` is not accurate, this is more about getting SOURCE_INSTANCE_TEMPLATE, it is not relevant to any requirement.

scripts/asm-installer/asm_vm
```diff
@@ -1110,6 +1110,22 @@ this project, run:
...
+      --format=json \
+      --project="${PROJECT_ID}")
+    local source_image_path="$(echo "${INSTANCE_TEMPLATE_REQ}" \
+    | jq '.properties.disks[].initializeParams.sourceImage')"
+
+    # check whether it is a  public gcp image
```
JimmyCYJ, line 1121: There are multiple spaces between "a  public". public is not accurate here. 
This is checking if the chosen image is supported. There are other images available for GCE VM, which are public but not supported.

scripts/asm-installer/asm_vm
```diff
@@ -1110,6 +1110,22 @@ this project, run:
...
+    if [[ "$source_image_path" == *projects* && "$source_image_path" == *global/images/* ]]; then
+      if [[ "$source_image_path" != *debian-9* ]] && [[ "$source_image_path" != *debian-10* ]]; then
+        fatal "found not support image, the image path is $source_image_path"
```
JimmyCYJ, line 1124: @jasonwzm are there any test files that should be used for adding tests to cover this change?

scripts/asm-installer/asm_vm
```diff
@@ -1110,6 +1110,22 @@ this project, run:
...
+
+    local INSTANCE_TEMPLATE_REQ=$(retry 2 run gcloud compute instance-templates \
+      describe "${SOURCE_INSTANCE_TEMPLATE}" \
+      --format=json \
+      --project="${PROJECT_ID}")
+    local source_image_path="$(echo "${INSTANCE_TEMPLATE_REQ}" \
+    | jq '.properties.disks[].initializeParams.sourceImage')"
+
+    # check whether it is a  public gcp image
+    if [[ "$source_image_path" == *projects* && "$source_image_path" == *global/images/* ]]; then
```
jasonwzm, line 1122: We need to check the full path, e.g. projects/[debian|centos]-cloud/global/images (syntax may not be correct)

scripts/asm-installer/asm_vm
jasonwzm, line 1124: Should have something like: "Image not supported. Supported images are..."

scripts/asm-installer/asm_vm
```diff
@@ -1110,6 +1110,22 @@ this project, run:
...
+      if [[ "$source_image_path" != *debian-9* ]] && [[ "$source_image_path" != *debian-10* ]]; then
+        fatal "found not support image, the image path is $source_image_path"
+      fi
+    else
+      info "custom image found, skip checking"
```
jasonwzm, line 1127: If you output a skip message here, you should also have a message before this saying "Verifying GCE image..."

scripts/asm-installer/asm_vm
jasonwzm: Yes. There are two test files in /tests folder that cover VM.

scripts/asm-installer/asm_vm
jasonwzm: Just use single local variable should be enough, as you are only verifying the image path in this function. Please capitalize the local variables as well to be consistent with other local variables.

scripts/asm-installer/asm_vm
```diff
@@ -1110,6 +1110,22 @@ this project, run:
   --project="${PROJECT_ID}"
 EOF
     fi
+
+    local INSTANCE_TEMPLATE_REQ=$(retry 2 run gcloud compute instance-templates \
+      describe "${SOURCE_INSTANCE_TEMPLATE}" \
+      --format=json \
+      --project="${PROJECT_ID}")
+    local source_image_path="$(echo "${INSTANCE_TEMPLATE_REQ}" \
+    | jq '.properties.disks[].initializeParams.sourceImage')"
```
jasonwzm, line 1119: What if there are more than one disks for this instance?

scripts/asm-installer/asm_vm
jasonwzm: Please also add test in the same PR.

jasonwzm (changes requested):

williamaronli:
Create a new one and close this one will address the comments in this PR: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/541

PR #543, opened 2021-03-12
Only override boot disk name when creating instance template

jasonwzm:
Boot disk on GCE is required and is always at index 0 of the GCE disk array.

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -1462,7 +1462,7 @@ create_gce_instance_template() {
       --format=json \
       --project="${PROJECT_ID}" \
       | jq --arg name "${NEW_INSTANCE_TEMPLATE}" \
-      '.name=$name | .properties.disks[].deviceName=$name' || true)"
+      '.name=$name | .properties.disks[0].deviceName=$name' || true)"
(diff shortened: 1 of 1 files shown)


williamaronli:
verified this with my disk

zerobfd (approved):

scripts/asm-installer/asm_vm
```diff
@@ -1462,7 +1462,7 @@ create_gce_instance_template() {
       --format=json \
       --project="${PROJECT_ID}" \
       | jq --arg name "${NEW_INSTANCE_TEMPLATE}" \
-      '.name=$name | .properties.disks[].deviceName=$name' || true)"
+      '.name=$name | .properties.disks[0].deviceName=$name' || true)"
```
stevenctl, line 1465: `'.name=$name | .properties.disks=(.properties.disks | map(if .deviceName==$oldname then .deviceName=$name else . end))'` would be a bit more robust

jasonwzm: `.deviceName==$oldname` only happens at the first disk and created instance template do not include disks with the same deviceName. I am not sure how that helps.

Issue #476, opened 2021-02-17
SemVer versioning for the install_asm script


zioproto:
The [install_asm](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm) script is released to users with URLs like:

* https://storage.googleapis.com/csm-artifacts/asm/install_asm  (latest)
* https://storage.googleapis.com/csm-artifacts/asm/install_asm_1.7 (for ASM 1.7)

The usage of those URLs is documented here: https://cloud.google.com/service-mesh/docs/scripted-install/asm-onboarding#downloading_the_script

The URL pointing to a major version, like`install_asm_1.7`, downloads the latest script version for that release.

Because the script itself does not contain a `--version` option to check the script minor version, so the user cannot know what exact version is using.

Please do versioning of this script with [SemVer](https://semver.org/) to allow effective troubleshooting of users issues.

Thanks

bharathkkb:
+1 this will be useful for https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/modules/asm

zerobfd:
Sorry for the late ack, I thought I already had but I think I got this confused with an internal bug with the same issue.

I know this is a pain point, it's just been a struggle to get time to do it properly. The scheme for versioning in SemVer is already done and approved: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/docs/versioning.md

If anyone isn't afraid of wading through 2k lines of bash I'm happy to advise anyone who wants to tackle this. Here are the concrete things that need to be done:

 * Add a CONFIG_VER variable to `install_asm` to match the config version
 * Add --version to the script that will print out "$MAJOR.$MINOR.$POINT-asm.$REV+config$CONFIG_VER"
 * Change the release script to also check git tags as well as POINT and CONFIG_VER, fail if the tag doesn't match what's in `install_asm` and echo "Don't forget to tag and push before releasing", and then also upload the file to `install_asm_"$MAJOR.$MINOR.$POINT-asm.$REV+config$CONFIG_VER"`

PR #544, opened 2021-03-13
Add proper SemVer support

zerobfd:
Related to #476. The release script part is going to require a refactor so I'm splitting it into a different PR.


elfinhe (approved):

vsubramani:
Pleasantly surprised to see this change :)

One suggestion I have is to print out the version information by default, if it is not too much of a hassle. Reason is that when users pass the output of the install_asm script when creating support tickets. Most of them may not remember to attach the output of the "--version", and might save a couple of days roundtrip before they can get actual help.

PR #546, opened 2021-03-17
Don't overwrite configuration from a different version

zerobfd:
Tested all three cases manually.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -39,6 +39,7 @@ K8S_MINOR=0
 
 ### File related constants ###
 VALIDATION_FIX_FILE_NAME="istiod-service.yaml"
+ASM_VERSION_FILE=".asm_version"
 ISTIO_FOLDER_NAME=""
 ISTIOCTL_REL_PATH=""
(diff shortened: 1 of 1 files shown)


elfinhe (approved):

scripts/asm-installer/install_asm
```diff
@@ -1397,13 +1398,33 @@ EOF
 
   info "Downloading ASM kpt package..."
   retry 3 kpt pkg get --auto-set=false "${KPT_URL}" asm
+  version_message > "${ASM_VERSION_FILE}"
 }
 
 necessary_files_exist() {
   if [[ ! -f "${OUTPUT_DIR}/${ISTIOCTL_REL_PATH}" ]]; then
     false
+    return
```
jasonwzm, line 1407: This alone will solve a bunch of issues.

jasonwzm (approved):

scripts/asm-installer/install_asm
zerobfd: That by itself won't change any of the behavior. There was nothing after the if block to execute so it would return anyways.

PR #547, opened 2021-03-17
Refactor release script

zerobfd:
In preparation for adding version check logic. Behavior unchanged, tested manually.

### DIFF
--- a/scripts/asm-installer/release_asm_installer
+++ b/scripts/asm-installer/release_asm_installer
@@ -17,6 +17,18 @@ BUCKET_URL="https://storage.googleapis.com"
 BUCKET_PATH="csm-artifacts/asm"; readonly BUCKET_PATH
 
 CURRENT_RELEASE="release-1.9-asm"; readonly CURRENT_RELEASE
+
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/release_asm_installer
```diff
@@ -61,15 +73,37 @@ $(other_releases)
...
+  if curl -O "${URI}.sha256"; then
+    if sha256sum -c --ignore-missing "${NAME}.sha256" >/dev/null 2>/dev/null; then
+      echo "No changes in ${NAME}, skipping"
+      rm "${NAME}.sha256"
+      false
+      return
+    fi
+    rm "${NAME}.sha256"
+  else
+      echo "New file {FILE_NAME}"
```
jasonwzm, line 89: nit: remove the extra tab

jasonwzm (approved):

scripts/asm-installer/release_asm_installer
zerobfd: Nice catch, thanks

PR #548, opened 2021-03-17
Require workload identity before environ registration

zerobfd:
Also a small refactor to group some internal variables together.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -97,31 +97,28 @@ SERVICE_ACCOUNT="${SERVICE_ACCOUNT:=}"
 KEY_FILE="${KEY_FILE:=}"
 OUTPUT_DIR="${OUTPUT_DIR:=}"
 
-CUSTOM_CA=0
 CA_CERT="${CA_CERT:=}"
 CA_KEY="${CA_KEY:=}"
 CA_ROOT="${CA_ROOT:=}"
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/install_asm
```diff
@@ -2028,6 +2025,12 @@ generate_membership_name() {
 register_cluster() {
   if is_cluster_registered; then return; fi
 
+  if can_modify_gcp_components; then
```
jasonwzm, line 2028: Can we instead just move register cluster after WI enablement. So that we don't have this duplicate logic here.

zerobfd: It would need to go at multiple places in main because the path splits based off of various things, I'd rather have it here just in case to prevent something like this from happening again.

jasonwzm: Just two branches right? One under `is_managed`, the other one `! is_managed`.

zerobfd: For now. I don't know about you, but I'm certainly not confident that I'm going to be able to keep track of the next time a branch changes in 2.5k lines of bash. I'd rather not worry about it for the same of a few redundant lines.

jasonwzm: fair enough...

jasonwzm (approved):

elfinhe (approved):

PR #549, opened 2021-03-18
VM script should fail when service accounts mismatch

jasonwzm:
Manually verified 3 cases.


zerobfd:
I don't see any obvious issues with the code, but the instance template test is failing. I'll review after it's fixed.

jasonwzm:
@zerobfd Test failed as expected to be what the change tries to catch. Good to review now.

zerobfd (approved):

PR #554, opened 2021-03-23
Add CA migration script.

myidpt:


zerobfd (approved):
Seems reasonable. Can you remove the .sh extension since it's an executable? Also, consider adding comments or a README with a (small) description and the dependencies required (jq, kubectl, etc.)

navinger:
If you remove the .sh extension, then the directory structure is a little weird to document: 
"scripts/ca-migration/ca-migration"

Would you be open to changing the name of the script to start with a verb? Perhaps: "migrate-ca"

For the README, in addition to a brief description of what the script is for, it would be good to state:

- The script is meant only for GKE on GCP. Non-GCP platforms aren't supported.
- You need to set the kube context to run the script.

zerobfd:
Good point on making it imperative, `migrate_ca` is better.

PR #557, opened 2021-03-25
Always pull the service proxy with matching version

jasonwzm:


zerobfd (approved):
Logic seems fine. You could probably get away with not using an array and just using one string so you rely less on bash specifics, but that's up to you whether it's worth the time.

jasonwzm:
Since we require bash4+, I am thinking it is not worth the effort. But if users complain, we can make the change.

PR #559, opened 2021-03-29
Don't enable cross-cluster workload entries by default

zerobfd:


stevenctl (approved):

stevenctl:
Can we add this back in https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/asm/istio/options/vm.yaml

zerobfd:
> Can we add this back in https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/asm/istio/options/vm.yaml

It's still in hub-meshca.yaml which is required with VMs, do you want it here for clarity or is that sufficient?

stevenctl:
That's probably fine, thanks!

PR #562, opened 2021-03-29
add centos 8 support in asm_vm

williamaronli:


jasonwzm (approved):
Based on our chat, this is manually tested. So LGTM.

PR #570, opened 2021-03-31
Installation Options to allow zero downtime migration when changing CA's from Citadel to MeshCA

shankgan:


zerobfd (approved):
Can you rebase on master and push again? Tests were broken but they should be fixed now.

myidpt (approved):

PR #580, opened 2021-04-02
add a file to display all available versions 

ZhengzheYang:
Add a pre-release script to generate all previously properly tagged releases. Also add the logic to maintain a version file that contains the mapping between the SemVer versions to the actual filename. 

Refactored the release script to share a `common.sh` file in order to reuse most of the logic.


scripts/asm-installer/release_asm_installer
```diff
@@ -18,7 +18,14 @@ BUCKET_PATH="csm-artifacts/asm"; readonly BUCKET_PATH
 
 CURRENT_RELEASE="release-1.9-asm"; readonly CURRENT_RELEASE
 
+ALWAYS_PUBLISH_VERSION="${ALWAYS_PUBLISH_VERSION:=0}"
+STABLE_VERSION_FILE="STABLE_VERSIONS"; readonly STABLE_VERSION_FILE
+VERSION_FILE_PATH="${BUCKET_PATH}/${STABLE_VERSION_FILE}"
+HOLD_TYPE="temp"
+trap 'gsutil retention "${HOLD_TYPE}" release gs://"${VERSION_FILE_PATH}"' ERR  # make sure the hold is cleared
```
ZhengzheYang, line 25: Do I need EXIT and SIGINT here? I tried but the command seems to run multiple times if I ctrl-c. Seems like the shell received multiple signal in addition to SIGINT.

scripts/asm-installer/release_asm_installer
```diff
@@ -118,6 +127,28 @@ get_stable_version() {
...
 }
 
+get_version_file_and_lock() {
+  local FILE_PATH; FILE_PATH=${BUCKET_PATH}/${STABLE_VERSION_FILE}
+
+  if [[ ! $(curl -O "${BUCKET_URL}/${FILE_PATH}") ]]; then
+    echo "Create New Version File ${STABLE_VERSION_FILE}"
+    true >| ${STABLE_VERSION_FILE}
+    gsutil cp "${STABLE_VERSION_FILE}" gs://"${FILE_PATH}"
+  fi
```
zerobfd, line 137: A few things here.

High level, I'm not sure we should do this in the script. I can't imagine why we'd ever delete this file once we have it, and if we do this then it's very possible that whoever is running the script might not even notice the log line saying that we're creating a new one. Unless we have some known situations where this might happen, I'd rather fail loudly so that it's extremely clear what's going on.

Style issues:

1. if you're checking the return value of a function you don't need [[]] `if ! curl -O ... ;then`
2. generally, if you're echoing something that you intend a human to read, default to putting it in `stderr` with >&2
3. to create a file, I generally see `touch $FILE` not `true > $FILE`.

scripts/asm-installer/release_asm_installer
```diff
@@ -118,6 +127,28 @@ get_stable_version() {
...
+
+  gsutil retention ${HOLD_TYPE} set "gs://${FILE_PATH}"
+}
+
+upload_version_file_and_unlock() {
+  local FILE_PATH; FILE_PATH=${BUCKET_PATH}/${STABLE_VERSION_FILE}
```
zerobfd, line 143: This is in multiple places. 1) you should quote "${BUCKET_PATH}/${STABLE_VERSION_FILE}" 2) consider putting it at the top of the file and making it `readonly`.

scripts/asm-installer/release_asm_installer
```diff
@@ -118,6 +127,28 @@ get_stable_version() {
...
+
+  sort -r ${STABLE_VERSION_FILE} | uniq > "tmpfile" && mv "tmpfile" ${STABLE_VERSION_FILE}
```
zerobfd, line 145: This will fail if `tmpfile` exists, and it might not be safe to overwrite it all of the time. Consider

```
TMPFILE="$(mktemp)"
sort ... > "${TMPFILE}" && mv "${TMPFILE}" ...
```

scripts/asm-installer/release_asm_installer
```diff
@@ -158,20 +189,31 @@ write_and_upload() {
   local FILE_PATH; FILE_PATH="${BUCKET_PATH}/${FILE_NAME}"
   local FILE_URI; FILE_URI="${BUCKET_URL}/${FILE_PATH}"
 
-  if ! changes_necessary "${FILE_URI}" "${FILE_NAME}"; then return; fi
```
zerobfd, line 161: I'm not sure why this got refactored, can you explain?

scripts/asm-installer/release_asm_installer
```diff
@@ -158,20 +189,31 @@ write_and_upload() {
...
+      upload "${SCRIPT_NAME}" "${FILE_NAME}" "${BUCKET_PATH}/${SCRIPT_NAME}" "${BUCKET_URL}/${BUCKET_PATH}/${SCRIPT_NAME}"
+    fi
 
-  sha256sum "${SCRIPT_NAME}" >| "${SCRIPT_NAME}.sha256"
-  cp "${SCRIPT_NAME}" "${FILE_NAME}"
-  sha256sum "${FILE_NAME}" >> "${SCRIPT_NAME}.sha256"
+    git restore "${SCRIPT_NAME}"
 
-  upload "${SCRIPT_NAME}" "${FILE_NAME}" "${FILE_PATH}" "${FILE_URI}"
+    append_version "$(./"${SCRIPT_NAME}" --version)" "${FILE_NAME}"
```
zerobfd, line 205: At first, I don't think we want to do this every time, just if it's on the release branch. i.e. whenever we check for a git tag.

scripts/asm-installer/release_asm_installer
zerobfd: Almost certainly don't want EXIT or else it would run even if it exits successfully. I think ERR includes SIGINT.

scripts/asm-installer/release_asm_installer
zerobfd: Actually now I realize that this is already at the top of the file, so just `readonly` that and use it.

zerobfd (changes requested):
What happens if the file already has a hold on it when `get_version_file_and_lock` is called?

scripts/asm-installer/release_asm_installer
ZhengzheYang: Thanks for the advice! As for 3, I use `true >| $FILE` because the `curl -O ...` will create that file, with the xml error message inside as the file content if the curl failed. Therefore, I meant to override/truncate that file.

scripts/asm-installer/release_asm_installer
ZhengzheYang: I am not not sure if this is the best way to do it, but now the script has an option to write all the versions (with -a flag), as opposed to only write the versions that contains updates. Therefore, if there's no necessary changes, I have an `elif` clause to check if we would like to append the version regardless. Does this make sense?

scripts/asm-installer/release_asm_installer
ZhengzheYang: So you mean we don't append the versions if it's unstable, staging, etc.?

scripts/asm-installer/release_asm_installer
zerobfd: Good point on `true >| $file`, I didn't think about that.

scripts/asm-installer/release_asm_installer
zerobfd: Right

scripts/asm-installer/release_asm_installer
zerobfd: What use case are you imagining for that flag?

ZhengzheYang: I'm thinking the case when we want to create that file for the first time or re-create that file.  I did not assume that we already have that file so that's why I have this flag. If we could assume that we will never delete that file, like you commented above, then I think it's safe to not have this flag. Now that I think about it, the creation of the file is not complete anyway: it only has the version with the largest config number for each release. We probably will need some manual or separate scripting to backfill all the versions.

zerobfd: Yep, I agree that a short script to initially backfill is a good idea. It's probably pretty small, but we should commit it just for transparency + code review.

scripts/asm-installer/release_asm_installer
ZhengzheYang: Done. Forgot that I have defined that on the top..

scripts/asm-installer/release_asm_installer
ZhengzheYang: Done

scripts/asm-installer/release_asm_installer
ZhengzheYang: Will work on the script.

scripts/asm-installer/release_asm_installer
ZhengzheYang: Done

ZhengzheYang:
> What happens if the file already has a hold on it when `get_version_file_and_lock` is called?

The action of putting a hold is idempotent. It basically edits the metadata so it should be fine.

zerobfd:
> > What happens if the file already has a hold on it when `get_version_file_and_lock` is called?
> 
> The action of putting a hold is idempotent. It basically edits the metadata so it should be fine.

It *should* fail, right? If we're using it as a lock, we don't want two things to be able to have the hold at one time.

ZhengzheYang:
> > > What happens if the file already has a hold on it when `get_version_file_and_lock` is called?
> > 
> > 
> > The action of putting a hold is idempotent. It basically edits the metadata so it should be fine.
> 
> It _should_ fail, right? If we're using it as a lock, we don't want two things to be able to have the hold at one time.

You are right. Will update.

ZhengzheYang:
@zerobfd updated. Take your time!

scripts/release-asm/common.sh
```diff
@@ -170,10 +211,23 @@ write_and_upload() {
     upload "${SCRIPT_NAME}" "${FILE_NAME}" "${BUCKET_PATH}/${SCRIPT_NAME}" "${BUCKET_URL}/${BUCKET_PATH}/${SCRIPT_NAME}"
   fi
 
+  if [[ "${BRANCH_NAME}" =~ "release" ]]; then
+    append_version "${VERSION}" "${FILE_NAME}"
+  elif is_proper_tag "${VERSION}"; then
+    append_version "${VERSION}" "${FILE_NAME}"
+  fi
```
zerobfd, line 218: Could this just be

```
if is_proper_tag "${VERSION}" || [[ "${BRANCH_NAME}" =~ "release" ]]; then
```

zerobfd (approved):

scripts/release-asm/common.sh
ZhengzheYang: I spent at least 15 minutes trying to write ``` if [[ "${BRANCH_NAME}" =~ "release" || is_proper_tag "${VERSION}"  ]]; then``` and failed to make the shellchecker happy.

scripts/release-asm/common.sh
```diff
@@ -72,6 +62,30 @@ $(other_releases)
 EOF
 }
 
+is_proper_tag() {
+  local TAG; TAG="${1}"
+  if [[ "${TAG}" =~ ^[0-9]+\.[0-9]+\.[0-9]+-asm\.[0-9]+\+config[0-9]+$ ]]; then
+    true
+  else
+    false
```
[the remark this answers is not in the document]
ZhengzheYang: Done

PR #583, opened 2021-04-08
added the flag to create the namespace

ZhengzheYang:
Fixed a bug that the `istio-namespace` will be created during validation. Also added a flag so that the user could choose to have this namespace created for them.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -238,10 +238,12 @@ main() {
     exit_if_not_cluster_admin
   fi
 
-  if can_create_namespace; then
-    create_istio_namespace
-  elif should_validate; then
-    exit_if_istio_namespace_not_exists
+  if ! is_managed; then
+    if can_create_namespace; then
+      create_istio_namespace
+    elif should_validate; then
+      exit_if_istio_namespace_not_exists
+    fi
   fi
 
   if [[ "${ONLY_VALIDATE}" -eq 1 ]]; then
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -572,8 +572,7 @@ can_register_cluster() {
 can_create_namespace() {
   if ! can_modify_at_all; then false; return; fi
 
-  if [[ "${ENABLE_ALL}" -eq 1 || "${ENABLE_NAMESPACE_CREATION}" -eq 1 ]] \
-    && ! is_managed; then
+  if [[ "${ENABLE_ALL}" -eq 1 || "${ENABLE_NAMESPACE_CREATION}" -eq 1 ]]; then
     true
   else
     false
@@ -639,6 +638,7 @@ FLAGS:
      --enable_gcp_iam_roles
      --enable_gcp_components
      --enable_registration
+     --enable_namespace_creation
 
      --managed
 
@@ -750,6 +750,8 @@ FLAGS:
                                       managed services and components
      --enable_registration            Allow the script to register the cluster
                                       to an environ
+     --enable_namespace_creation      Allow the script to create the istio-system
+                                      namespace for the user
 
      --managed                        Provision a remote, managed control plane
                                       instead of installing one in-cluster.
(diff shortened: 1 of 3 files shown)


scripts/asm-installer/install_asm
```diff
@@ -2206,11 +2224,20 @@ EOF
...
+}
+
+exit_if_istio_namespace_not_exists() {
   info "Checking for istio-system namespace..."
   if ! istio_namespace_exists; then
-    info "Creating istio-system namespace..."
-    retry 2 kubectl create ns istio-system
+    { read -r -d '' MSG; fatal "${MSG}"; } <<EOF || true
+Current user must have the istio-system namespace created.
+Please create the "istio-namespace" and retry, or run the script with the
```
zerobfd, line 2237: Rewording:

```
The istio-system namespace doesn't exist. Please create the istio-system namespace
and retry, or run the script with the...
```

scripts/asm-installer/install_asm
```diff
@@ -566,6 +569,17 @@ can_register_cluster() {
   fi
 }
 
+can_create_namespace() {
+  if ! can_modify_at_all; then false; return; fi
+
+  if [[ "${ENABLE_ALL}" -eq 1 || "${ENABLE_NAMESPACE_CREATION}" -eq 1 ]] \
+    && ! is_managed; then
```
zerobfd, line 576: I think we still need the namespace created even with MCP, we can drop the `&& ! is_managed`.

scripts/asm-installer/install_asm
```diff
@@ -861,6 +875,10 @@ parse_args() {
         ENABLE_REGISTRATION=1
         shift 1
         ;;
+      --enable_namespace_creation | --enable-namespace-creation)
```
zerobfd, line 878: We need to also add this to the two usage prompts.

zerobfd (changes requested):

scripts/asm-installer/install_asm
ZhengzheYang: Done

scripts/asm-installer/install_asm
ZhengzheYang: Updated. I included this because of [this line](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm#L240).

scripts/asm-installer/install_asm
ZhengzheYang: Done

ZhengzheYang:
@zerobfd could you please take a look at the [fake kubectl handler](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/583/files#diff-bdb3c0285b1cc5b15ca4726eb777ff17366ae16eb5c6e576433e932f9e59b7b7R48-R60)? Let me know if it makes sense or if there are better ways. I have to make this change so that the `run-cli-tests` could pass.

scripts/asm-installer/install_asm
```diff
@@ -2206,11 +2226,20 @@ EOF
   fi
 }
 
-ensure_istio_namespace_exists() {
+create_istio_namespace() {
+  info "Creating istio-system namespace..."
+  retry 2 kubectl create ns istio-system
```
zerobfd, line 2231: I think we should probably also check to see if it exists here and just return if it does.

zerobfd (approved):

scripts/asm-installer/install_asm
ZhengzheYang: Done

scripts/asm-installer/install_asm
```diff
@@ -237,8 +238,12 @@ main() {
     exit_if_not_cluster_admin
   fi
 
-  if should_validate && ! is_managed; then
-    ensure_istio_namespace_exists
+  if ! is_managed; then
```
ZhengzheYang, line 241: Added this check to be consistent with the logic before this PR and avoid creating the ns which might break MCP.

ZhengzheYang:
running test again to ensure it's not flaky

PR #584, opened 2021-04-08
update to simulate the previous validation logic

ZhengzheYang:
After #577 is introduced, `--only_validate` does not work anymore and it would misleadingly appear to the user that most of the IAM roles are not required anymore if they try to validate only (since we added logic to only conditionally require certain roles). This PR will simulate the previous validation logic that basically list all the required IAM roles. User could use this as a reference to the complete list of IAM roles required.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -1804,25 +1804,26 @@ EOF
 required_iam_roles() {
   # meshconfig.admin - required for init, stackdriver, UI elements, etc.
   # servicemanagement.admin/serviceusage.serviceUsageAdmin - enables APIs
-  if can_modify_gcp_components || \
+  if should_validate || \
+     can_modify_gcp_components || \
      can_modify_cluster_labels || \
      can_modify_cluster_roles; then
     echo roles/container.admin
   fi
-  if can_modify_gcp_components; then
+  if should_validate || can_modify_gcp_components; then
     echo roles/meshconfig.admin
   fi
-  if can_modify_gcp_apis; then
+  if should_validate || can_modify_gcp_apis; then
     echo roles/servicemanagement.admin
     echo roles/serviceusage.serviceUsageAdmin
   fi
-  if can_modify_gcp_iam_roles; then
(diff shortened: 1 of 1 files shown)


ZhengzheYang:
@zerobfd the reason I try to simulate the previous validation logic is that 1) user might not remember all the roles they need so passing flags and checking might not be the best user experience. Having a complete list of required roles seems more reasonable to me if I were to validate if I have all the required roles. and 2) this will not introduce breaking changes to the user behavior when they try to validate. If we validate only with `--enable-*` flags, we would require the user to adapt and we need to update the docs. If the these two reasons do not make sense or you disagree please let me know!

scripts/asm-installer/install_asm
```diff
@@ -1804,25 +1804,26 @@ EOF
 required_iam_roles() {
   # meshconfig.admin - required for init, stackdriver, UI elements, etc.
   # servicemanagement.admin/serviceusage.serviceUsageAdmin - enables APIs
-  if can_modify_gcp_components || \
+  if should_validate || \
```
zerobfd, line 1807: `should_validate` will be true most of the time, so this practically gets rid of the conditions for the IAM roles.

What we really want to check is `[[ "${ONLY_VALIDATE}" -eq 1 ]] && [[ "${ENABLE_ALL}" -eq 1 || "${ENABLE_FOO}" -eq 1`. I can't really think of a decent way to do that cleanly.

zerobfd (changes requested):
The more I think about this, the more I think this would probably be a lot easier to fix after refactoring a bit. What do you think about leaving it as-is for now and revisiting it in a few weeks or so?

scripts/asm-installer/install_asm
ZhengzheYang: Oh right, I meant to check `ONLY_VALIDATE` case. I create this PR to show a potential direction to fix. I agree that it should be easier after refactoring. Closing this PR.

ZhengzheYang:
reopen this to test the CI

PR #585, opened 2021-04-08
fix CI test

ZhengzheYang:
#580 refactored the release scripts which caused the CI test to fail.

### DIFF
--- a/scripts/asm-installer/cloudbuild-master.yaml
+++ b/scripts/asm-installer/cloudbuild-master.yaml
@@ -78,7 +78,7 @@ steps:
     args:
       - '-c'
       - >
-        ./release_asm_installer
+        ./../release-asm/release_asm_installer
     env:
       - '_DEBUG=1'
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/cloudbuild-master.yaml
```diff
@@ -78,7 +78,7 @@ steps:
     args:
       - '-c'
       - >
-        ./release_asm_installer
+        ./../release-asm/release_asm_installer
```
jasonwzm, line 81: You can directly use `../release-asm/release_asm_installer`

jasonwzm (changes requested):

scripts/asm-installer/cloudbuild-master.yaml
ZhengzheYang: Done

zerobfd (approved):

PR #590, opened 2021-04-12
rename the custom vm instance 

FENGXIANGLI:
add function to remove duplicate image instances and instance templates and rename the image name

### DIFF
--- a/scripts/asm-installer/tests/common.sh
+++ b/scripts/asm-installer/tests/common.sh
@@ -843,7 +843,12 @@ create_source_instance_template() {
 
 create_custom_source_instance_template() {
   echo "Creating custom source instance template ${CUSTOM_SOURCE_INSTANCE_TEMPLATE_NAME}..."
-
+  exist_instance=$(gcloud compute instances list --filter="name=${CUSTOM_IMAGE_NAME}")
+  if [[ "${exist_instance}" == "" ]]
+  then
+    gcloud compute instances delete "${CUSTOM_IMAGE_NAME}" --zone "${CUSTOM_IMAGE_LOCATION}" \
+    --project "${PROJECT_ID}" --quiet
+  fi
   gcloud compute instances create "${CUSTOM_IMAGE_NAME}" \
     --project "${PROJECT_ID}" \
     --zone "${CUSTOM_IMAGE_LOCATION}"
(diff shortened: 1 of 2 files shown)


scripts/asm-installer/tests/common.sh
```diff
@@ -843,7 +843,12 @@ create_source_instance_template() {
 
 create_custom_source_instance_template() {
   echo "Creating custom source instance template ${CUSTOM_SOURCE_INSTANCE_TEMPLATE_NAME}..."
-
+  exist_instance=$(gcloud compute instances list --filter="name=${CUSTOM_IMAGE_NAME}")
```
jasonwzm, line 846: Instead of cleaning up in each test run, can we clean up all the old instance template in https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/tests/setup_longterm_cluster#L34 which we are already doing?

jasonwzm (changes requested):

scripts/asm-installer/tests/common.sh
FENGXIANGLI: I don't have permission to do that

FENGXIANGLI: resolved

FENGXIANGLI:
@googlebot I signed it

scripts/asm-installer/tests/run_basic_instance_template_creation
```diff
@@ -76,6 +76,10 @@ spec:
     serviceAccount: ${WORKLOAD_SERVICE_ACCOUNT}
 EOF
 
+  cleanup_old_instance_templates
```
jasonwzm, line 79: Can we make sure the cleanups all happen at the same step? e.g. https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/tests/setup_longterm_cluster#L34

It is a lot easier for debugging and understand what each step is doing.

jasonwzm (changes requested):

williamaronli:
another PR merged close this one: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/594

PR #591, opened 2021-04-12
Add injected gateway yaml for MCP

ruigulala:


zerobfd (commented):
Does this work for just MCP or any installation?

ruigulala:
> Does this work for just MCP or any installation?

It'll work for any installation (with the proper revision argument substitution).

ruigulala:
Based on our discussion, we decided to *not* add the default injected gateway file here to encourage users manage their own gateway. Abandon this pull request.

Issue #588, opened 2021-04-12
is_user_cluster_admin check fails if only Google IAM role is given


sebastian-a-neb-db:
Hey everyone,
we have a service account running the `install_asm` script. The Service account itself has the Google Cloud IAM role `roles/container.admin` which is listed as requirement [here](https://cloud.google.com/service-mesh/docs/installation-permissions). This means that this service account is cluster-admin in the cluster and can change all API objects within the cluster. Nevertheless the check `is_user_cluster_admin` in the `install_asm` script only checks for clusterrolebindings within the cluster and therefore this check fails.

This is in my opinion a bad behaviour. The check should not fail if the service account has the Google Cloud IAM `roles/container.admin` role.

Any other opinions on this or any chance to fix this?

sebastian-a-neb-db:
Apart from that: I tried validating the installation with `--only_validate` and still could see the log lines:

```bash
install_asm: Creating istio-system namespace...
namespace/istio-system created
```

I would not expect any changes on the cluster and therefore a creation of the `istio-system` namespace with the `--only_validate` flag.

zerobfd:
That's a good point about the IAM roles, thanks for reporting.

The `istio-system` bug just got fixed in #583, and will be in the next release.

sebastian-a-neb-db:
Thanks for the quick reply and the PR. Looking forward for the fix :)

zerobfd:
This was released in the last batch a couple of days ago. Thanks again for reporting!

PR #592, opened 2021-04-12
count "roles/container.admin" as cluster admin

ZhengzheYang:
this in addition allows "roles/container.admin" as the cluster admin. Fixing #588

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -2192,21 +2192,23 @@ EOF
 
 is_user_cluster_admin() {
   local GCLOUD_USER; GCLOUD_USER="$(gcloud config get-value core/account)"
-  local IAM_USER; IAM_USER="$(iam_user)"
+  local IAM_USER; IAM_USER="$(local_iam_user)"
   local ROLES
-  ROLES=$(cat <<EOF
-$(kubectl get clusterrolebinding \
-  --all-namespaces \
-  -o jsonpath='{range .items[?(@.subjects[0].name=="'"${GCLOUD_USER}"'")]}[{.roleRef.name}]{end}'\
-  2>/dev/null)
-$(gcloud projects \
+  ROLES="$(\
+    kubectl get clusterrolebinding \
+    --all-namespaces \
+    -o jsonpath='{range .items[?(@.subjects[0].name=="'"${GCLOUD_USER}"'")]}[{.roleRef.name}]{end}'\
+    2>/dev/null)"
+  if echo "${ROLES}" | grep -q cluster-admin; then return; fi
+
+  ROLES="$(gcloud projects \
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/install_asm
```diff
@@ -2192,14 +2192,21 @@ EOF
...
+  -o jsonpath='{range .items[?(@.subjects[0].name=="'"${GCLOUD_USER}"'")]}[{.roleRef.name}]{end}'\
+  2>/dev/null)
+$(gcloud projects \
+    get-iam-policy "${PROJECT_ID}" \
+    --flatten='bindings[].members' \
+    --filter="bindings.members:${IAM_USER}" \
+    --format='value(bindings.role)' 2>/dev/null)
+EOF
+)
+  if ! (echo "${ROLES}" | grep -q roles/container.admin) && (echo "${ROLES}" | grep -q cluster-admin); then false; fi
```
ZhengzheYang, line 2209: better way to do this?

scripts/asm-installer/install_asm
```diff
@@ -2192,14 +2192,21 @@ EOF
 
 is_user_cluster_admin() {
   local GCLOUD_USER; GCLOUD_USER="$(gcloud config get-value core/account)"
+  local IAM_USER; IAM_USER="$(iam_user)"
```
zerobfd, line 2195: You should use `local_iam_user` here, since we (almost) always want the identity of the client side and not any remote users.

scripts/asm-installer/install_asm
zerobfd: I would just do the same thing except in two pieces.

```
  ROLES="$(\
    kubectl get clusterrolebinding \
    --all-namespaces \
    -o jsonpath='{range .items[?(@.subjects[0].name=="'"${GCLOUD_USER}"'")]}[{.roleRef.name}]{end}'\
    2>/dev/null)"

    if echo "${ROLES}" | grep -q cluster-admin; then return; fi

    ROLES=$(gcloud projects \
    get-iam-policy "${PROJECT_ID}" \
    --flatten='bindings[].members' \
    --filter="bindings.members:${IAM_USER}" \
    --format='value(bindings.role)' 2>/dev/null)

    if echo "${ROLES}" | grep -q roles/container.admin; then return; fi

    false
```

zerobfd (changes requested):

scripts/asm-installer/install_asm
ZhengzheYang: Smart. Potentially saving a command call. Will update.

scripts/asm-installer/install_asm
ZhengzheYang: Done

zerobfd (approved):

PR #614, opened 2021-04-20
Update repo docs

backward-compatible:


docs/developer-guide.md
```diff
@@ -0,0 +1,23 @@
+## Dev Workflow
+* Checkout the repo (this assumes the basic GitHub setup already completed)
```
zerobfd, line 2: I think the expectation is that people will fork the repo and make PRs from their own repo and not work off of this one.

zerobfd (changes requested):
If you re-write the exact same thing with forking instead of making branches on this repo then it's perfect.

docs/developer-guide.md
```diff
@@ -0,0 +1,23 @@
...
+  ```
+  git clone https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git
+  ```
+
+* Create a new branch (_outlining feature branch workflow as an example here_)
```
ZhengzheYang, line 7: Actually what I would do is to fork the repo and create the branch there in my own repo? Otherwise the asm repo will be filled with all contributors branches since they have to push. Not sure if this is what we wanted.

docs/developer-guide.md
backward-compatible: updated to suggest forking instead

docs/developer-guide.md
backward-compatible: updated to suggest forking instead

zerobfd (approved):

PR #615, opened 2021-04-20
rename the install_asm to asmcli

ZhengzheYang:


zerobfd (changes requested):
Let's move the asmcli folder up to the root instead of putting it under scripts. Also, let's go ahead and rename cloudbuild-master to cloudbuild-main.

ZhengzheYang:
need to update the cloudbuild yamls to cope with the renaming.

ZhengzheYang:
@zerobfd The tests have passed. When you have time please take another look. Thanks!

zerobfd (approved):
Thanks!

PR #618, opened 2021-04-21
integrate bazel and add the preliminary logic to merge scripts

ZhengzheYang:
As decided, we integrate Bazel to Scriptaro v2 to easily manage build targets. Updated dev guide for instructions. Follow the guide to build and test the merged standalone script. Currently we don't have other files to include other than `asmcli`, but this is a start.


ZhengzheYang:
@zerobfd Hi Clay, the check has passed so if you have time you could take a look. I updated the cloud build steps to call `bazel build`. The contentious part might be the `printf` commands in `merge.sh`.

merge.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# The first path is the output path, default to in bazel-bin/
+out="${1}"
+shift
+touch "${out}"
+
+{
+  printf "%s\n" "#!/usr/bin/env bash" "set -CeE" "set -o pipefail";
+  cat "$@";
+  printf "\nmain \"\${@}\"\n";
```
zerobfd, line 13: Can we do something similar with one command?

```
cat <<EOF >>"${out}"
#!/usr/bin/env bash
set -CeE
set -o pipefail

$@

main "\${@}"
EOF
```

docs/developer-guide.md
```diff
@@ -5,3 +5,28 @@
...
+bazel clean --expunge
+```
+or to cleanup asynchronously:
+```shell
+bazel clean --expunge_async
+```
+
+## During Development
+* **Important**: make sure all global variables and functions are `readonly` to prevent unexpected overwriting.
+* Don't add `#!/usr/bin/env bash` or equivalanet lines to any modules/files. This line will be added during the compilation.
```
zerobfd, line 30: s/equivalanet/equivalent

merge.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
zerobfd, line 1: Executables shouldn't have a .sh extension

asmcli/asmcli
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env bash
```
zerobfd, line 1: Non-executables should have a .sh extension

BUILD
```diff
@@ -0,0 +1,17 @@
...
+
+merge(
+    name = "merge",
+    out = "asmcli",
+    chunks = glob(
+        ["asmcli/**/*.sh", "asmcli/asmcli"],
+        exclude = ["**/test*/**"],  # exclude any directory that starts with test
+    ),
+    merge_tool = "//:merge_script"
+)
```
zerobfd, line 12: Is there a way to do this without needing a custom Bazel rule? It feels like we should be able to get away with just a `sh_binary` rule called `merge`.

zerobfd (changes requested):
Is the WORKSPACE file necessary?

merge.sh
ZhengzheYang: Done

docs/developer-guide.md
ZhengzheYang: Done

merge.sh
ZhengzheYang: Done

asmcli/asmcli
ZhengzheYang: Done

ZhengzheYang:
`WORKSPACE` is required to mark the root of a package. This file could be empty but could also be used to link external packages. See https://docs.bazel.build/versions/master/build-ref.html#packages_targets:

> A workspace is a directory on your filesystem that contains the source files for the software you want to build. Each workspace directory has a text file named WORKSPACE which may be empty, or may contain references to external dependencies required to build the outputs.
Directories containing a file called WORKSPACE are considered the root of a workspace. Therefore, Bazel ignores any directory trees in a workspace rooted at a subdirectory containing a WORKSPACE file (as they form another workspace).

BUILD
ZhengzheYang: I looked around more and I don't think `sh_binary` would work since the native build targets are used for building or compiling, whereas what we want to do here is actually executing an executable (any executable not limited to bash). If we build a `sh_binary`, all it does is bring all dependencies to the `bazel-bin` directory. Shell does not need additionally building since we could just execute that. 

Similarly in other languages (`Java`, `Golang`, etc.), the process of building a target is generating an executable, but what we want to do here is to *execute* an executable. I followed this [example](https://github.com/bazelbuild/examples/tree/master/rules/actions_run), but I think in additions to `ctx.actions.run`, we could also use `ctx.actions.executable`. 

In summary, I think having a custom rule is necessary to fit in our goal, but if ever we found out that a single `sh_binary` would work, I will be responsible for switching to using that, which shouldn't be too hard. Does this sound reasonable?

BUILD
```diff
@@ -0,0 +1,13 @@
...
+MERGE_OUT = ["asmcli"]
+
+genrule(
+    name = "merge",
+    srcs = ALL_MODULES,
+    outs = MERGE_OUT,
+    cmd = "$(location scripts/release-asm/merge) $(OUTS) $(SRCS)",
+    tools = ["scripts/release-asm/merge"],
+    executable = True,
+)
```
zerobfd, line 13: fix the no newline at end of file

zerobfd (changes requested):
I like genrule a lot better. Can we suffix the `asmcli` release with the full version? e.g. generate `asmcli_1.9.2-asm.3+config.3`?

BUILD
ZhengzheYang: Done

BUILD
ZhengzheYang: Updated to use `genrule`

PR #620, opened 2021-04-22
Configure MCP to use GKE Connect

zerobfd:


zerobfd:
Unfortunately the MCP test is going to fail because it's going to use the config from the master branch instead of this commit. I tested locally and it worked fine.

costinm (approved):
Thanks !

PR #621, opened 2021-04-22
Add timestamps if ts available and verbose mode

zerobfd:


scripts/asm-installer/install_asm
```diff
@@ -470,7 +476,11 @@ error() {
 }
 
 info() {
-  echo "${SCRIPT_NAME}: ${1}" >&2
+  if [[ "${VERBOSE}" -eq 1 && "${HAS_TS}" -eq 1 ]]; then
+    echo "${SCRIPT_NAME}: ${1}" | ts >&2
```
howardjohn, line 480: I vaguely recall the time format here being really bad (didn't include seconds?). I have an alias with a subjectively better format configured I can comment tomorrow.

But any time is better than no time

zerobfd: I should have put an example in the PR, my bad.

```
Apr 22 01:46:11 install_asm: Setting up necessary files...
Apr 22 01:46:11 install_asm: Creating temp directory...
```

jasonwzm (approved):

PR #626, opened 2021-04-24
Force update operation when the version is the same to apply overlays…

rialg:
In order to enable [optional features](https://cloud.google.com/service-mesh/docs/enable-optional-features#egress_gateways), such as the [egress gateway](https://cloud.google.com/service-mesh/docs/scripted-install/gke-upgrade#upgrading_with_an_option), by applying custom overlays to the `IstioOperator`, allow to force an update of the control plane when the ASM version to install is the same as the one in use.


zerobfd:
Can I ask what problem you ran into? This should work without extra flags. i.e. if I do

`install_asm --mode install $OPTS`

and then

`install_asm --mode install $OPTS -o egressgateways`

it should work. If it doesn't, can you copy/paste the output?

Issue #630, opened 2021-04-26
This detection seems to miss our service account because for format is different


matthewkrupnik:
A check if the current account is a service account is failing for us. Our build in CloudBuild is running under the default service account (`xxxxxxxx@cloudbuild.gserviceaccount.com`), which is a service account but does not match the check.
https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/1fa84c122f33afcc966689c66d96168150658abb/scripts/asm-installer/install_asm#L1838

I think the check should be changed to a more broad version:
`if is_sa || [[ "${ACCOUNT_NAME}" = *.gserviceaccount.com ]]; then`

matthewkrupnik:
Now that PR https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/631 is merged, I've tested the updated script and it works for us. Thanks @zerobfd! Closing this issue

PR #631, opened 2021-04-26
Change service account detection pattern

matthewkrupnik:
Using a more broad pattern for detecting if current gcloud account is a service account or not. Old pattern was missing *@cloudbuild.gserviceaccount.com.

Proposed for for https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/issues/630


matthewkrupnik:
@googlebot I signed it!

zerobfd:
Thanks for the PR! Change looks fine. I'm re-running the test, and I'll accept when it passes.

PR #633, opened 2021-04-27
install_asm: allow setting GCLOUD_USER_OR_SA

### LINKED ISSUE #632: Unable to set user or service account in ASM installation script
Currently, `asm_install` has a variable to specify the user or service account: `GCLOUD_USER_OR_SA`.

The first section of the `local_iam_user` function checks whether this variable is set. If it is, it returns the value. However, the value is always empty, given that it is initialized to the empty string. 

Assuming the intent of the variable is for users to be able to specify the user or service account, it seems we need to modify the script to allow for this.

alexbrand:
Fixes #632


zerobfd:
That wasn't the original intent for the variable, but looking at the PR it makes sense. Was there a specific case that this was necessary for? This will probably live on for awhile as a not-very-documented feature for power users. If there's a concrete issue motivating this PR, I'd like to solve it in a way that doesn't require manually specifying it.

alexbrand:
Thanks @zerobfd! The motivation is to enable the use/impersonation of a service account when using the terraform module. I opened an issue in that repo as well: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/879

I'm open to alternatives if you think there's a better approach.

zerobfd (approved):

zerobfd:
LGTM. I can't merge it until the CLA is accepted, but as soon as I get a notification I'll get it in.

alexbrand:
Let's see if this works.

@googlebot I signed it!

PR #634, opened 2021-04-27
intial integration of bats

ZhengzheYang:
This PR tries to add unit testing framework to Scriptaro. It assumes the developers has `npm` which we use to manage the versions of `bats`. The tests serve as POC and should be easy to extend further with more complex tests.


asmcli/asmcli.sh
```diff
@@ -335,7 +335,7 @@ init() {
 # If DRY_RUN or VERBOSE is enabled, it will print the command, and if DRY_RUN is
 # not enabled it runs the command.
 #######
-run() {
+_run() {
```
ZhengzheYang, line 338: I made this change because `run` is a `bats` thing to run commands. However, this breaks the cloud build CI. Would need some other way to bypass this.

BUILD
```diff
@@ -1,13 +1,37 @@
 package(default_visibility = ["//visibility:public"])
 
-ALL_MODULES = glob(["asmcli/**/*.sh"], exclude = ["**/test*/**"],)
+ASMCLI_MODULES = glob(["asmcli/**"])
+NODE_MODULES = glob(["node_modules/**"])
+ASMCLI_SOURCES = glob(["asmcli/**/*.sh"], exclude = ["**/test*/**"],)
+ALL_UNIT_TESTS = glob(["asmcli/tests/*.bats"])
```
zerobfd, line 6: Let's just say TESTS in case we decide to put some non unit ones in there.

BUILD
```diff
@@ -1,13 +1,37 @@
...
+    data = ASMCLI_MODULES,
+)
+
+sh_library(
+    name = "node_modules",
+    data = NODE_MODULES,
+)
+
+sh_test(
+    name = "unit_test",
```
zerobfd, line 29: I think these are traditionally just called "test", but if you saw another example somewhere ignore me.

asmcli/asmcli.sh
zerobfd: Don't you need to change all occurrences in the entire file?

package.json
```diff
@@ -0,0 +1,10 @@
+{
+    "private": true,
+    "devDependencies": {
+        "bats": "^1.3.0",
+        "bats-assert": "^2.0.0"
+    },
+    "scripts": {
+        "test": "bazel test ..."
+    }
```
zerobfd, line 9: I don't think we need to worry about enabling `npm test`. If we do, then it shouldn't just proxy to `bazel` since devs might not have it installed.

BUILD
```diff
@@ -1,13 +1,37 @@
...
+    size = "small",
+    srcs = ["scripts/tests/run_tests"],
```
zerobfd, line 31: Instead of having a specific script, can we just do something like

```
sh_test(
  name = "test",
  ...
  env = {"SOURCE_FILES" : ASMCLI_SOURCES},
  srcs = ["node_packages/bats/bin/bats"],
  args = ALL_UNIT_TESTS
```

zerobfd (changes requested):

asmcli/asmcli.sh
ZhengzheYang: Are you ok with the naming? Should I name it something else?

BUILD
ZhengzheYang: Done

BUILD
ZhengzheYang: Done

package.json
ZhengzheYang: This seems to be auto-generated. I will remove it.

BUILD
ZhengzheYang: This is way too smart. Will experiment to see if it works. Thanks!

asmcli/asmcli.sh
zerobfd: Now that I look at it, it seems like starting identifiers with an underscore isn't valid all the time. Maybe change it to "do" instead?

ZhengzheYang: I changed it to `run_commands`.  I'll see if the tests passed. `do` seems to be a keyword in bash?

BUILD
```diff
@@ -1,13 +1,38 @@
...
+    name = "test",
+    size = "small",
+    srcs = ["node_modules/bats/bin/bats"],
+    deps = [
+            ":asmcli_modules",
+            ":node_modules",
+            ],
+    env = {"SOURCE_FILES": '\n'.join(ASMCLI_SOURCES)},
+    args = TESTS,
+)
```
zerobfd, line 38: fix no newline at end of file

scripts/tests/run_tests
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
zerobfd, line 1: We don't need this file anymore, right?

zerobfd (changes requested):

scripts/tests/run_tests
ZhengzheYang: No we don't. When the tests passed I will clean up!

BUILD
ZhengzheYang: Done

ZhengzheYang:
@backward-compatible I added you as a reviewer and will keep doing this to keep you in the loop.

asmcli/tests/main_test.bats
```diff
@@ -0,0 +1,21 @@
...
+teardown() {
+  echo "Cleaned up"
+}
+
+@test "test main method --version" {
+  run main --version
+  assert_success
+}
+
+@test "test main method print usage" {
```
zerobfd, line 18: I think "test main method with no arguments returns nonzero exit code" is more descriptive of what's actually being tested here.

zerobfd (approved):

asmcli/tests/main_test.bats
ZhengzheYang: Done

PR #635, opened 2021-04-27
NEG options

mathieu-benoit:
The intent is to have an easy way/option/snippet to setup a NEG/GCLB in front of ASM to be able to leverage then Cloud Armor, Managed Certificates, etc.
- NEG annotation
- NEG and `BackendConfig` annotations

Like discussed with @ameer00 and @theemadnes, we may want to reuse this for the refresh of this [Edge to Mesh](https://cloud.google.com/architecture/exposing-service-mesh-apps-through-gke-ingress) doc we are working on.


mathieu-benoit:
Ready for your review and very open to feedback and suggestion, first time I'm contributing to this repo. Thanks!

zerobfd (approved):

mathieu-benoit:
Tests:

```
kpt cfg set asm gcloud.container.cluster.ingress.backendConfig "{"default": "ingress-test"}"
asm/
set 1 field(s) of setter "gcloud.container.cluster.ingress.backendConfig" to value "{default: ingress-test}"
```

PR #636, opened 2021-04-28
Install istio-cni with securityContext for SELinux

zioproto:
The `istioinstall-cni` container needs a `securityContext` with `Privileged: true` to install `istio-cni` with SELinux enabled.

Related discussion on Istio PR https://github.com/istio/istio/pull/32360

### DIFF
--- a/asm/istio/options/cni-selinux.yaml
+++ b/asm/istio/options/cni-selinux.yaml
@@ -1,0 +1,37 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


bianpengyuan (commented):
lgtm. Not sure what is the criteria to add a new option here.

zerobfd (approved):
Seems reasonable. If the intention is to use this with `install_asm` we should probably add a test.

zioproto:
@zerobfd do you mean implementing a test in the folder `scripts/asm-installer/tests` ?

I am trying to understand if there is at least already a test in that folder for the `cni-gcp.yaml` or `cni-onprem.yaml` options, to use it as a reference, but I cant find anything relevant.

Could you please confirm that the test should be implemented in that folder ? Can you confirm that the test is necessary here to merge this PR ? thanks

zerobfd:
Test is not necessary for this PR. If it's tested in Istio OSS then it's probably fine there, I'm just not aware of anywhere that we test the SELinux piece specifically.

It was more just a note in general, not directed at you. Sorry that was ambiguous!

PR #641, opened 2021-04-29
Do not explicitly specify COS where not needed

SergeyKanzhelev:
Starting with [GKE node version 1.19](https://cloud.google.com/kubernetes-engine/docs/concepts/using-containerd#:~:text=starting%20with%20gke%20node%20version%201.19), COS image type is deprecated. I suggest to not explicitly specify any parameters where defaults would work.


zerobfd:
Fixed in a different commit, thanks for raising the issue.

PR #644, opened 2021-04-29
NEG options (take 2)

mathieu-benoit:
Take 2 of the PR #635, the latter introduced an error with this command:
```
kpt pkg get https://github.com/mathieu-benoit/anthos-service-mesh-packages.git/asm@patch-1 .
fetching package "/asm" from "https://github.com/mathieu-benoit/anthos-service-mesh-packages" to "asm"
failed to set "gcloud.core.project" automatically in package "asm" with error: json: cannot unmarshal object into Go struct field setter.setter.value of type string
failed setting auto-setter "gcloud.project.projectNumber" in package "asm" with error: json: cannot unmarshal object into Go struct field setter.setter.value of type string
```
This new and current PR fixes this error.

PR #645, opened 2021-04-30
NEG options (take 3)

mathieu-benoit:
Take 3 of the PR #644, the latter introduced an error with this command:
```
kpt cfg list-setters asm
# The output was an empty list
```

Furthermore, I took the opportunity to add a tests section in the developer guide.


mathieu-benoit:
Actually because the value of the `backendConfig` contains `"` I'm not able to make this working properly. For example this command:
```
kpt cfg set asm gcloud.container.cluster.ingress.backendConfig "{"default": "ingress-test"}"
```
Will result with this:
```
kpt cfg list-setters asm
gcloud.container.cluster.ingress.backendConfig            {default:  ingress-backendconfig}
```
So instead of being ` {"default":  "ingress-backendconfig"}` we have this `{default:  ingress-backendconfig}` which will land later with an issue in Kubernetes like this `Translation failed: invalid ingress spec: error getting BackendConfig for port "80" on service "istio-system/istio-ingressgateway", err: BackendConfig annotation is invalid json`.

So not able to make it working... the quesiton is how to properly set a string containing `"`?

mathieu-benoit:
I also tried with substitution, but not working either...

PR #648, opened 2021-04-30
update dev guide for writing/running tests

ZhengzheYang:
Now that the initial `bats` integration is merged, we update the dev guide to document the behavior and instructions.

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -37,10 +37,10 @@ We integrated [`bats`](https://github.com/bats-core/bats-core) as our
 testing system. To add tests, read the
 [official documentation](https://bats-core.readthedocs.io/en/latest/)
 and update/add `*.bats` files to the `asmcli/tests` folder.
-If there are any set-ups that could be shared among tests, update
+If there's any set-up that could be shared among tests, update
 `unit_test_common.bash` in the test folder.
 
-The `//:test` target will by default run all the tests. If you want
+The `:test` target will by default run all the tests. If you want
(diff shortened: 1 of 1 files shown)


docs/developer-guide.md
```diff
@@ -30,3 +30,28 @@ bazel clean --expunge_async
...
 * Put all code in functions. The merge logic will add the call to `main`. In other words, don't add any entry point to actually
 execute the script.
+
+
+## Writing Tests
+We integrated [`bats`](https://github.com/bats-core/bats-core) as our
+testing system. To add tests, read the
+[official documentation](https://bats-core.readthedocs.io/en/latest/)
+and update/add `*.bats` files to the `asmcli/tests` folder.
+If there are any set-ups that could be shared among tests, update
```
zerobfd, line 40: Iif there is any set-up that could be shared..."

docs/developer-guide.md
```diff
@@ -30,3 +30,28 @@ bazel clean --expunge_async
...
+`unit_test_common.bash` in the test folder.
+
+The `//:test` target will by default run all the tests. If you want
```
zerobfd, line 43: Do you need the // before :test ?

zerobfd (approved):

docs/developer-guide.md
ZhengzheYang: We don't, since we only have this one package and no subpackages. I put the "//:" to indicate it's a target. I guess a ":" would do the job as well.

docs/developer-guide.md
ZhengzheYang: Done

PR #650, opened 2021-05-03
add samples directory with online-boutique manifests

cloud-pharaoh:


zerobfd:
Are there any changes from the upstream microservices demo repo? If yes, what's the plan to keep these samples updated as upstream makes changes? If no, why are we copying them here instead of just referencing that repo?

I'm not opposed to the idea, just want to make sure the reasoning is clear and captured in PR/commit history.

cloud-pharaoh:
Hi Clay,

Yes. There are 2 main differences, one is that each service is in its own namespace and using a specific, non-default service account. The intent is to use this for tutorials in docs and we want to make sure we are following best practices.
The upstream repo will at some point in the future have kustomize support to allow for these modifications to be done easily in the tutorials but in the meantime the recommendation from the team was to maintain our own copy.

It is not ideal and we will have to update the images with new releases and make sure the manifests are up-to-date with upstream repo but it is a stop-gap solution for now.

zerobfd (approved):

zerobfd:
Thanks for the context! Merging now. Do you need this in any release branches, or is this just to reference from documentation?

PR #651, opened 2021-05-03
refactor args parsing and validation

ZhengzheYang:
This split the argument parsing and validation to `asmcli/lib/parse.sh`. We interact with the `CONTEXT_JSON` variable in parse and validation just to avoid unnecessary I/O. After arguments are validated, we dump the json to the local workspace in `set_up_local_workspace`.

This changes the behavior of accessing and writing global variables. The source of truth for the arguments will be in the `CONTEXT_JSON` file. All variables will read this file first.

### DIFF
--- a/asmcli/asmcli.sh
+++ b/asmcli/asmcli.sh
@@ -72,62 +72,10 @@ KPT_BRANCH=""
 NAMESPACE_EXISTS=0
 KUBECONFIG_SUPPLIED=0
 
-### Context init JSON ###
-CONTEXT_JSON=$(cat <<EOF
-{
-  "flags": {
-    "PROJECT_ID": "${PROJECT_ID:-}",
-    "CLUSTER_NAME": "${CLUSTER_NAME:-}",
-    "CLUSTER_LOCATION": "${CLUSTER_LOCATION:-}",
-    "MODE": "${MODE:-}",
-    "CA": "${CA:-}",
-    "KUBECONFIG": "${KUBECONFIG_FILE:-}",
-    "CONTEXT": "${CONTEXT:-}",
-    "CUSTOM_OVERLAY": "${CUSTOM_OVERLAY:-}",
-    "OPTIONAL_OVERLAY": "${OPTIONAL_OVERLAY:-}",
-    "ENABLE_ALL": ${ENABLE_ALL:-0},
-    "ENABLE_CLUSTER_ROLES": ${ENABLE_CLUSTER_ROLES:-0},
-    "ENABLE_CLUSTER_LABELS": ${ENABLE_CLUSTER_LABELS:-0},
-    "ENABLE_GCP_APIS": ${ENABLE_GCP_APIS:-0},
-    "ENABLE_GCP_IAM_ROLES": ${ENABLE_GCP_IAM_ROLES:-0},
-    "ENABLE_GCP_COMPONENTS": ${ENABLE_GCP_COMPONENTS:-0},
-    "ENABLE_REGISTRATION": ${ENABLE_REGISTRATION:-0},
-    "ENABLE_NAMESPACE_CREATION": ${ENABLE_NAMESPACE_CREATION:-0},
-    "DISABLE_CANONICAL_SERVICE": ${DISABLE_CANONICAL_SERVICE:-0},
-    "PRINT_CONFIG": ${PRINT_CONFIG:-0},
-    "SERVICE_ACCOUNT": "${SERVICE_ACCOUNT:-}",
-    "KEY_FILE": "${KEY_FILE:-}",
-    "OUTPUT_DIR": "${OUTPUT_DIR:-}",
-    "CA_CERT": "${CA_CERT:-}",
-    "CA_KEY": "${CA_KEY:-}",
-    "CA_ROOT": "${CA_ROOT:-}",
-    "CA_CHAIN": "${CA_CHAIN:-}",
-    "CA_NAME": "${CA_NAME:-}",
-    "DRY_RUN": ${DRY_RUN:-0},
-    "ONLY_VALIDATE": ${ONLY_VALIDATE:-0},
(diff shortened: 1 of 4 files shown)


asmcli/lib/context.sh
```diff
@@ -0,0 +1,83 @@
...
+  # Dump CONTEXT_JSON for the first time
+  # This file from now on will only be read and udpated.
+  echo "${CONTEXT_JSON}" >| CONTEXT.json
+}
+
+context_get-option-from-file() {
+  local FILE; FILE="${1}"
+  local OPTION; OPTION="${2}"
+
+  jq -r --arg FILE "${FILE}" --arg OPTION "${OPTION}" '.flags[$OPTION]' "${FILE}"
```
zerobfd, line 48: I don't think the `--arg FILE "${FILE}"` is needed. (but good thinking on the `--arg`, it helps readability.)

asmcli/lib/parse.sh
```diff
@@ -0,0 +1,390 @@
...
+
+  # make sure to redirect stdout as soon as possible if we're dumping the config
+  local PRINT_CONFIG; PRINT_CONFIG="$(context_get-option-from-variable "${CONTEXT_JSON}" "PRINT_CONFIG")"
+  if [[ "${PRINT_CONFIG}" -eq 1 ]]; then
+    exec 3>&1
+    exec 1>&2
+  fi
+}
+
+validate_args() {
```
zerobfd, line 187: High level comment: I'm imagining that specific components will be responsible for some of these checks. For example, MCP should be responsible for validating that there aren't any custom overlays passed in, and this function should mostly be concerned with validating flags that apply to multiple components or subcommands (like "you can't use --print-config and --enable at the same time").

Nothing needs to necessarily change for this PR, but I wanted to make sure we didn't spend too much time testing or worrying about the format here since it will shrink as more files get split out.

zerobfd (changes requested):
As discussed offline: we should need non-file versions of context option get/sets.

asmcli/asmcli.sh
```diff
@@ -71,64 +71,71 @@ PREVIOUS_RELEASE_LINE=""
...
 
 main() {
   init
-  parse_args "${@}"
 
-  # make sure to redirect stdout as soon as possible if we're dumping the config
-  if [[ "${PRINT_CONFIG}" -eq 1 ]]; then
-    exec 3>&1
-    exec 1>&2
+  # sanity check
```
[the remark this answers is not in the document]
ZhengzheYang: Got rid of it.

asmcli/lib/context.sh
ZhengzheYang: Done

asmcli/lib/parse.sh
ZhengzheYang: Ack

asmcli/lib/context.sh
```diff
@@ -0,0 +1,31 @@
+context_init() {
+  local JSON; JSON="${1}"
+  context_FILE_LOCATION="$(mktemp)"; readonly context_FILE_LOCATION
+  export context_FILE_LOCATION
+
+  echo "${JSON}" >> "${context_FILE_LOCATION}"
```
zerobfd, line 6: Out of curiosity, why >> instead of >|? With `mktemp` it shouldn't matter, but personally >| feels more intuitive since we're initializing a new file and not appending to an existing one.

asmcli/lib/context.sh
```diff
@@ -0,0 +1,31 @@
...
+
+  jq -r --arg OPTION "${OPTION}" '.flags[$OPTION]' "${context_FILE_LOCATION}"
+}
+
+context_set-option() {
+  local OPTION; OPTION="${1}"
+  local VALUE; VALUE="${2}"
+  local TEMP_FILE; TEMP_FILE="$(mktemp)"
+
+  jq --arg OPTION "${OPTION}" --arg VALUE "${VALUE}" \
```
zerobfd, line 20: Can we try and use the `-S` flag when writing values from `jq` so that the keys are sorted? If it doesn't work then let's not block this PR on it, but I think it will be useful to have deterministic output to help debugging and testing.

asmcli/asmcli.sh
```diff
@@ -72,66 +72,63 @@ KPT_BRANCH=""
...
-PRINT_VERSION=0
-CUSTOM_CA=0
-USE_HUB_WIP=0
-USE_VM=0
-ENVIRON_PROJECT_ID=""
-HUB_MEMBERSHIP_ID=""
-CUSTOM_REVISION=0
-WI_ENABLED=0
+### Context init JSON ###
+CONTEXT_JSON=$(cat <<EOF
```
zerobfd, line 76: WDYT about putting this initial skeleton inside of context.sh instead of here? That feels like a natural place, and it should also mean we don't need to copy/paste this into the unit test file later in this PR.

zerobfd (approved):
LGTM. A few suggestions that I'd like to try, but if they don't work immediately or you think they're a bad idea then this is good to merge as-is.

asmcli/lib/context.sh
ZhengzheYang: I just used `>>` because the file already exists. I would change it to `>|`.

asmcli/lib/context.sh
ZhengzheYang: Will try.

asmcli/asmcli.sh
ZhengzheYang: Done

asmcli/lib/context.sh
ZhengzheYang: It worked. Updated the `-S` flag.

ZhengzheYang:
Merging without passing the cloudbuild. The attempts to pass the CI will come in the next few PRs.

zerobfd:
sgtm.

PR #653, opened 2021-05-04
Add support for using kubeconfig

backward-compatible:


asmcli/asmcli.sh
```diff
@@ -1486,7 +1541,10 @@ EOF
 validate_cluster() {
   local RESULT; RESULT=""
 
-  info "Confirming cluster information for ${PROJECT_ID}/${CLUSTER_LOCATION}/${CLUSTER_NAME}..."
+  local CURRENT_CONTEXT="$(kubectl config current-context)"
+  info "Confirming cluster information for ${CURRENT_CONTEXT}"
+  IFS="_" read GKE_PREFIX PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME <<< "${CURRENT_CONTEXT}"
```
zerobfd, line 1546: For compatibility, can we use a heredoc instead of the bash specific herestring? e.g.

```
IFS="_" read -r GKE_PREFIX PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME <<EOF
$(kubectl config current-context)
EOF
```

asmcli/asmcli.sh
```diff
@@ -1025,12 +1048,39 @@ validate_args() {
...
+    fi
+  done <<EOF
 CLUSTER_LOCATION
 CLUSTER_NAME
 PROJECT_ID
-MODE
 EOF
 
+  if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 1 && "${CLUSTER_DETAIL_VALID}" -eq 0 ]]; then
+    fatal_with_usage "Missing one or more required options for [CLUSTER_LOCATION|CLUSTER_NAME|PROJECT_ID]"
```
ZhengzheYang, line 1069: Pardon me if I am mistaken here, but to me we shouldn't fatal here. The point of having `MISSING_ARGS` in my opinion is to give out as many warnings as possible before exiting to save the users' time. Do you think it's reasonable to `warn` here and let `MISSING_ARGS` handle the `fatal_with_usage` case?

asmcli/asmcli.sh
```diff
@@ -70,13 +70,16 @@ RELEASE_LINE=""
 PREVIOUS_RELEASE_LINE=""
 KPT_BRANCH=""
 NAMESPACE_EXISTS=0
+KUBECONFIG_SUPPLIED=0
```
zerobfd, line 73: It seems like the couple of places that this needs to be checked, it could just be replaced with `[[ -n "${KUBECONFIG"} ]]` and we could get rid of this variable. Is there a specific reason to pull it into a global variable?

If it's just for readability, I think having a function is a bit cleaner so data isn't duplicated:

```
kubeconfig_defined() {
  if [[ -z "${KUBECONFIG}" ]]; then false; fi
}
```

asmcli/asmcli.sh
```diff
@@ -70,13 +70,16 @@ RELEASE_LINE=""
...
 
 ### Option variables ###
 PROJECT_ID="${PROJECT_ID:=}"
 CLUSTER_NAME="${CLUSTER_NAME:=}"
 CLUSTER_LOCATION="${CLUSTER_LOCATION:=}"
 MODE="${MODE:=}"
 CA="${CA:=}"
+KUBECONFIG="${KUBECONFIG:=}"
```
zerobfd, line 81: I'm slightly concerned about the variable name here. I think that KUBECONFIG is something that's probably set and exported in shells pretty regularly, which makes me worry that customers running the script might inadvertently pass this without realizing it.

If you don't think it's an issue, feel free to ignore. Otherwise maybe KUBECONFIG_FILE or something is a safer name for the variable/flag?

asmcli/asmcli.sh
```diff
@@ -611,6 +617,8 @@ OPTIONS:
   -l|--cluster_location  <LOCATION>
   -n|--cluster_name      <NAME>
   -p|--project_id        <ID>
+  --kubeconfig           <KUBECONFIG>
+  --context              <CONTEXT>
```
zerobfd, line 621: I imagine these two are going to be used fairly often, what do you think about also accepting `--kc` and `--ctx`?

asmcli/asmcli.sh
```diff
@@ -1272,8 +1322,13 @@ set_up_local_workspace() {
   fi
   pushd "$OUTPUT_DIR" > /dev/null
 
-  KUBECONFIG="asm_kubeconfig"
-  export KUBECONFIG
+  # If KUBECONFIG file is supplied, keep using that.
+  if [[ "${KUBECONFIG_SUPPLIED}" -eq 0 ]]; then
```
ZhengzheYang, line 1326: Is there any way or need to validate the `KUBECONFIG` provided by the users is valid or existing? Or we should just use that until we encounter error?

asmcli/asmcli.sh
```diff
@@ -1272,8 +1322,13 @@ set_up_local_workspace() {
...
+    KUBECONFIG="asm_kubeconfig"
+    export KUBECONFIG
```
ZhengzheYang, line 1328: This is probably a shell question, but why we don't need to export if it's provided?

zerobfd, line 1328: I feel like we should export it regardless to make sure that children processes pick up the right file.

zerobfd (changes requested):
High level looks good. I'm extremely pleased at how short this turned out, thanks for taking the time to be thorough with the up-front work in order to keep the implementation simple.

asmcli/asmcli.sh
backward-compatible: i kept these to match **kubectl** and similar cli flags for the two values.

asmcli/asmcli.sh
backward-compatible: makes sense, thanks.

asmcli/asmcli.sh
backward-compatible: if the file doesn't exist, or if it exists but doesn't contain the proper content, the "testing cluster connectivity" check should capture it down the flow. does that suffice?

asmcli/asmcli.sh
backward-compatible: done!

asmcli/asmcli.sh
backward-compatible: done!

asmcli/asmcli.sh
backward-compatible: we need it to avoid regenerating the credentials inside **validate_cluster** and **configure_kubectl** because **set_up_local_workspace** export a _KUBECONFIG_ before they run.

asmcli/asmcli.sh
backward-compatible: replied in other comment that **KUBECONFIG_SUPPLIED** is needed.

asmcli/asmcli.sh
zerobfd: That makes sense. We can definitely have an internal `KUBECONFIG` variable, but maybe we should explicitly set it to an empty string instead of setting it to the existing value. (maybe just `KUBECONFIG="${KUBECONFIG_FILE:=}"`?)

asmcli/asmcli.sh
zerobfd: Right, I meant

```
--kubeconfig|--kc \t\t <KUBECONFIG>
--context|--ctx \t\t <CONTEXT>
```

asmcli/asmcli.sh
backward-compatible: thanks for pointing that out, rearranged and changed to warn.

asmcli/asmcli.sh
backward-compatible: done!

asmcli/asmcli.sh
backward-compatible: sorry i misunderstood the earlier comment. i agree with the concern and will update it to _KUBECONFIG_FILE_

asmcli/asmcli.sh
ZhengzheYang: Yes!

asmcli/asmcli.sh
backward-compatible: done, changed from **KUBECONFIG** -> **KUBECONFIG_FILE**

zerobfd (approved):

PR #654, opened 2021-05-05
Add validation for the existence of a current-context in the KUBECONFIG when context is not passed explicitly

backward-compatible:


asmcli/asmcli.sh
```diff
@@ -1084,15 +1090,21 @@ EOF
...
-  if [[ -n "${KUBECONFIG}" ]]; then
-    KUBECONFIG_SUPPLIED=1
-  fi
-
   if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 0 && "${KUBECONFIG_SUPPLIED}" -eq 0 ]]; then
     MISSING_ARGS=1
     warn "At least one of the following is required: 1) --kubeconfig or 2) --cluster_location, --cluster_name, --project_id"
   fi
 
+  if [[ "${KUBECONFIG_SUPPLIED}" -eq 1 && "${CONTEXT}" -eq 0 ]]; then
```
zerobfd, line 1098: I think this should be `-z "${CONTEXT}"` and not `"${CONTEXT}" -eq 0`.

asmcli/asmcli.sh
```diff
@@ -1061,7 +1062,7 @@ validate_args() {
 MODE
 EOF
 
-  local CLUSTER_DETAIL_SUPPLIED=0
+  local CLUSTER_DETAIL_SUPPLIED=0 # gke cluster param usage intended
```
zerobfd, line 1065: Since this variable is set to 0 and KUBECONFIG_SUPPLIED is set to 1 below, maybe the comments should be less parallel?  "whether gke cluster params were passed via CLI" and "whether a kubeconfig file was supplied" are clearer to me.

zerobfd (approved):
lgtm, two non blockers

asmcli/asmcli.sh
backward-compatible: corrected!

ZhengzheYang (approved):

asmcli/asmcli.sh
backward-compatible: done! made the comments consistent.

PR #656, opened 2021-05-06
Add CNI for MCP

zerobfd:


zerobfd:
Relevant lines from the Cloud Build logs for reviewers:

```
Step #11 - "run-basic-suite-managed": install_asm: Running: '/usr/bin/kubectl -n istio-system get --ignore-not-found cm asm-options -o jsonpath={.data.ASM_OPTS}'
Step #11 - "run-basic-suite-managed": install_asm: -------------
Step #11 - "run-basic-suite-managed": install_asm: Configuring control plane options...
Step #11 - "run-basic-suite-managed": install_asm: Running: 'kubectl apply --overwrite=true -f -'
Step #11 - "run-basic-suite-managed": install_asm: -------------
Step #11 - "run-basic-suite-managed": install_asm: Running: '/usr/bin/kubectl apply --overwrite=true -f -'
Step #11 - "run-basic-suite-managed": install_asm: -------------
Step #11 - "run-basic-suite-managed": configmap/asm-options created
```
...
```
Step #11 - "run-basic-suite-managed": install_asm: Installing CNI into the cluster...
Step #11 - "run-basic-suite-managed": install_asm: Running: '/usr/bin/kubectl apply -f asm/istio/options/cni-managed.yaml'
Step #11 - "run-basic-suite-managed": install_asm: -------------
Step #11 - "run-basic-suite-managed": serviceaccount/istio-cni created
Step #11 - "run-basic-suite-managed": clusterrole.rbac.authorization.k8s.io/istio-cni created
Step #11 - "run-basic-suite-managed": clusterrole.rbac.authorization.k8s.io/istio-cni-repair-role created
Step #11 - "run-basic-suite-managed": clusterrolebinding.rbac.authorization.k8s.io/istio-cni created
Step #11 - "run-basic-suite-managed": clusterrolebinding.rbac.authorization.k8s.io/istio-cni-repair-rolebinding created
Step #11 - "run-basic-suite-managed": configmap/istio-cni-config created
Step #11 - "run-basic-suite-managed": daemonset.apps/istio-cni-node created
```

scripts/asm-installer/install_asm
```diff
@@ -2563,6 +2594,11 @@ scrape_managed_urls() {
 }
 
 install_managed_components() {
+  if [[ "${USE_MCP_CNI}" -eq 1 ]]; then
+    info "Installing CNI into the cluster..."
+    kubectl apply -f "${MANAGED_CNI}"
```
howardjohn, line 2599: MCP checks for *pods* with `k8s-app=istio-cni-node`. I think this introduces a race condition if pods are slow to start. Probably unlikely though..

But it may be more robust to add a MCP specific label (so we can distinguish from OSS install?) to the Daemonset (so its always there) and check that.

Or we could wait for the pods here?

scripts/asm-installer/install_asm
```diff
@@ -2563,6 +2594,11 @@ scrape_managed_urls() {
 }
 
 install_managed_components() {
+  if [[ "${USE_MCP_CNI}" -eq 1 ]]; then
+    info "Installing CNI into the cluster..."
```
howardjohn, line 2598: Whats our philosophy on overwriting user installed (oss?) components? should we warn/prompt/fail or anything?

howardjohn (commented):
lgtm

scripts/asm-installer/install_asm
zerobfd: So far the customer feedback has been: if they have to add an extra flag to do something (e.g. `--cni-managed`) then they're aware that dependencies are going to be installed and might overwrite things. Soon(tm) when we have more defined interactive and non-interactive modes, we'll have a more consistent stance.

scripts/asm-installer/install_asm
zerobfd: After 10 minutes of Googling I'm not convinced I know what the best thing to do here is, so whatever you think the best way is I'm happy to do. Is this something that needs to go into this PR to be safe, or can we merge as-is to start allowing staging environments to test CNI and add it later?

PR #657, opened 2021-05-06
Update tests, small usage prompt fix

zerobfd:


asmcli/asmcli.sh
```diff
@@ -1119,32 +1131,27 @@ EOF
   done <<EOF
 CLUSTER_LOCATION
 CLUSTER_NAME
-PROJECT_ID
```
backward-compatible, line 1122: Without PROJECT_ID arg, how would the install command work with just cluster name and location?

We wanted 2 ways to install:
1) cluster name, location and project id
2) cluster kubeconfig and optional context (optional because current-context could suffice)

If #1 and #2 args are mixed, we do not proceed.

asmcli/tests/common.sh
```diff
@@ -705,27 +708,31 @@ run_basic_test() {
...
     --output-dir ${OUTPUT_DIR} \
     ${EXTRA_FLAGS}"
   # shellcheck disable=SC2086
   _CI_REVISION_PREFIX="${LT_NAMESPACE}" \
     ../asmcli ${KEY_FILE} ${SERVICE_ACCOUNT} \
-    -l "${LT_CLUSTER_LOCATION}" \
-    -n "${LT_CLUSTER_NAME}" \
-    -p "${PROJECT_ID}" \
     -m "${MODE}" \
+    -p "${PROJECT_ID}" \
```
backward-compatible, line 727: is this negative test? this should fail because of the mixed arguments

asmcli/asmcli.sh
```diff
@@ -1119,32 +1131,27 @@ EOF
...
   if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 1 && "${CLUSTER_DETAIL_VALID}" -eq 0 ]]; then
     MISSING_ARGS=1
-    warn "Missing one or more required options for [CLUSTER_LOCATION|CLUSTER_NAME|PROJECT_ID]"
+    warn "Missing one or more required options for [CLUSTER_LOCATION|CLUSTER_NAME]"
   fi
 
   if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 0 && "${KUBECONFIG_SUPPLIED}" -eq 0 ]]; then
     MISSING_ARGS=1
-    warn "At least one of the following is required: 1) --kubeconfig or 2) --cluster_location, --cluster_name, --project_id"
+    warn "At least one of the following is required: 1) --kubeconfig or 2) --cluster_location, --cluster_name"
```
backward-compatible, line 1148: i think we need project id

asmcli/asmcli.sh
```diff
@@ -1119,32 +1131,27 @@ EOF
...
   # Script will not infer the intent between the 2 use cases in case both values are provided
   if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 1 && "${KUBECONFIG_SUPPLIED}" -eq 1 ]]; then
-    fatal_with_usage "Incompatible arguments. Kubeconfig cannot be used in conjuntion with [--cluster_location|--cluster_name|--project_id]."
+    fatal_with_usage "Incompatible arguments. Kubeconfig cannot be used in conjuntion with [--cluster_location|--cluster_name."
   fi
 
   if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 1 && "${CLUSTER_DETAIL_VALID}" -eq 0 ]]; then
     MISSING_ARGS=1
-    warn "Missing one or more required options for [CLUSTER_LOCATION|CLUSTER_NAME|PROJECT_ID]"
+    warn "Missing one or more required options for [CLUSTER_LOCATION|CLUSTER_NAME]"
```
backward-compatible, line 1143: i think we need project id

asmcli/asmcli.sh
```diff
@@ -1119,32 +1131,27 @@ EOF
...
 EOF
 
-  if [[ -n "${KUBECONFIG}" ]]; then
-    KUBECONFIG_SUPPLIED=1 # kubeconfig usage intended
-  fi
-
   # Script will not infer the intent between the 2 use cases in case both values are provided
   if [[ "${CLUSTER_DETAIL_SUPPLIED}" -eq 1 && "${KUBECONFIG_SUPPLIED}" -eq 1 ]]; then
-    fatal_with_usage "Incompatible arguments. Kubeconfig cannot be used in conjuntion with [--cluster_location|--cluster_name|--project_id]."
+    fatal_with_usage "Incompatible arguments. Kubeconfig cannot be used in conjuntion with [--cluster_location|--cluster_name."
```
backward-compatible, line 1138: i think we need project id

backward-compatible (changes requested):

asmcli/asmcli.sh
zerobfd: `PROJECT_ID` is still required, it just got moved up to another block. I figured it was going to be required even for platforms where the kube context might not have the project information, so we might as well make it required all the time.

backward-compatible: Would the future need be for the environ id (may or may not be cluster's project) rather than the project id (cluster's project)?

zerobfd: Yep, you're right. Fixed.

backward-compatible (changes requested):

ZhengzheYang (approved):

PR #658, opened 2021-05-06
append the yaml and add unit tests

ZhengzheYang:
this PR adds the implementation for the yaml files appending and unit tests.


zerobfd (approved):
lgtm. Before merging, can you add a test for 1) calling `list-yamls` without appending any to the list and asserting empty output, and 2) adding two different files?

PR #659, opened 2021-05-07
update version

jasonwzm:


zerobfd (approved):

howardjohn (commented):
Dumb question - how is this supposed to be used, since 1.9.5 is not published yet? I have just been overwriting POINT to an older one in my testing but not sure if I am missing something

jasonwzm:
@howardjohn Once it is updated, we will merge it to staging-1.9-asm branch. From there you can use `_CI_ASM_PKG_LOCATION`, `_CI_ASM_IMAGE_LOCATION` and `_CI_ASM_KPT_BRANCH` to override the package and image location to use the internally available path.

Issue #661, opened 2021-05-10
ENVIRON_PROJECT_ID is not set correctly on MacOS


YpNo:
Hi,

The "sed" command to retrieve project id from workload identity doesn't works on MacOS.

Line 2117: 
`ENVIRON_PROJECT_ID="$(kubectl get memberships.hub.gke.io membership -o=json | jq .spec.workload_identity_pool | sed 's/^\"\(.*\).\(svc\|hub\).id.goog\"$/\1/g')"`

Output on MacOS: "<my_project_id>.svc.id.goog" instead of "<my_project_id>".

To resolve it, I purpose the following sed which works on MacOS and Linux:
`ENVIRON_PROJECT_ID="$(kubectl get memberships.hub.gke.io membership -o=json | jq .spec.workload_identity_pool | sed -E 's/^\"(.*)\.(svc|hub)\.id\.goog\"$/\1/')"`

Tested with latest install_asm script and version 1.9.3-2 of ASM

Regards.

zerobfd:
I think this issue was fixed by the merged commit? I thought it would auto-close. Please re-open if it's still an issue.

PR #662, opened 2021-05-10
Fix sed command to retrieve project id from workload identity

YpNo:
The previous sed command didn't work on MacOS.

Linked issue: #661


YpNo:
@googlebot I signed it!

zerobfd (approved):
Thanks for the PR! We don't have a good way to test automatically with macOS so we rely on reports to keep a reasonable level of compatibility.

PR #666, opened 2021-05-11
using context append istio/kube yamls 

ZhengzheYang:


asmcli/asmcli.sh
```diff
@@ -2657,19 +2668,21 @@ EOF
...
   if [[ "${MANAGED}" -eq 1 ]]; then
     cat "${MANAGED_MANIFEST}"
     return
   fi
 
   PARAMS="-f ${OPERATOR_MANIFEST}"
-  while read -d ',' -r yaml_file; do
-    PARAMS="${PARAMS} -f ${yaml_file} "
+  while read -r yaml_file; do
+    if [[ -n ${yaml_file} ]]; then
```
zerobfd, line 2679: Why do we need to check for empty variables here? Can we do it when we're appending to the list instead of every time we're reading it?

If that's not possible, please double quote the "${yaml_file}" here.

asmcli/asmcli.sh
```diff
@@ -2657,19 +2668,21 @@ EOF
...
+      PARAMS="${PARAMS} -f ${yaml_file} "
+    fi
   done <<EOF
-${CUSTOM_OVERLAY}
+$(context_list-istio-yamls)
 EOF
+  echo "${PARAMS}"
```
zerobfd, line 2685: Is this a debug statement or here on purpose?

zerobfd (changes requested):

asmcli/asmcli.sh
ZhengzheYang: Sorry. Debug params.

asmcli/asmcli.sh
ZhengzheYang: It will output empty string if the list is empty. I could update the `context_list-istio-yamls` to output nothing if empty?

zerobfd: Ah, I should have caught that in the last review. In the `list-istio-yamls` function use `-j` as the flag to `jq` instead of `-r`. `-j` is the same except it suppresses the trailing newline character.

ZhengzheYang: Sounds good! Learning everyday 😄

zerobfd (approved):
Thanks for all the offline back-and-forth. LGTM as long as tests work.

asmcli/asmcli.sh
ZhengzheYang: Discussed offline. Decision is to use for loop instead of heredoc to avoid the empty line.

PR #670, opened 2021-05-11
Vsphere onprem 

xulingqing:


jasonwzm:
I don't think this will work. `istioctl` takes `IstioOperator` as input, the other Kubernetes resources should be applied thru `kubectl`.

PR #677, opened 2021-05-14
adding option to impersonate service accounts when executing gcloud commands

ddlfcloud:
At the moment you must provide a key file to execute as a service account, but security best practice is to not download long lived key files and use account impersonation.

### DIFF
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -951,7 +951,7 @@ parse_args() {
         PRINT_CONFIG=1
         shift 1
         ;;
-      -s | --impersonate_service_account | --impersonate_service-account)
+      -s | --impersonate_service_account | --impersonate-service-account)
         arg_required "${@}"
         IMPERSONATE_SERVICE_ACCOUNT="${2}"
         shift 2
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -952,7 +952,7 @@ parse_args() {
         PRINT_CONFIG=1
         shift 1
         ;;
-      -s | --impersonate_service_account | --impersonate-service-account)
+      -i | --impersonate_service_account | --impersonate-service-account)
         arg_required "${@}"
         IMPERSONATE_SERVICE_ACCOUNT="${2}"
         shift 2
--- a/scripts/asm-installer/install_asm
+++ b/scripts/asm-installer/install_asm
@@ -585,7 +585,7 @@ can_modify_gcp_apis() {
 }
 
 can_modify_gcp_iam_roles() {
-  if ! can_modify_at_all; then false; return; fi
+  if ! can_modify_at_all || [[ -n "${IMPERSONATE_SERVICE_ACCOUNT}" ]]; then false; return; fi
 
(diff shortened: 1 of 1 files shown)


ddlfcloud:
For whatever reason when testing with master branch the latest istio asm release tar seems broken:
```sh
2021-05-14T05:02:17.484304 install_asm: https://storage.googleapis.com/gke-release/asm/istio-1.9.5-asm.2-linux-amd64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   214  100   214    0     0   1209      0 --:--:-- --:--:-- --:--:--  1209

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```
```sh
# 1.9.5 vs 1.9.3

macbookpro% curl https://storage.googleapis.com/gke-release/asm/istio-1.9.5-asm.2-linux-amd64.tar.gz | tar xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   214  100   214    0     0    829      0 --:--:-- --:--:-- --:--:--   829
tar: Error opening archive: Unrecognized archive format

macbookpro% curl https://storage.googleapis.com/gke-release/asm/istio-1.9.3-asm.2-linux-amd64.tar.gz | tar xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 41.7M  100 41.7M    0     0  13.0M      0  0:00:03  0:00:03 --:--:-- 13.0M
```

zerobfd:
Thanks for the PR! I didn't know that was best practice. I have a few comments/questions:

First, since I haven't figured out how to make test output more accessible, `shellcheck` found issues with the PR. I think even on macbooks this isn't too hard to get running locally, unlike most of the other tests.

```
In install_asm line 955:
      -s | --impersonate_service_account | --impersonate-service-account)
      ^-- SC2221: This pattern always overrides a later one.


In install_asm line 960:
      -s | --service_account | --service-account)
      ^-- SC2222: This pattern never matches because of a previous pattern.
```

Second, about the failures on master branch: generally `install_asm` isn't expected to work publicly except for versions in release branches. We have a lot of compatibility tests that we run internally that expect the latest version, so master/staging branches will reference versions that are only available internally most of the time. For PRs like this it's easiest just to switch the MAJOR/MINOR/POINT etc. variables to something that you can access since the version of ASM itself doesn't matter.

Third: if you *can* impersonate a service account, is there any reason to use a key file instead? If it's possible to avoid adding another argument to the CLI and just attempt to impersonate a service account if a key file isn't specified, then I think it would make it easier for customers to do the right thing.

ddlfcloud (commented):
needs to be -i . copy pasta fail

ddlfcloud:
For point 3: I think that customers may want have the key-file option, I'm not confident it will won't be missed so for now maybe keep both ? We're trying to protect against: 

```
developers checking keys into the source code or leaving them in the Downloads directory of their workstation.
```
https://cloud.google.com/iam/docs/understanding-service-accounts#managing_service_account_keys

ddlfcloud:
This needs more thought. I missed a crucial issue with the kubectl gcloud container cluster creds generation.

PR #678, opened 2021-05-14
[1.10 Only] Update hub workload pool suffix

JimmyCYJ:


JimmyCYJ:
DO NOT SUBMIT

scripts/asm-installer/asm_vm
```diff
@@ -946,7 +946,15 @@ uses_environ_workload_identity() {
 
   if [[ "$(retry 2 kubectl get configmap "${CONFIGMAP}" -n istio-system \
     -ojsonpath='{.data.mesh}' | grep -x -c "^trustDomain: ${PROJECT_ID}.hub.id.goog$" \
```
jasonwzm, line 948: We can achieve the check in one line with an OR in the regex. e.g. https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm#L2201

jasonwzm (changes requested):
@jtrbs Please also review.

scripts/asm-installer/asm_vm
JimmyCYJ: Done. Please take a look. Thanks!

scripts/asm-installer/asm_vm
```diff
@@ -945,8 +945,11 @@ uses_environ_workload_identity() {
   fi
 
   if [[ "$(retry 2 kubectl get configmap "${CONFIGMAP}" -n istio-system \
-    -ojsonpath='{.data.mesh}' | grep -x -c "^trustDomain: ${PROJECT_ID}.hub.id.goog$" \
-    || true)" -eq 0 ]]; then
+    -ojsonpath='{.data.mesh}' | grep -x -c "^trustDomain: ${PROJECT_ID}.\(hub\|svc\).id.goog$" \
+    || true)" -eq 1 ]]; then
```
jasonwzm, line 949: I guess we can do `-eq 0` here and we can just `false` below, so we don't need the else below.

jasonwzm (changes requested):

scripts/asm-installer/asm_vm
jasonwzm: LGTM

scripts/asm-installer/asm_vm
JimmyCYJ: Forgot to change it back. Done.

jasonwzm (approved):

jtrbs (approved):
LGTM

PR #680, opened 2021-05-14
refactor configure

ZhengzheYang:
This PR refactors the configure steps to take place in install subcommand and in each individual components.

### DIFF
--- a/asmcli/commands/install.sh
+++ b/asmcli/commands/install.sh
@@ -1,50 +1,27 @@
 install_subcommand() {
-  # TODO: validate (mesh_ca | citadel | private_ca).
-  # TODO: validate (mcp | in-cluster) control plane.
-
-  # TODO: configure (mesh_ca | citadel | private_ca).
-  # TODO: configure (mcp | in-cluster) control plane.
-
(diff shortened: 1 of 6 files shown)


asmcli/commands/install.sh
```diff
@@ -200,3 +177,49 @@ outro() {
...
+    local CA_CHAIN; CA_CHAIN="$(context_get-option "CA_CHAIN")"
+
+    info "Installing certificates into the cluster..."
+    kubectl create secret generic cacerts -n istio-system \
+      --from-file="${CA_CERT}" \
+      --from-file="${CA_KEY}" \
+      --from-file="${CA_ROOT}" \
+      --from-file="${CA_CHAIN}"
+  fi
+}
```
zerobfd, line 211: Everything looks good except this is Citadel specific, so it should co into a similar file as meshca.sh

zerobfd (approved):

asmcli/commands/install.sh
ZhengzheYang: Moved to an install function in citadel.sh

PR #682, opened 2021-05-17
Handle when a user is using SA impersonation and allow roles/owners to satisfy cluster permissions check

ddlfcloud:
I opened https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/677, earlier but realised that the way kubectl gcloud auth works means this wont work. As the gcloud command inside $HOME.kube/config can't have the 'impersonate-service-account' suffix. I think the best approach is to recognise a user is using SA impersonation and handle it within the script the same as a regular service account.

I don't want to set the impersonation gcloud setting inside the script because the user needs to remember to unset it later on. This is an issue currently when using --service-account option, the script will set the SA as the user but once it's done theres no unsetting. 

Also if you are an owner of the project the cluster permissions check will fail, added a line to accept that. roles/owner as all the container.* permissions.


scripts/asm-installer/install_asm
```diff
@@ -1860,6 +1865,12 @@ local_iam_user() {
     ACCOUNT_TYPE="serviceAccount"
   fi
 
+  if is_sa_impersonation; then
+    ACCOUNT_NAME="$(gcloud config get-value auth/impersonate_service_account)"
+    ACCOUNT_TYPE="serviceAccount"
+    warn "Service account impersonation current configured to impersonate '${ACCOUNT_NAME}'."
```
zerobfd, line 1871: "current" should be "currently"

zerobfd (approved):
Thanks!

scripts/asm-installer/install_asm
ddlfcloud: ack. done

PR #690, opened 2021-05-19
refactor print-config

ZhengzheYang:


zerobfd (approved):
Just as a note: before we finish the split and have subcommands call these files, we should move the if blocks to the main asmcli flow and not the subcommands. (i.e. by the time we call `print-config_subcommand` we shouldn't need to check whether `${PRINT_CONFIG} -eq 1` because otherwise we wouldn't be in this command.)

This PR is a reasonable step in that direction, just wanted to make the direction clear.

PR #692, opened 2021-05-20
fix: fix istiod service account (applicable from asm 1.11) when granting privateca roles

shankgan:


howardjohn (commented):
If this broke here it probably will break others in OSS @monkeyanator?

Also, is install_asm only for a specific version? Since I think this is only correct for 1.11?

Monkeyanator:
@howardjohn good point... opens up the broader question if users should be able to depend on resource names as being stable? I'm guessing this particular change would be more likely to break other vendors than regular OSS users

zerobfd (approved):
FYI: the refactoring going on in the v2 branch is planned to become the new tool by 1.11. At some point we'll go through and reconcile bugfixes between that branch split and launch so don't feel like you need to also open a PR there, but I wanted to make it clear just in case.

PR #693, opened 2021-05-20
add subcommands

ZhengzheYang:


zerobfd (changes requested):
Looks good except for one thing: after the subcommand split there's not really a point in having a --mode install/whatever, so let's just get rid of it.

In the `parse_args` function, we should warn "As of version 1.10 the --mode flag is deprecated and will be ignored." but continue operation, instead of complaining about an unknown option and exiting with an error.

zerobfd (approved):
Chatted offline, will add changes to the next PR.

PR #694, opened 2021-05-20
[1.10 Only] Use new Hub WP and update GKE_CLUSTER_URL

JimmyCYJ:

PR #695, opened 2021-05-20
[1.10 Only] Use new Hub workload pool

JimmyCYJ:
This PR includes https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/694.

Roll forward https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/678 with some fixes to GKE_CLUSTER_URL

PR #697, opened 2021-05-21
Revert "[1.10 Only] Use new Hub WP and update GKE_CLUSTER_URL"

maasen:
Reverts GoogleCloudPlatform/anthos-service-mesh-packages#694

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -121,11 +121,6 @@ openAPI:
         setter:
           name: anthos.servicemesh.hubMembershipID
           value: MEMBERSHIP_ID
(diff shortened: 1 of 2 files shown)


JimmyCYJ:
Can we get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/695 merge first? If that still breaks the tests, I will revert them back.

howardjohn:
Please revert first and debug second. You are breaking all other developers in the meantime.

maasen:
My preference is to rollback by default rather than roll forward

jasonwzm (approved):

PR #703, opened 2021-05-25
Enable cross cluster workload entry again

jasonwzm:
Upgrade from 1.9->1.10 should not hit the permission issue again as istio-reader clusterrole has the workloadentry read access.

Also remove the autoregistration env var as it is enabled by default.


stevenctl (approved):
LGTM, I'd give it a test with the latest. How does the script handle older versions..if I want 1.9 (for whatever reason) do I get an older script?

jasonwzm:
@stevenctl Yep, I did a smoke test using staging 1.10 build. It does not have issues. Yes, you get a 1.9 script in that case.

PR #704, opened 2021-05-26
remove global references

ZhengzheYang:
This PR manually ensures all global variables are read only and all other variables are local and use context interfaces.


zerobfd (approved):
Offline context: we're working on a linter/test to help verify this and enforce it going forward. This is an iterative step that helps future development, and that linter is coming in a PR soon(tm).

PR #707, opened 2021-05-27
add linter for global and local references

ZhengzheYang:


asmcli/asmcli.sh
```diff
@@ -212,7 +212,7 @@ gcloud() {
 }
 
 kubectl() {
-  local KCF
+  local KCF KCC
```
zerobfd, line 215: I didn't know `local` could take multiple arguments! Nice.

scripts/release-asm/merge
```diff
@@ -16,3 +17,48 @@ $(cat "$@")
...
+
+skip_pattern() {
+  local PATTERN
+
+  PATTERN="[[:space:]]\-"
+  PATTERN="${PATTERN}|\-\-"
+  PATTERN="${PATTERN}|shellcheck"
+  PATTERN="${PATTERN}|^#"
+  PATTERN="${PATTERN}|echo"
+  PATTERN="${PATTERN}|.split\("
```
zerobfd, line 59: Can you add comments for some of these justifying why we're skipping them? The `^#` I understand (although I feel like it should probably be `^[[:space:]]*#`), but some of them seem like perfectly valid cases. e.g. we shouldn't skip

`VAR=VAL || echo "Failed to set"`

zerobfd (changes requested):
Can we pull the linting out into a separate executable instead of having it run as part of merge? That would make it easier to incorporate in pipelines. We can always just call the executable from inside of merge if we want that behavior. (It's not clear that we do, though.)

scripts/release-asm/merge
ZhengzheYang: Chatted offline to lint ALL_CAPs only.

asmcli/tests/lint_variables
```diff
@@ -0,0 +1,41 @@
+main() {
```
zerobfd, line 1: Let's make sure the bash header is here `#!/usr/bin/env bash`

asmcli/asmcli.sh
```diff
@@ -51,7 +51,7 @@ MANAGED_WEBHOOKS=""
 EXPOSE_ISTIOD_SERVICE=""
 CANONICAL_CONTROLLER_MANIFEST=""
 
-SCRIPT_NAME="${0##*/}"
+SCRIPT_NAME=""
```
zerobfd, line 54: I forgot if I asked this already or not, but can we just make this

```
SCRIPT_NAME="${0##*/}"; readonly SCRIPT_NAME
```

instead of setting it later?

zerobfd (approved):
Thanks!

asmcli/asmcli.sh
ZhengzheYang: Done

asmcli/tests/lint_variables
ZhengzheYang: Oops. Done.

PR #709, opened 2021-06-02
Revert "Enable Network Policy for MCP with CNI (#668)"

bianpengyuan:
I confirmed with GKE networking team that workload identity should imply PTP CNI enablement, which makes network policy enablement not necessary for Istio CNI. This is an internal tracking bug for more information: [b/189281678](https://b.corp.google.com/issues/189281678).


ruigulala:
Does the GKE people know why existing clusters doesn't have CNI enabled even with WI? I remember you've asked them to debug.

zerobfd (approved):

PR #712, opened 2021-06-03
asm_vm: prepend project id to network id

stevenctl:
To be consistent with https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/asm-installer/install_asm#L1601

### DIFF
--- a/scripts/asm-installer/asm_vm
+++ b/scripts/asm-installer/asm_vm
@@ -1261,7 +1261,7 @@ retrieve_cluster_network() {
   info "Retrieving the cluster's network name..."
 
   # TODO possible to share code with scriptaro? or read from something in-cluster to make this more stable.
-  NETWORK="${PROJECT_ID}-$(retry 2 run gcloud container clusters describe "${CLUSTER_NAME}" \
+  NETWORK="$(retry 2 run gcloud container clusters describe "${CLUSTER_NAME}" \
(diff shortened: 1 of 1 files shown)


scripts/asm-installer/asm_vm
```diff
@@ -1260,7 +1260,8 @@ retrieve_expansiongateway() {
 retrieve_cluster_network() {
   info "Retrieving the cluster's network name..."
 
-  NETWORK="$(retry 2 run gcloud container clusters describe "${CLUSTER_NAME}" \
+  # TODO possible to share code with scriptaro? or read from something in-cluster to make this more stable.
```
jasonwzm, line 1263: I think we should just read from meshconfig in the cluster to get the value.

jasonwzm (changes requested):

scripts/asm-installer/asm_vm
stevenctl: It doesn't exist in meshconfig, it resides in:
* injection template for "cluster wide" network setting 
* gateway env variables

The eastwest gateway env variable would be the most accurate thing to use imo.

jasonwzm: That sg. Since we already check the existence of eastwest gateway, so we can get the value from there.

jasonwzm (approved):

PR #713, opened 2021-06-07
Better error message for showing executed command when all retries failed

kyasbal-1994:
Hi team, 

Current install_asm command is hard to debug when it failed with some error of called command from it because it won't show what commands were executed.
This is just a short modification to enable the retry function to show executed command when it failed all of attemptions.


kyasbal-1994:
Closing once. Commit message was bad.

PR #718, opened 2021-06-08
Update docker file

ZhengzheYang:
Update the docker file to use the /workspace directly without having to copy. Also, add steps to the cloudbuild yaml to perform `npm install` and run unit tests.

### DIFF
--- a/asmcli/cloudbuild-main.yaml
+++ b/asmcli/cloudbuild-main.yaml
@@ -3,19 +3,24 @@ steps:
   ########
   # SETUP
   ########
-  - name: 'gcr.io/cloud-builders/bazel'
-    entrypoint: 'bazel'
-    args: ['build', 'merge']
-  - name: 'bash'
-    args: ['cp', '${_BAZEL_BIN}/${_SCRIPT_NAME}', asmcli/]
   - name: 'gcr.io/cloud-builders/docker'
-    dir: 'asmcli'
     id: 'build-tester-image'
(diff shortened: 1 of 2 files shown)


asmcli/cloudbuild-main.yaml
```diff
@@ -6,17 +6,22 @@ steps:
   - name: 'gcr.io/cloud-builders/bazel'
     entrypoint: 'bazel'
     args: ['build', 'merge']
```
zerobfd, line 8: We should put this after the image build.

asmcli/cloudbuild-main.yaml
```diff
@@ -6,17 +6,22 @@ steps:
...
+  
   - name: 'bash'
     args: ['cp', '${_BAZEL_BIN}/${_SCRIPT_NAME}', asmcli/]
```
zerobfd, line 11: I think you're already aware, but we should remove this once we figure out a way to get the artifact committed. No action on this PR, just mentioning it.

zerobfd (approved):

asmcli/cloudbuild-main.yaml
ZhengzheYang: Will do.

asmcli/cloudbuild-main.yaml
ZhengzheYang: Done

PR #720, opened 2021-06-09
Add Fleet Id as an input

backward-compatible:


ZhengzheYang (approved):
LGTM. Should we add usage info for this?

backward-compatible:
> LGTM. Should we add usage info for this?

Just updated the usage with gcp vs non-gcp behavior. Lmk if I should add more info there.

asmcli/commands/help.sh
```diff
@@ -62,11 +62,14 @@ OPTIONS:
   -l|--cluster_location  <LOCATION>   The GCP location of the target cluster.
   -n|--cluster_name      <NAME>       The name of the target cluster.
   -p|--project_id        <ID>         The GCP project ID.
-  --kc|--kubeconfig      <KUBECONFIG> Path to the kubeconfig file to use for CLI requests.
+  --kc|--kubeconfig <KUBECONFIG_FILE> Path to the kubeconfig file to use for CLI requests.
                                       Required if not supplying --cluster_location,
                                       --cluster_name, --project_id in order to locate
                                       and connect to the intended cluster.
   --ctx|--context        <CONTEXT>    The name of the kubeconfig context to use.
+  --fleet_id             <FLEET ID>   The ASM Fleet host project ID. Required for non-GCP
```
zerobfd, line 70: Let's avoid the phrasing "ASM Fleet". In this instance I think we can just leave out the word "ASM" entirely and say "The Fleet host project ID."

If you know what at Fleet is it makes sense, if you don't then adding ASM isn't going to help.

asmcli/asmcli.sh
```diff
@@ -1971,16 +1987,16 @@ is_membership_crd_installed() {
 }
 
 populate_environ_info() {
```
zerobfd, line 1989: Since you were super detailed and renamed all of the other occurrences, can we rename this function as well?

zerobfd (approved):
We should change the usage wording to remove "ASM", otherwise it looks great. Thanks for being so thorough.

asmcli/commands/help.sh
backward-compatible: done

PR #723, opened 2021-06-09
add precommit script for asmcli

ZhengzheYang:


zerobfd (approved):
let's change the filename from "cp-asmcli" to "precommit" to make it more general, since we might add more. (and right now it does more than `cp` anyways.)

ZhengzheYang:
@backward-compatible FYI, we now are keeping `asmcli/asmcli` committed and up to date.

PR #725, opened 2021-06-10
fix: option file for migrated meshca trustDomainAlias

shankgan:


shankgan:
Needs to be cherry-picked into staging-1.10 and release.10 branches

jtrbs (approved):
Hi Shankar, may you share some sample kpt commands for trust domain alias setting used by the ca migration so that we can track the reason for removing the pattern? Thanks!

shankgan:
trustdomainAliases must include user-defined trustDomains, for instance "cluster.local", which is the default option

zerobfd (approved):

PR #729, opened 2021-06-10
add trustdomain field for hub

williamaronli:
when install for asm 1.10 we need to add hub.id.goog trust domain alias to enable new workload to old workload which uses the hub.id.goog workload pool


williamaronli:
closed this because a PR already listed here: https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/730

PR #732, opened 2021-06-11
set trust domain aliases to include both new Hub WIP and old Hub WIP to achieve no downtime upgrade for VM

williamaronli:


williamaronli:
@myidpt

myidpt (approved):

PR #735, opened 2021-06-12
delete workloadgroup before delete namespace to prevent flaky delete issue and create new cluster to runtests

williamaronli:


scripts/asm-installer/tests/run_basic_instance_template_creation
```diff
@@ -86,6 +86,7 @@ EOF
   verify_template_instance_with_custom_image
 
   cleanup_workload_service_account
+  kubectl delete workloadgroup "${WORKLOAD_NAME}" -n "${LT_NAMESPACE}"
```
howardjohn, line 89: We must not make it required to do this for users before they delete their namespace. And I think we should test what our users can/should do. So this doesn't seem like a good idea to me. If we must do it to get tests passing we should open a P0 bug to revert it once we fix the underlying issue.

williamaronli: It is only a flaky issue for the tests. I rerun the tests without these lines. It doesn't trigger the workloadgroup not deleted issues

williamaronli: It is only a flaky issue for the tests. I rerun the tests without these lines. It doesn't trigger the workloadgroup not deleted issues

maasen: We should definitely resolve the underlying issue. However submitting this to unblock tests sounds reasonable.

maasen (approved):

scripts/asm-installer/cloudbuild-master.yaml
```diff
@@ -187,7 +187,7 @@ steps:
       --PROJECT_ID "${PROJECT_ID}"
       --BUILD_ID "${BUILD_ID}"
     env:
-    - '_LT_CLUSTER_NAME=long-term-test-cluster-environ'
+    - '_LT_CLUSTER_NAME=long-term-test-cluster-environ-newtwo'
```
JimmyCYJ, line 190: Shall we give a better name, e.g. long-term-test-cluster-environ-vm

williamaronli: ok

JimmyCYJ (approved):

williamaronli:
@zerobfd

zerobfd (approved):

PR #738, opened 2021-06-14
set trust domain aliases to include both new Hub WIP and old Hub WIP to achieve no downtime upgrade for VM

williamaronli:


ZhengzheYang (approved):
Approving this one because #732 is approved but not passing CLA.

PR #739, opened 2021-06-14
apply CPR's crd/cr

ZhengzheYang:


scripts/asm-installer/install_asm
```diff
@@ -2751,6 +2755,18 @@ install_managed_components() {
   if [[ "$DISABLE_CANONICAL_SERVICE" -eq 0 ]]; then
     install_canonical_controller
   fi
+
+  info "Configuring ASM managed control plane revision CRD..."
+  kubectl apply -f "${CRD_CONTROL_PLANE_REVISION}"
```
zerobfd, line 2760: Since the CRDs have to be finished processing before the CRs can be applied, I think it makes sense to do this in the "start_managed_control_plane" function where the other base CRDs are applied. (`run kubectl apply --overwrite=true -f "${BASE_REL_PATH}"`)

scripts/asm-installer/install_asm
```diff
@@ -2751,6 +2755,18 @@ install_managed_components() {
...
+
+  info "Configuring ASM managed control plane revision CR for regular channel..."
+  kubectl apply -f "${CR_CONTROL_PLANE_REVISION_REGULAR}"
```
zerobfd, line 2763: Please prefix all of the `kubectl` statements with `run` since they modify the cluster

zerobfd (approved):
Nit: all of the yaml files are already under the "control-plane-revision" folder so I don't think they also need "control-plane-revision" in the file name. I think just "crd.yaml" and "cr_stable.yaml" etc. would be OK.

Otherwise LGTM, thanks!

asm/control-plane-revision/crd_control_plane_revision.yaml
```diff
@@ -0,0 +1,103 @@
...
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: ControlPlaneRevisionSpec defines the desired state of ControlPlaneRevision
+          properties:
+            channel:
+              description: ReleaseChannel determines the aggressiveness of upgrades.
+              enum:
+                - regular
```
maasen, line 42: Is this a direct copy? IIUC we have only allowed for regular right now and will be adding rapid/stable soon

asm/control-plane-revision/cr_control_plane_revision_rapid.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: mesh.cloud.google.com/v1alpha1
+kind: ControlPlaneRevision
+metadata:
+  name: asm-managed-rapid
```
davidhauck, line 5: Are we sure we want to include these already?

zerobfd (changes requested):
We need to add something to the CRs

scripts/asm-installer/install_asm
ZhengzheYang: Done

scripts/asm-installer/install_asm
ZhengzheYang: Done

ZhengzheYang:
Synced offline: added the status field to the CRs to reconciled state to avoid breakage.

zerobfd (approved):

ZhengzheYang:
The decision made offline is to add Transition Timestamp in later PRs. The command to add the timestamp is potentially `date --utc +%FT%TZ`. We will need to figure out a way to add the time to the file without touching kpt setters.

PR #740, opened 2021-06-15
make --hub_meshca behavior consistent between ASM 1.9 and ASM 1.10 for GKE-on-GCP

jtrbs:
For ASM 1.10

- GKE-on-GCP: The Mesh CA with Environ installation option will be kept unchanged, which installs the cluster with the old HUB WIP.  
- VM: new Hub WIP and its IDP will be used. @JimmyCYJ let me know if you want to keep it the same as GKE-on-GCP to use the old Hub WIP. 
- Onprem: new Hub WIP and its IDP will be used.

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -62,12 +62,6 @@ openAPI:
         setter:
           name: gcloud.project.environProjectNumber
           value: PROJECT_NUMBER
-    io.k8s.cli.setters.gcloud.project.environProjectID:
-      type: string
-      x-k8s-cli:
-        setter:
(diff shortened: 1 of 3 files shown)


jtrbs:
@myidpt, just a note, for GKE-on-GCP, without --hub-meshca option, auto Hub registration is not enabled. Hub WIP usage will then depends on whether user has registered the cluster before they run Scriptaro. If they has registered the cluster, the cluster will use Hub WIP.

myidpt (approved):

JimmyCYJ (approved):

PR #742, opened 2021-06-15
stop generating gateway config

ZhengzheYang:


zerobfd (changes requested):
I think if we're doing this, then we need to also change the `istio/options/managed-control-plane.yaml` to not include any gateways, and apply the meshconfig values in that file if necessary.

ZhengzheYang:
logic incorporated along #753

PR #744, opened 2021-06-15
Migration script

shankgan:


shankgan:
Preliminary. Do not merge

scripts/migration/migrate-14-managed
```diff
@@ -0,0 +1,550 @@
...
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+  fi
+  echo -e "${green}OK${clr}"
+}
+
+rollback_mesh_ca() {
```
shankgan, line 243: doesn't work...will be rolled back in the next commit

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
+#!/bin/bash
```
zerobfd, line 1: nit: executables shouldn't have .sh extensions

zerobfd, line 1: Use `#!/usr/bin/env bash` instead for compatibility

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+#
+# The next step is to migrate sidecars to ASM. The script will automatically reconfigure the sidecar injection
+# to inject new pods (in namespaces selected by the `istio-injection=enabled` label). Once this is complete,
+# existing pods will still use the old version; these workloads must be manually re-deployed to be updated to use ASM.
+# This can be done using `kubectl rollout restart deployment` or similar tools.
+#
+# After all workloads have been migrated to the the newer ASM control plane, the Istio on GKE addon
+# can be cleaned up using the cleanup() routine
+
+TMP_DIR=$(mktemp -d)
```
zerobfd, line 33: Please surround $() with double quotes

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+    echo "Selected cluster: '$(kubectl config current-context 2> /dev/null)'"
+    echo
+    prompt "Starting migration..."
+  fi
+}
+
+check_prerequisites() {
+  echo -e "${blue}Checking prerequisites...${clr}"
+  command -v kubectl >| /dev/null 2>&1 || die "kubectl must be installed, aborting."
+  command -v istioctl >| /dev/null 2>&1 || die "istioctl must be installed, aborting."
```
zerobfd, line 181: I don't think I've ever once installed `istioctl`. Can we just have this tool download and extract it if necessary?

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+  echo -e "${green}OK${clr}"
+}
+
+configure_mesh_ca() {
+  prompt "Configuring Istio Addon to trust Anthos Service Mesh..."
+  pushd "${TMP_DIR}" > /dev/null
+  HUB_WIP="$(kube get cm -n istio-system istio-asm-managed -oyaml | grep 'trustDomain: .*' | cut -d':' -f2 | xargs)"
+  if [[ "${HUB_WIP}" == "" ]]; then
+    popd > /dev/null
```
zerobfd, line 190: nit: popd is unnecessary if you're going to `die` right afterwards

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+  # Wait for Citadel to distribute the cert everywhere. Note: this always takes at least one minute
+  # due to a bug/feature in Citadel causing it to delay provisioning at startup.
+  heading "Waiting for root certificate to distribute to all pods. This will take a few minutes..."
+  for namespace in $(kubectl get namespaces -oname | cut -d/ -f2); do
+    # Check against a portion of the root to see if its been distributed.
+    # Note: this doesn't ensure the workloads have actually hot-restarted to pickup the change.
+    ROOT_CERT="$(kubectl get secret -n "${namespace}" istio.default -ojsonpath='{.data.root-cert\.pem}')"
+    LAST_ERR=$?
+    if [[ "${LAST_ERR}" != 0 ]]; then
+      # Likely namespace was deleted while in progress
```
zerobfd, line 228: Can we check here and do some better error reporting if the namespace still exists?

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+      "volumeMounts": [{"mountPath": "/var/run/root-certs", "name": "meshca-root", "readOnly": true}]
+    }],
+    "volumes": [{"name": "meshca-root", "secret":{"secretName": "meshca-root"}}]
+  }}}}'
+  echo -e "${green}OK${clr}"
+
+  # Wait for Citadel to distribute the cert everywhere. Note: this always takes at least one minute
+  # due to a bug/feature in Citadel causing it to delay provisioning at startup.
+  heading "Waiting for root certificate to distribute to all pods. This will take a few minutes..."
+  for namespace in $(kubectl get namespaces -oname | cut -d/ -f2); do
```
zerobfd, line 222: Can we just get the namespaces with the injection label?

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+  fi
+  echo -e "${green}OK${clr}"
+}
+
+disable_galley_webhook() {
+  prompt "Replacing the Istio validation webhook with an Anthos Service Mesh validation webhook..."
+  kube patch clusterrole -n istio-system istio-galley-istio-system --type='json' -p='[{"op": "replace", "path": "/rules/2/verbs/0", "value": "get"}]'
+  kube delete ValidatingWebhookConfiguration istio-galley --ignore-not-found
```
zerobfd, line 259: Is this possible to roll back?

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+    echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+  fi
+  echo -e "${green}OK${clr}"
+}
+
+disable_galley_webhook() {
```
zerobfd, line 256: `replace_galley_webhook` seems like a better name

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+      actualSHA=$(echo "${hd}" | tr -d '\n' | sha256sum | awk '{print $1}')
+      builtinSHA=${rsHashMap[${type}]}
+      if [[ $builtinSHA != *"${actualSHA}"* ]];then
+        echo "custom mixer resource found, resource type: ${type}, name: ${rn}"
+        CUSTOM_MIXER="true"
+      fi
+    done
+  done
+  if [[ "${CUSTOM_MIXER}" == "true" ]];then
+    return 1;
```
zerobfd, line 457: unnecessary semicolon

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+  FILTERS=$(kubectl get envoyfilter -A -l 'operator.istio.io/component!=Pilot')
+  if [[ "${FILTERS}" != "" ]]; then
+    prompt_unsupported "Custom Envoy Filters are not supported by ASM. Please remove these if possible" "${FILTERS}"
+  fi
+
+  # Plugin Cert
+  local PLUGINCERT
+  PLUGINCERT=$(kubectl get secret cacerts -n istio-system --ignore-not-found)
+  if [[ "${PLUGINCERT}" != "" ]]; then
+    prompt_unsupported "Detected custom plugin cert" "The plugin certs will not be migrated to ASM.."
```
zerobfd, line 479: Get rid of extra period at the end. Also, the error messages aren't consistent with punctuation or capitalization.

scripts/migration/migrate-addon.sh
```diff
@@ -0,0 +1,653 @@
...
+  fi
+  replace_gateway
+  write_marker "COMPLETE"
+  # replace_webook and cleanup are invoked seperately and needed for completion
+  print_restart_instructions
+}
+
+usage() {
+    echo "Usage:"
+    echo "  $0 [OPTIONS] [--command] [command-name]"
```
zerobfd, line 587: Let's make it more obvious that if you leave out `--command` it runs all of them.

zerobfd (changes requested):
Overall looks OK. My biggest request is to add TODO comments to places where we know there's work, so it's more apparent to reviewers.

Also, unless you source a shell file (which is mostly a bad idea unless you know what you're doing) then it can't affect your environment, so all of the pushd/popd is unnecessary.

scripts/migration/migrate-addon.sh
shankgan: The tool isn't meant to run in isolation. Since we also expect the Managed control-plane to be installed prior to the migration tool, that step should end up installing istioctl. The docs will reflect this.

scripts/migration/migrate-addon.sh
shankgan: This actually is just disabling the galley webook. Will change the header message.

scripts/migration/migrate-addon.sh
zerobfd: We can rely on the docs to guide users to download `istioctl`, but "installing" it (e.g. putting it in $PATH) isn't something that's normally done (to my knowledge). We should support a way to either specify the `istioctl` binary location or put it in a consistent location ourselves.

shankgan: To be more specific: The first step in the docs is to use the scriptaro installer to download the appropriate version of istioctl. The location of the binary is one of the arguments to the installer script. We could add the same argument here to this script and add logic to find istioctl there - does that work?

scripts/migration/migrate-addon.sh
shankgan: The tool now requires an explicit command for it to "run everything". This command is hidden from usages as it is not meant to be used in interactive mode

scripts/migration/migrate-addon.sh
shankgan: we now store the configuration before deleting it..the rollback can rely on that if needed

scripts/migration/migrate-addon.sh
zerobfd: Sure. Or we could make it `./istioctl` and have users `cd` into the output directory that `install_asm` makes, since it puts a copy of `istioctl` there.

zerobfd (approved):
lgtm after you decide what to do with the `istioctl` thing. If you haven't made a bug/issue already then make sure you have some way to remember to follow up on the TODOs and testing.

PR #746, opened 2021-06-16
Adding option revisioned-istio-egressgateway

zioproto:
Makes possible to have a revisioned egressgateway the same way there is already a revisioned ingressgateway.

Tested with:
```
./install_asm --version
1.9.5-asm.2+config1
```

Introduces the new option ` --option revisioned-istio-egressgateway`

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -292,6 +292,14 @@ openAPI:
           values:
           - marker: ASM_REV
             ref: '#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev'
+    io.k8s.cli.substitutions.istio-egressgateway-deploy-name:
+      x-k8s-cli:
+        substitution:
+          name: istio-egressgateway-deploy-name
+          pattern: istio-egressgateway-ASM_REV
+          values:
+          - marker: ASM_REV
+            ref: '#/definitions/io.k8s.cli.setters.anthos.servicemesh.rev'
     io.k8s.cli.setters.anthos.servicemesh.custom-ingressgateway-name:
       type: string
       x-k8s-cli:
--- a/asm/istio/options/revisioned-istio-egressgateway.yaml
+++ b/asm/istio/options/revisioned-istio-egressgateway.yaml
@@ -1,0 +1,38 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


howardjohn:
I don't know the context behind this but I strongly recommend following https://istio.io/latest/docs/setup/additional-setup/gateway/ which this seems a bit different from?

jasonwzm:
@zioproto Can you rebase your branch?

zioproto:
@howardjohn this PR for the revisioned **egress** gateway it is exactly the same approach that we already have for the revisioned **ingress** gateway [here](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/asm/istio/options/revisioned-istio-ingressgateway.yaml)

Please I would like @elfinhe feedback since he originally proposed https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/388

Thanks

howardjohn:
My comments apply to that PR as well, except that was before we had this new functionality in Istio.

zerobfd:
@elfinhe came up with this method in order to have a way to safely test gateway changes until we had injected gateways. We have injected gateways now, so I don't see a strong reason to continue this method going forward.

Is there a particular reason that we're looking to add this now? If there's a particular customer that wants it to unblock something in 1.8 or 1.9 then that would make sense to put it in just those branches, but I don't want to make this a standard documented thing.

zioproto:
I am afraid this is already a standard documented thing.

Please be aware that all files published under [options folder](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/tree/master/asm/istio/options) have big customer visibility because the [documentation](https://cloud.google.com/service-mesh/docs/enable-optional-features#using_the_install_asm_script) links to that folder.

This means that we have customers using the  `--option revisioned-istio-ingressgateway` in production and there was an ask to have the same option for the  `revisioned-istio-egressgateway`.

I am fine closing this PR without merged commits if there is a better long term plan. We should also drop other option files that are in the folder if those were not meant for production usage.

Thank you for your help.

elfinhe:
> @elfinhe came up with this method in order to have a way to safely test gateway changes until we had injected gateways. We have injected gateways now, so I don't see a strong reason to continue this method going forward.
> 
> Is there a particular reason that we're looking to add this now? If there's a particular customer that wants it to unblock something in 1.8 or 1.9 then that would make sense to put it in just those branches, but I don't want to make this a standard documented thing.

+1 for what Clay recommended. Previously the revision based ingressgateway is a temporary solution to safe upgrade before auto-injection for gateway is designed and implemented. We need to align with OSS istio's recommendation, and give customers guidance to use auto-injection for gateways.

```
Using auto-injection for gateway deployments is recommended
```
-- https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway

PR #749, opened 2021-06-17
add-to-mesh subcommand

ZhengzheYang:


asmcli/commands/add-to-mesh.sh
```diff
@@ -0,0 +1,87 @@
...
+  while read -r PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI; do
+    context_set-option "PROJECT_ID" "${PROJECT_ID}"
+    context_set-option "CLUSTER_LOCATION" "${CLUSTER_LOCATION}"
+    context_set-option "CLUSTER_NAME" "${CLUSTER_NAME}"
+    
+    MEMBERSHIP_NAME="$(generate_membership_name)"
+    info "Registering the cluster ${PROJECT_ID}/${CLUSTER_LOCATION}/${CLUSTER_NAME} as ${MEMBERSHIP_NAME}..."
+
+    local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
+    retry 2 gcloud beta container hub memberships register "${MEMBERSHIP_NAME}" \
```
zerobfd, line 80: Use `run` here too. (or whatever the name is now in v2)

asmcli/lib/context.sh
```diff
@@ -101,3 +103,27 @@ context_append-kube-yaml() {
 context_list-kube-yamls() {
   jq -S -r '.kubectlFiles[]' "${context_FILE_LOCATION}"
 }
+
+context_append-clusters-info() {
+  local CLUSTER; CLUSTER="${1}"
+  local TEMP_FILE; TEMP_FILE="$(mktemp)"
+
+  jq -S --arg CLUSTER "${CLUSTER}" '.clustersInfo += [$CLUSTER]' "${context_FILE_LOCATION}" >| "${TEMP_FILE}" \
```
zerobfd, line 111: It looks like you can pull some of this logic into a separate function and share it.

asmcli/commands/add-to-mesh.sh
```diff
@@ -0,0 +1,87 @@
...
+}
+
+add_all_to_mesh() {
+  local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
+  local PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI
+  local MEMBERSHIP_NAME; MEMBERSHIP_NAME=""
+  while read -r PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI; do
+    context_set-option "PROJECT_ID" "${PROJECT_ID}"
+    context_set-option "CLUSTER_LOCATION" "${CLUSTER_LOCATION}"
+    context_set-option "CLUSTER_NAME" "${CLUSTER_NAME}"
```
zerobfd, line 74: Are these just set for the `generate_membership_name` function?

ZhengzheYang: Yes. Do you want me to refactor the `generate_membership_name` to take in some parameters instead of reading from context file?

asmcli/lib/context.sh
ZhengzheYang: Done

asmcli/commands/add-to-mesh.sh
ZhengzheYang: synced offline: skip the refactoring in this PR but we should create a bug to refactor these functions alike in the future for better and easier testing.

asmcli/commands/add-to-mesh.sh
ZhengzheYang: Done

asmcli/commands/add-to-mesh.sh
```diff
@@ -0,0 +1,95 @@
...
+
+add_all_to_mesh() {
+  local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
+  local PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI
+  local MEMBERSHIP_NAME
+  while read -r PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI; do
+    if [[ -n "${PROJECT_ID}" && -n "${CLUSTER_LOCATION}" && -n "${CLUSTER_NAME}" && -n "${GKE_CLUSTER_URI}" ]]; then
+      context_set-option "PROJECT_ID" "${PROJECT_ID}"
+      context_set-option "CLUSTER_LOCATION" "${CLUSTER_LOCATION}"
+      context_set-option "CLUSTER_NAME" "${CLUSTER_NAME}"
```
zerobfd, line 81: Why do we need to set these here?

asmcli/commands/add-to-mesh.sh
```diff
@@ -0,0 +1,95 @@
...
+EOF
+  fi
+}
+
+add_all_to_mesh() {
+  local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
+  local PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI
+  local MEMBERSHIP_NAME
+  while read -r PROJECT_ID CLUSTER_LOCATION CLUSTER_NAME GKE_CLUSTER_URI; do
+    if [[ -n "${PROJECT_ID}" && -n "${CLUSTER_LOCATION}" && -n "${CLUSTER_NAME}" && -n "${GKE_CLUSTER_URI}" ]]; then
```
zerobfd, line 78: I'd like an `else` block to this statement to print an error just in case it ever somehow happens.

asmcli/tests/run_add_to_mesh
```diff
@@ -0,0 +1,151 @@
...
+
+  local ARGS; ARGS="add-to-mesh ${PROJECT_ID}"
+  while read -r CLUSTER_NAME; do
+    ARGS="${ARGS} ${PROJECT_ID}/${CLUSTER_LOCATION}/${CLUSTER_NAME}"
+  done <<EOF
+$(get_test_clusters)
+EOF
+
+  # Test starts here
+  echo "Adding clusters to fleet..."
```
zerobfd, line 42: If we're going to use long-term test clusters for this, we should validate that they're actually unregistered before we start the test, or we might get a false positive.

asmcli/cloudbuild-main.yaml
```diff
@@ -174,6 +174,24 @@ steps:
       - '_CI_ASM_PKG_LOCATION=${_PKG_LOCATION}'
       - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
     timeout: 600s
+  
+  - name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
+    dir: 'asmcli'
+    id: 'run-add-to-mesh'
```
zerobfd, line 180: I'm not sure if this is something we actually want to run on every pre-submit. Let's add it to the staging version now that it works.

zerobfd (approved):
LGTM after the changes are addressed. The refactor looks much better!

asmcli/commands/add-to-mesh.sh
ZhengzheYang: For the `generate_membership_name` function. http://b/191506634 for tracking the expected refactor.

asmcli/commands/add-to-mesh.sh
ZhengzheYang: Done

asmcli/tests/run_add_to_mesh
ZhengzheYang: I tried to unregister the clusters in setup and cleanup, will that suffice? If I check if it's unregistered before the test, and, say, it's registered, should we tried to unregister it (again), or fail the test (but it's our fault since we didn't properly set up the testing cluster)?

zerobfd: Let's just fail. If we tried twice to unregister and it's still registered here we're obviously in a weird unexpected state, so trying again might just mask bigger issues.

asmcli/cloudbuild-main.yaml
ZhengzheYang: Done

PR #750, opened 2021-06-21
 Allow staging image testing before release

jasonwzm:


asm/istio/expansion/vm-eastwest-gateway.yaml
```diff
@@ -45,3 +47,8 @@ spec:
               - name: tls-webhook
                 port: 15017
                 targetPort: 15017
+  values:
+    gateways:
+      istio-ingressgateway:
+        # Enable gateway injection
+        injectionTemplate: gateway
```
zerobfd, line 54: This doesn't look related to staging images, was this included on purpose?

jasonwzm: Yes. This is added to use the injection based gateway so that 1.10 eastwest gateway can be created.

zerobfd (approved):

maasen (approved):
LGTM

PR #753, opened 2021-06-22
add crd/cr files for cpr and apply them (#739)

zerobfd:
I had to do a lot of manual conflict resolution so review it like new code please :)

### DIFF
--- a/asmcli/commands/install.sh
+++ b/asmcli/commands/install.sh
@@ -35,14 +35,19 @@ install() {
 }
 
 install_managed_components() {
+  info "Configuring ASM managed control plane revision CRD..."
+  context_append-kube-yaml "${CRD_CONTROL_PLANE_REVISION}"
+
   info "Configuring base installation for managed control plane..."
   context_append-kube-yaml "${BASE_REL_PATH}"
 
   info "Configuring ASM managed control plane validating webhook config..."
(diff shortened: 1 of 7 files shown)


asmcli/commands/install.sh
```diff
@@ -35,14 +35,19 @@ install() {
...
+  info "Configuring ASM managed control plane revision CRD..."
+  context_append-kube-yaml "${CRD_CONTROL_PLANE_REVISION}"
+
   info "Configuring base installation for managed control plane..."
   context_append-kube-yaml "${BASE_REL_PATH}"
 
   info "Configuring ASM managed control plane validating webhook config..."
   context_append-kube-yaml "${MANAGED_WEBHOOKS}"
 
-  info "Configuring ASM managed control plane components..."
```
ZhengzheYang, line 44: so are we going to do anything in the gateway yaml file? Are we clear about the meshconfig settings in that file?

ZhengzheYang (approved):
LGTM and we might want to merge this first so my other PR could update the use of  context_append-kube-yaml in this PR.

asmcli/commands/install.sh
zerobfd: For 1.10 I think we're safe, so the v2 branch we can remove it. (If not it'll come up in the bug bash.)

PR #755, opened 2021-06-23
Revert "Require Fleet registration with MCP"

howardjohn:
Reverts GoogleCloudPlatform/anthos-service-mesh-packages#754

This is breaking tests


zerobfd:
I reverted this manually on the staging branch, for now the decision is to keep it in master.

PR #757, opened 2021-06-24
Update citadel ca overlay to clear proxyMetadata fields

backward-compatible:


asm/istio/options/citadel-ca.yaml
```diff
@@ -25,13 +25,17 @@ spec:
           - name: TOKEN_AUDIENCES
             value: "istio-ca,PROJECT_ID.svc.id.goog" # {"$ref":"#/definitions/io.k8s.cli.substitutions.token-audiences"}
           - name: ENABLE_STACKDRIVER_MONITORING
-            value: "true" # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.controlplane.monitoring.enabled"}
+            value: "false" # {"$ref":"#/definitions/io.k8s.cli.setters.anthos.servicemesh.controlplane.monitoring.enabled"}
```
zerobfd, line 28: Does it still work if we just remove this completely instead of explicitly setting it to false?

backward-compatible: removed

zerobfd (approved):

PR #760, opened 2021-06-24
install remote secrets on the clusters

ZhengzheYang:


asmcli/commands/add-to-mesh.sh
```diff
@@ -89,7 +96,7 @@ add_all_to_mesh() {
         --gke-uri="${GKE_CLUSTER_URI}" \
         --enable-workload-identity
     else
-      { read -r -d '' MSG; fatal "${MSG}"; } <<EOF || true
+      { read -r -d '' MSG; warn "${MSG}"; } <<EOF || true
```
ZhengzheYang, line 99: this is tentative because the heredoc thing with while loop is coming back. If we are skipping all the clusters (because clusters are already registered to the specified fleet), it will enter the loop nonetheless. Last time we solve this by using for loop, but I am not sure how to read into multiple variables with for loop.

asmcli/commands/install.sh
```diff
@@ -250,3 +250,16 @@ install_control_plane() {
 
   apply_kube_yamls
 }
+
+install_remote_secret() {
```
zerobfd, line 254: I think this belongs in the `add_to_mesh` file, I don't think it will ever get called in `asmcli install blablabla`

asmcli/commands/add-to-mesh.sh
```diff
@@ -72,15 +75,19 @@ EOF
 
 add_all_to_mesh() {
```
zerobfd, line 76: Is there a way to refactor this into different functions? It's pretty hard to follow.

zerobfd (commented):
Core logic looks right, but we need to break up the complexity

asmcli/commands/install.sh
ZhengzheYang: I put it here because 1) it has install in its name and 2) we might need it for apply. Should I put this in `util.sh` or just in`add-to-mesh.sh`?

zerobfd: For now, in `add-to-mesh`. It might move to `util` later but I'm not 100% sure we'll need it in apply, and I want discrete commits to be consistent.

ZhengzheYang: Done

asmcli/commands/add-to-mesh.sh
ZhengzheYang: Refactored.

asmcli/commands/add-to-mesh.sh
ZhengzheYang: Used while loop with `< <(do_stuff)` structure to avoid blank lines.

zerobfd (commented):
Much nicer looking! Can we add one or two unit tests and mock out kubectl/istioctl? (The answer might be no, that's OK.)

ZhengzheYang:
> Much nicer looking! Can we add one or two unit tests and mock out kubectl/istioctl? (The answer might be no, that's OK.)

@zerobfd added one unit test

zerobfd (approved):

PR #761, opened 2021-06-25
Update citadel and off-gcp overlays for multicloud support

backward-compatible:


asm/istio/options/citadel-ca.yaml
```diff
@@ -28,8 +28,8 @@ spec:
     trustDomain: "cluster.local"
     defaultConfig:
       proxyMetadata:
-        GKE_CLUSTER_URL: "" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gke-cluster-url"}
-        GCP_METADATA: "" # {"$ref":"#/definitions/io.k8s.cli.substitutions.gke-metadata"}
+        GKE_CLUSTER_URL: ""
+        GCP_METADATA: ""
```
zerobfd, line 32: I bet the right thing to do is actually remove these two completely from citadel-ca and move them to off-gcp without the refs.

zerobfd (approved):

asm/istio/options/citadel-ca.yaml
backward-compatible: ok. would you say the same about **GCE_METADATA_HOST** field too (expand citadel-ca.yaml)?

zerobfd: Yep, good catch

backward-compatible: done

PR #763, opened 2021-06-25
1.4 migration : Use MCP rapid channel

shankgan:


howardjohn (approved):
Might be good to make it an a variable (maybe just env var) as in the future we will change it. No big deal though, it will be rapid for a while

howardjohn (approved):

PR #766, opened 2021-06-28
Add multicloud support

backward-compatible:


asmcli/asmcli.sh
```diff
@@ -167,7 +171,8 @@ init() {
   readonly RELEASE; readonly RELEASE_LINE; readonly PREVIOUS_RELEASE_LINE; readonly KPT_BRANCH;
 
   KPT_URL="https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages"
-  KPT_URL="${KPT_URL}.git/asm@${KPT_BRANCH}"; readonly KPT_URL;
+  # KPT_URL="${KPT_URL}.git/asm@${KPT_BRANCH}"; readonly KPT_URL;
+  KPT_URL="${KPT_URL}.git/asm@v2"; readonly KPT_URL;
```
zerobfd, line 175: I updated the trigger to pass in the right variable here, you shouldn't need to hardcode it.

asmcli/asmcli.sh
```diff
@@ -327,9 +336,12 @@ configure_kubectl(){
   local ADDR
   ADDR="$(kubectl config view --minify=true -ojson | \
     jq .clusters[0].cluster.server -r)"
+  local DEST
+  DEST="${ADDR:8:${#ADDR}}"
+  DEST="${DEST%:*}"
 
   local RETVAL; RETVAL=0;
-  run_command nc -zvw 10 "${ADDR:8:${#ADDR}}" 443 || RETVAL=$?
+  run_command nc -zvw 10 "${DEST}" 443 || RETVAL=$?
```
zerobfd, line 344: nit: I don't think ADDR is used anymore so you can just keep using that variable instead of adding a new one

asmcli/asmcli.sh
```diff
@@ -1935,22 +1960,34 @@ generate_membership_name() {
...
+  else
+    local KCF; KCF="$(context_get-option "KUBECONFIG")"
+    local KCC; KCC="$(context_get-option "CONTEXT")"
+
+    retry 2 gcloud beta container hub memberships register "${MEMBERSHIP_NAME}" \
+      --project="${FLEET_ID}" \
+      --kubeconfig="${KCF}" \
+      --context="${KCC}" \
+      --enable-workload-identity
+  fi
```
zerobfd, line 1990: In other places we've used a command-builder pattern for this, maybe something like:

```
local CMD
CMD="gcloud beta container hub memberships register ${MEMBERSHIP_NAME}"
CMD="${CMD} --project=${FLEET_ID}"
CMD="${CMD} --enable-workload-identity"
if is_gcp; then
  CMD="${CMD} --gke-uri=${GKE_CLUSTER_URI}"
else
  CMD="${CMD} --kubeconfig=${KCF} --context=${KCC}"
fi
retry 2 run "${CMD}"
```

asmcli/commands/install.sh
```diff
@@ -74,21 +74,19 @@ install_in_cluster_control_plane() {
...
-    PARAMS="${PARAMS} -f ${BETA_CRD_MANIFEST}"
-  fi
-  PARAMS="${PARAMS} -c ${KUBECONFIG}"
   PARAMS="${PARAMS} --skip-confirmation"
 
+  print_config >| "${RAW_YAML}"
+  istioctl manifest generate \
+    <"${RAW_YAML}" \
+    >|"${EXPANDED_YAML}"
+
```
zerobfd, line 89: Why is this moved before installation?

asmcli/lib/util.sh
```diff
@@ -0,0 +1,22 @@
+# expects var PARAM already defined and populates it with the
+# appropriate manifest overlays
+init_install_params() {
```
zerobfd, line 3: I would much rather this `echo` the params at the end, and then call it with `PARAMS="$(gen_install_params)"`

Even better would be to have an inner function which took the CA and istioctlFiles as arguments with the actual logic, so that we can inject test params in the inner function and still be able to use the no-argument version for actual client logic.

zerobfd (changes requested):

asmcli/commands/install.sh
backward-compatible: it was for debugging later i found another way to accomplish it so will move it back down.

asmcli/asmcli.sh
backward-compatible: sweet

asmcli/asmcli.sh
backward-compatible: done

asmcli/asmcli.sh
backward-compatible: thanks for the exact snippet!

asmcli/asmcli.sh
```diff
@@ -922,10 +932,19 @@ EOF
...
-    FLEET_ID="${PROJECT_ID}"
-    context_set-option "FLEET_ID" "${FLEET_ID}"
+  if is_gcp; then
+    # GCP only: when no Fleet Id is provided, default to the cluster's project as the Fleet host.
+    if [[ -z "${FLEET_ID}" ]]; then
+      FLEET_ID="${PROJECT_ID}"
+      context_set-option "FLEET_ID" "${FLEET_ID}"
+    fi
+  else
+    # non-GCP only: set Project Id to same as Fleet Id.
```
zerobfd, line 942: nit: I don't think we need "GCP only" and "non-GCP" right below the if statement that checks whether it's GCP or not.

zerobfd (approved):
lgtm. Please dd a unit test for the happy path through gen_install_params before committing.

asmcli/asmcli.sh
backward-compatible: assuming that this is about the comment being redundant - i will fix that.

asmcli/lib/util.sh
backward-compatible: made the former modification. i can refactor this into more decouple functions in the next pass to simplify unit testing.

PR #767, opened 2021-06-29
rename add-to-mesh subcommand

ZhengzheYang:


asmcli/commands/create-mesh.sh
```diff
@@ -127,6 +127,12 @@ install_one_remote_secret() {
   local CTX_CLUSTER2; CTX_CLUSTER2="${2}"
   local SECRET_NAME; SECRET_NAME="${CTX_CLUSTER1//_/-}"
 
+  if [[ "${#SECRET_NAME}" -gt 63 ]]; then
+    local RAND
+    RAND="$(tr -dc "a-z0-9" </dev/urandom | head -c8 || true)"
+    SECRET_NAME="${SECRET_NAME:0:54}-${RAND}"
+  fi
+
```
zerobfd, line 135: I think it's probably better here to use something deterministic and not random, to handle the case where the mesh gets re-created with two overlapping but distinct sets of clusters.

Let's assume 50 as a reasonable upper bound on how many clusters will be in a mesh. 20 chars gives a space of 2^20 bits, and that means any two hashes have a ~0.12% chance of colliding, so taking the first 20 chars of a hash and then using the existing secret name feels like a decent tradeoff between human readability and preventing collisions.

You can generate the hash with sha256sum:
```
$ echo "test" | sha256sum
f2ca1bb6c7e907d06dafe4687e579fce76b37e4e93b7605022da52e6ccc26fd2  -
```

zerobfd (changes requested):

asmcli/commands/create-mesh.sh
```diff
@@ -154,3 +154,15 @@ prepare_add_to_mesh_environment() {
     organize_kpt_files
   fi
 }
+
+generate_secret_name() {
+  local SECRET_NAME; SECRET_NAME="${1}"
+
+  if [[ "${#SECRET_NAME}" -gt 63 ]]; then
+    local HASH
+    HASH="$(echo "${SECRET_NAME}" | sha256sum | head -c20 || true)"
```
zerobfd, line 163: very nit: the data is a "digest", the function is a "hash"

asmcli/commands/create-mesh.sh
```diff
@@ -154,3 +154,15 @@ prepare_add_to_mesh_environment() {
     organize_kpt_files
   fi
 }
+
+generate_secret_name() {
+  local SECRET_NAME; SECRET_NAME="${1}"
+
+  if [[ "${#SECRET_NAME}" -gt 63 ]]; then
```
zerobfd, line 161: very nit: since we do this same check somewhere else, maybe we should pull KUBE_TAG_MAX_LEN into a readonly variable in util.sh?

zerobfd (approved):

asmcli/commands/create-mesh.sh
ZhengzheYang: Done

asmcli/commands/create-mesh.sh
ZhengzheYang: Done

asmcli/commands/create-mesh.sh
ZhengzheYang: Done

PR #769, opened 2021-06-29
Add migrate-istio script

zerobfd:


scripts/migration/migrate-istio
```diff
@@ -0,0 +1,1360 @@
...
+PRE_MIGRATE=0
+INSTALL_ASM=0
+RELABEL=0
+FINALIZE=0
+ROLLBACK=0
+
+ASM_VERSION="1.10.2-asm.2"
+KPT_URL="https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages"
+KPT_URL="${KPT_URL}.git/asm@release-1.10-asm"; readonly KPT_URL
+REVISION_LABEL="asm-1102-2"; readonly REVISION_LABEL
```
shankgan, line 28: nit: any intention of allowing users to configure this in future? What if users want to jump to ASM 1.11?

scripts/migration/migrate-istio
```diff
@@ -0,0 +1,1360 @@
...
+    --format=json
+}
+
+valid_pool_query() {
+  cat <<EOF | tr '\n' ' '
+    config.machineType.split(sep="-").slice(-1:) >= $1
+EOF
+}
+
+validate_node_pool() {
```
shankgan, line 993: nit: leaving this as a remainder to add this to install_asm, if not already added

shankgan (approved):

scripts/migration/migrate-istio
```diff
@@ -0,0 +1,1360 @@
...
+  local TOKEN; TOKEN="$(retry 2 gcloud --project="${PROJECT_ID}" auth print-access-token)"
+  echo "${TOKEN}"
+}
+
+auth_header() {
+  local TOKEN; TOKEN="${1}"
+  echo "--header \"Authorization: Bearer ${TOKEN}\""
+}
+
+relabel_for_asm() {
```
shankgan, line 1130: nit: Could this accept an argument to relabel a particular namespace for canary based migration? 
Also, migration of namespaces could also be done using revision tags, refer https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/scripts/migration/migrate-addon#L355

scripts/migration/migrate-istio
```diff
@@ -0,0 +1,1360 @@
...
+  retry 2 run istioctl install $PARAMS
+
+  # Prevent the stderr buffer from ^ messing up the terminal output below
+  sleep 1
+
+  local RAW_YAML; RAW_YAML="${REVISION_LABEL}-manifest-raw.yaml"
+  local EXPANDED_YAML; EXPANDED_YAML="${REVISION_LABEL}-manifest-expanded.yaml"
+
+  PARAMS="-f ${OPERATOR_MANIFEST}"
+  if [[ "${UPGRADE_GATEWAYS}" -eq 0 ]]; then
```
shankgan, line 1091: nit: In the help strings, should we say that UPGRADE_GATEWAYS (default option) will result in a new IP?

scripts/migration/migrate-istio
zerobfd: I'd rather give us the opportunity to test it internally with a new version before allowing anyone to set the version at runtime.

scripts/migration/migrate-istio
zerobfd: I don't think it should result in a new IP since the service isn't changing. It didn't happen to me during testing. If it happened to you let me know since that's pretty unexpected.

scripts/migration/migrate-istio
zerobfd: Updated to do this namespace by namespace

PR #771, opened 2021-06-30
Adding TrustDomain Alias to scriptaro based on option

shankgan:


myidpt (approved):

zerobfd (changes requested):
I'm OK with the approach, but since we're now having this logic in two different places, can you add a variable at the top, the line after HUB_IDP_URL, that initializes TRUST_DOMAIN_ALIASES to "", and then create an add_trust_domain function that will append to it?

scripts/asm-installer/install_asm
```diff
@@ -2491,15 +2494,15 @@ configure_package() {
...
     # set the trust domain aliases to include both new Hub WIP and old Hub WIP to achieve no downtime upgrade.
-    TRUST_DOMAIN_ALIASES="${PROJECT_ID}.svc.id.goog ${PROJECT_ID}.hub.id.goog"
+    TRUST_DOMAIN_ALIASES="${PROJECT_ID}.svc.id.goog ${PROJECT_ID}.hub.id.goog ${TRUST_DOMAIN_ALIASES}"
     run eval kpt cfg set asm anthos.servicemesh.trustDomainAliases "${TRUST_DOMAIN_ALIASES}"
   fi
 
   if [[ "${CA}" == "mesh_ca" && -n "${_CI_TRUSTED_GCP_PROJECTS}" ]]; then
     # Gather the trust domain aliases from projects.
-    TRUST_DOMAIN_ALIASES="${PROJECT_ID}.svc.id.goog"
+    TRUST_DOMAIN_ALIASES="${PROJECT_ID}.svc.id.goog ${TRUST_DOMAIN_ALIASES}"
```
williamaronli, line 2503: Do we append the `${PROJECT_ID}.svc.id.goog twice` if TRUST_DOMAIN_ALIASES is appended in the if [[ "${CA}" == "mesh_ca" ]] and ten we append one more time here?

zerobfd:
Also to @williamaronli 's point, in the add_trust_domain function we should probably check to see if the string already exists in TRUST_DOMAIN_ALIASES with a pattern match.

scripts/asm-installer/install_asm
shankgan: depends on whether the previous logic is triggered.. that issue might predate this PR

williamaronli (approved):

williamaronli (approved):
lgtm

scripts/asm-installer/install_asm
```diff
@@ -642,6 +643,14 @@ needs_asm() {
   fi
 }
 
+add_trust_domain_alias() {
+  if [[ "${TRUST_DOMAIN_ALIASES}" == *${1}* ]]; then
+    echo "${1} already present in trust domain alias list"
```
zerobfd, line 648: We shouldn't echo anything. If there are duplicates then it's our fault, not whoever is running the script as the end user.

zerobfd (approved):
lgtm after removing echo

scripts/asm-installer/install_asm
shankgan: done

PR #774, opened 2021-06-30
Download kpt when necessary

zerobfd:
With the upcoming 1.0 release of `kpt`, we need to make sure that users get a version that works with the config here. This will download `kpt` if it's necessary for the action taken (install vs print config etc.) and one of the following:
 * the system version is too new
 * there isn't a `kpt` binary in the specified OUTPUT_DIR already


zerobfd:
The tests are failing, but tests that are sufficient to use the new code path are passing. There seems to be some kind of flake with the ingress gateway.

ZhengzheYang (approved):
LGTM

PR #780, opened 2021-07-01
Allow VM setup for MCP

jasonwzm:
The requirements are:
- Cluster should be registered to a fleet
- GCE IdentityProvider should be registered
- Service mesh feature should be enabled

### DIFF
--- a/asmcli/cloudbuild-main.yaml
+++ b/asmcli/cloudbuild-main.yaml
@@ -170,6 +170,44 @@ steps:
       - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
     timeout: 600s
 
+  - name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
+    dir: 'asmcli'
+    id: 'run-basic-suite-vm'
+    entrypoint: '/bin/bash'
+    args:
+    - '-c'
+    - >
+      ./tests/run_basic_suite_vm
+      --PROJECT_ID "${PROJECT_ID}"
+      --BUILD_ID "${BUILD_ID}"
+    env:
+    - '_LT_CLUSTER_NAME=long-term-test-cluster-environ'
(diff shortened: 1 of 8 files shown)


asmcli/commands/validate.sh
```diff
@@ -37,17 +38,23 @@ validate_dependencies() {
...
     if can_modify_gcp_components; then
       enable_workload_identity
       enable_stackdriver_kubernetes
+      if should_enable_service_mesh_feature; then
+        enable_service_mesh_feature
+      fi
     else
       exit_if_no_workload_identity
       exit_if_stackdriver_not_enabled
+      if should_enable_service_mesh_feature; then
```
zerobfd, line 55: Let's not use `should_` here. We need a new function `needs_servicemesh_feature` like `needs_asm`. The fact that we require something is separate from whether we should enable it or not.

zerobfd (changes requested):

asmcli/commands/validate.sh
jasonwzm: Done

asmcli/cloudbuild-main.yaml
```diff
@@ -170,6 +170,44 @@ steps:
...
+      --BUILD_ID "${BUILD_ID}"
+    env:
+    - '_LT_CLUSTER_NAME=long-term-test-cluster-environ'
+    - 'SERVICE_ACCOUNT=${_SERVICE_ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com'
+    - 'KEY_FILE=${_KEY_FILE}'
+    - '_CI_ASM_IMAGE_LOCATION=${_IMAGE_LOCATION}'
+    - '_CI_ASM_PKG_LOCATION=${_PKG_LOCATION}'
+    - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
+    timeout: 600s
+
```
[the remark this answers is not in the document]
jasonwzm: Removed and added unit test.

zerobfd (approved):

PR #788, opened 2021-07-03
V2

zerobfd:
Cherry-picking all of the important commits from master.


asmcli/asmcli
```diff
@@ -2717,6 +2797,8 @@ FLAGS:
   --version
 EOF
 }
+MANAGED_CNI="${OPTIONS_DIRECTORY}/cni-managed.yaml"; readonly MANAGED_CNI;
```
ZhengzheYang, line 2800: is there a reason for this var to be here instead of with all other global readonly vars in `init()`?

asmcli/asmcli
```diff
@@ -2767,6 +2849,37 @@ install_managed_components() {
...
+kind: ConfigMap
+metadata:
+  name: asm-options
+  namespace: istio-system
+data:
+  ASM_OPTS: "CNI=${CNI}"
+EOF
+
+    context_append "kubectlFiles" "mcp_configmap.yaml"
+  fi
```
ZhengzheYang, line 2878: so we apply this config map even when we want the CNI to be off?

ZhengzheYang (approved):
I mainly reviewed `asmcli/asmcli` and LGTM.

asmcli/asmcli
zerobfd: Yep.

asmcli/asmcli
zerobfd: Mostly because I don't think it actually needs to be global so I wanted to scope it to the file, but I'm not confident enough to move it to a local var yet.

asmcli/tests/run_basic_suite_managed
```diff
@@ -15,7 +15,7 @@ main() {
   # CLI setup
   parse_args "$@"
 
-  run_basic_test "install" "mesh_ca" "--managed"; RETVAL=$?;
+  run_basic_test "install" "mesh_ca" "--managed -o cni-managed"; RETVAL=$?;
```
backward-compatible, line 18: Please verify the file path. Looks like its erroring out because it tried to use absolute path maybe?

`asmcli: Running: '/usr/bin/kubectl --kubeconfig /tmp/tmp.tHmcwE8h75 --context gke_asm-scriptaro-oss_us-central1-c_long-term-test-cluster apply --overwrite=true -f /cni-managed.yaml'
asmcli: -------------
error: the path "/cni-managed.yaml" does not exist`

asmcli/asmcli.sh
```diff
@@ -2282,14 +2339,39 @@ configure_package() {
...
+    if [[ -n "${_CI_TRUSTED_GCP_PROJECTS}" ]]; then
+      # Gather the trust domain aliases from projects.
+      while IFS=',' read -r trusted_gcp_project; do
+        add_trust_domain_alias "${trusted_gcp_project}.svc.id.goog"
+      done <<EOF
+${_CI_TRUSTED_GCP_PROJECTS}
+EOF
+    fi
+    # shellcheck disable=SC2046
+    kpt cfg set asm anthos.servicemesh.trustDomainAliases $(context_get-option "TRUST_DOMAIN_ALIASES")
```
ZhengzheYang, line 2375: should this be 
`run_command eval kpt cfg set asm anthos.servicemesh.trustDomainAliases "${TRUST_DOMAIN_ALIASES}"`

also, should this if block exist in `configure_ca` in `components/ca/mesh-ca.sh` instead of there?

zerobfd: Yes and yes. However, I'm going to put those in another PR because the test isn't passing now since it requires yaml files that are in this commit. I'll commit this as-is and then add another PR with this fix and make sure tests pass.

PR #792, opened 2021-07-08
Move Mesh CA logic to proper location

zerobfd:


ZhengzheYang (approved):
LGTM as long as the test starts passing. I will continue debugging as well.

zerobfd:
I added some refactoring. I'll probably go ahead and merge if tests pass, but if you could take another quick look just to make sure I didn't do something stupid I'd appreciate it.

PR #795, opened 2021-07-08
automate release

ZhengzheYang:
automate the release for ASM 1.10

### DIFF
--- a/scripts/release-asm/common.sh
+++ b/scripts/release-asm/common.sh
@@ -1,4 +1,7 @@
 _DEBUG="${_DEBUG:=}"
+SCRIPT_PATH="${SCRIPT_PATH:=asmcli}"
+RELEASE_MODE="${RELEASE_MODE:=default}"
+PRECOMMIT="scripts/release-asm/precommit"
 if [[ "${_DEBUG}" -eq 1 ]]; then
   gsutil() {
     echo "DEBUG: would have run 'gsutil ${*}'" >&2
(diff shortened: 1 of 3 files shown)


scripts/release-asm/common.sh
```diff
@@ -1,4 +1,7 @@
 _DEBUG="${_DEBUG:=}"
+SCRIPT_PATH="${SCRIPT_PATH:=asmcli}"
+RELEASE_MODE="${RELEASE_MODE:=default}"
+PRECOMMIT="scripts/release-asm/precommit"
```
zerobfd, line 4: Looks like you never actually use this

scripts/release-asm/common.sh
```diff
@@ -236,3 +243,112 @@ setup() {
...
+  fi
+}
+
+update_config_number() {
+  local OLD_NUMBER; OLD_NUMBER="${1}"
+  local NEW_NUMBER; NEW_NUMBER="${2}"
+
+  sed -i.bak "s/CONFIG_VER:=\"${OLD_NUMBER}\"/CONFIG_VER:=\"${NEW_NUMBER}\"/g" "${SCRIPT_PATH}.sh"
+
+  bazel build //:merge && cp ../bazel-bin/asmcli .
```
zerobfd, line 272: I think maybe you meant to use the pre-release here?

zerobfd (changes requested):
All of the logic looks fine. It looks like none of it actually uses much stuff that's already in `common.sh`, so it feels like that's not the best place for it.

Maybe at first we should have a separate script for the "pre-release" steps, and then combine it with the existing release script/logic after we're comfortable with using it. What do you think?

scripts/release-asm/common.sh
ZhengzheYang: the `precommit` script assumes it's at the package root so I have to re-write the steps here.

zerobfd (approved):

PR #796, opened 2021-07-09
Add experimental subcommand

zerobfd:
Also adds a fix to the variable lint to skip if it doesn't find any.


ZhengzheYang (approved):

ZhengzheYang:
Thanks for fixing the bug 😁

PR #803, opened 2021-07-13
refactor the asmcli.sh to use functions from lib/

ZhengzheYang:
Split the logic in `asmcli/asmcli.sh` to fils in `lib/`. Added `lib/checks.sh` and `lib/iam.sh`, in which contain relevant util functions.


asmcli/commands/install.sh
```diff
@@ -206,3 +206,68 @@ install_control_plane() {
     install_canonical_controller
   fi
 }
+
+### Installation functions ###
```
zerobfd, line 210: We shouldn't need to keep these kind of comments after splitting things up into files.

asmcli/commands/install.sh
```diff
@@ -206,3 +206,68 @@ install_control_plane() {
...
+configure_package() {
```
zerobfd, line 211: `configure_package` is also used to print config. I'm not sure exactly the best place to put it--maybe lib/config.sh or something?

In general, the stuff in "commands" I'd like to be as tightly scoped to just that command as possible.

asmcli/lib/iam.sh
```diff
@@ -0,0 +1,148 @@
+required_iam_roles_mcp_sa() {
```
zerobfd, line 1: About the file name: there's a lot of stuff in here that doesn't have anything to do with IAM. Maybe grouping all of the prerequisite requirements and associated functions into one file would be a better first step? Just IAM seems too narrow.

asmcli/lib/util.sh
```diff
@@ -22,3 +22,978 @@ gen_install_params() {
...
+  echo "Alternatively, use --enable_all|-e to allow this tool to handle all dependencies."
+}
+
+arg_required() {
+  if [[ ! "${2:-}" || "${2:0:1}" = '-' ]]; then
+    fatal "Option ${1} requires an argument."
+  fi
+}
+
+parse_args() {
```
zerobfd, line 151: I think arg parsing functions probably don't belong in util.sh. Especially since we're adding `experimental` etc. and subcommands might have different parsing logic, we should probably have a dedicated file for shared arg parsing/validation logic.

asmcli/lib/util.sh
```diff
@@ -22,3 +22,978 @@ gen_install_params() {
...
+  fi
+
+  ln -s "${ISTIOCTL_REL_PATH}" .
+
+  info "Downloading ASM kpt package..."
+  retry 3 kpt pkg get --auto-set=false "${KPT_URL}" asm
+  version_message > "${ASM_VERSION_FILE}"
+}
+
+#######
```
zerobfd, line 440: I think from here down to `get_auth_token` would be good candidates for combining with the stuff in `iam.sh` and calling it "installation_dependencies.sh" or something.

zerobfd (changes requested):
Overall looks a lot better! Thanks for putting the effort into this.

asmcli/commands/install.sh
ZhengzheYang: Done

asmcli/commands/install.sh
ZhengzheYang: Put it to lib/config.sh

asmcli/lib/util.sh
ZhengzheYang: Done.

asmcli/lib/iam.sh
ZhengzheYang: Put it to `installation_dependencies.sh`

asmcli/lib/util.sh
ZhengzheYang: Done.

zerobfd (approved):
Thanks! This is a huge first step.

I think for the next step, we should probably take some of the logic out of commands/validate.sh and put it into lib/validate.sh, just so we keep the logic for "running the validate command" and "actually validating things" separate. We should absolutely merge this as-is though.

zerobfd (approved):

PR #805, opened 2021-07-13
Bring in the latest istio cni manfiest.

bianpengyuan:
This is the manifest from istio master. We will migrate AFC install so this probably will never be shipped with ASM 1.11. Mostly this is to make prow test happy.

cc @richardwxn @ruigulala


zerobfd (approved):

bianpengyuan:
@zerobfd can you help merging this?

PR #807, opened 2021-07-14
Fix registration

jasonwzm:
Unquote the command so that it is not treated as a file to run.

### DIFF
--- a/asmcli/asmcli.sh
+++ b/asmcli/asmcli.sh
@@ -2061,7 +2061,7 @@ register_cluster() {
   else
     CMD="${CMD} --kubeconfig=${KCF} --context=${KCC}"
   fi
-  retry 2 "${CMD}"
+  retry 2 ${CMD}
 }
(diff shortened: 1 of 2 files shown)


asmcli/asmcli.sh
```diff
@@ -2061,7 +2061,8 @@ register_cluster() {
   else
     CMD="${CMD} --kubeconfig=${KCF} --context=${KCC}"
   fi
-  retry 2 "${CMD}"
+  # shellcheck disable=SC2086
+  retry 2 ${CMD}
```
zerobfd, line 2065: Can you also add a `run_command` between 2 and ${CMD}?

zerobfd (approved):

asmcli/asmcli.sh
jasonwzm: done.

zerobfd (approved):

PR #811, opened 2021-07-15
Add x vm prepare-cluster subcommand

jasonwzm:
Most of it is copied from `asm_vm` with some changes aligning to asmcli.

Manually tested for both unmanaged ASM and MCP.


asmcli/commands/experimental/vm.sh
```diff
@@ -0,0 +1,438 @@
...
+}
+
+init_vm() {
+  ASM_REVISIONS=""
+
+  EXPANSION_GATEWAY_NAME="istio-eastwestgateway"; readonly EXPANSION_GATEWAY_NAME
+  ASM_REVISION_LABEL_KEY="istio.io/rev"; readonly ASM_REVISION_LABEL_KEY
+
+  INSTALL_EXPANSION_GATEWAY=0
+  INSTALL_IDENTITY_PROVIDER=0
```
zerobfd, line 18: These should be added to `context.sh`. We're trying to avoid any global variable that aren't readonly.

asmcli/commands/experimental/vm.sh
```diff
@@ -0,0 +1,438 @@
...
+  fi
+
+  # shellcheck disable=SC2064
+  trap "$(shopt -p nocasematch)" RETURN
+  shopt -s nocasematch
+  if [[ "$(uname -m)" != "x86_64" ]]; then
+    fatal "Installation is only supported on x86_64."
+  fi
+}
+
```
zerobfd, line 279: You should just be able to use the existing `validate_cli_dependencies`, all of the dependencies there are either optional or included in the same packages as these.

asmcli/commands/experimental/vm.sh
```diff
@@ -0,0 +1,438 @@
...
+    || true)
+
+  if [[ -z "${RESULT}" ]]; then
+    { read -r -d '' MSG; fatal "${MSG}"; } <<EOF
+Unable to find project ${PROJECT_ID}. Please verify the spelling and try
+again. To see a list of your projects, run:
+  gcloud projects list --format='value(project_id)'
+EOF
+  fi
+}
```
zerobfd, line 297: Should be able to use existing.

asmcli/commands/experimental/vm.sh
```diff
@@ -0,0 +1,438 @@
...
+
+validate_cluster_registration() {
+  info "Verifying cluster fleet registration..."
+  if ! is_cluster_registered; then
+    { read -r -d '' MSG; fatal "${MSG}"; } <<EOF || true
+Cluster is not registered to a Fleet. Please make sure your cluster is
+registered and try again.
+EOF
+  fi
+}
```
zerobfd, line 307: Should be able to use existing.

zerobfd (changes requested):
Logic looks good, a couple of style points. Also once https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/803 is merged, we should make sure this rebases cleanly. I don't think anything should affect this but just in case.

asmcli/commands/experimental/vm.sh
jasonwzm: Done

asmcli/commands/experimental/vm.sh
jasonwzm: Done

asmcli/commands/experimental/vm.sh
jasonwzm: Done.

asmcli/commands/experimental/vm.sh
jasonwzm: Done.

zerobfd (approved):

PR #814, opened 2021-07-16
Use New Hub WIP as default for GCP

jtrbs:
This PR will use new Hub WIP as default option for unmanaged GCP with MeshCA. 
1. The command to install ASM with MeshCA with asmcli is the following. 
```
./asmcli install \
  --project_id $PROJECT_ID \
  --cluster_name $CLUSTER_NAME \
  --cluster_location $CLUSTER_LOCATION \
  --enable_all
```
In this case, we will help user register the cluster  and use Hub WIP. 

2. If the user specified `hub-meshca` option, populate a message to tell user that `hub-meshca` is not needed. 

3. When user does not specify enable_all or enable_registration, and the cluster has not been registered into Hub before the installation, user is allowed to install ASM with GKE WIP.

### DIFF
--- a/asmcli/commands/validate.sh
+++ b/asmcli/commands/validate.sh
@@ -36,12 +36,6 @@ validate_dependencies() {
     exit_if_apis_not_enabled
   fi
 
-  if can_register_cluster && is_gcp; then
-    register_cluster
-  elif should_validate && [[ "${USE_HUB_WIP}" -eq 1 || "${USE_VM}" -eq 1 ]]; then
-    exit_if_cluster_unregistered
-  fi
-
   if is_gcp; then
     if can_modify_gcp_components; then
       enable_workload_identity
@@ -56,6 +50,12 @@ validate_dependencies() {
         exit_if_service_mesh_feature_not_enabled
       fi
     fi
+  fi
+
+  if can_register_cluster && is_gcp; then
+    register_cluster
+  elif should_validate && [[ "${USE_HUB_WIP}" -eq 1 || "${USE_VM}" -eq 1 ]]; then
+    exit_if_cluster_unregistered
   fi
 
   get_project_number
@@ -118,6 +118,14 @@ validate_hub() {
 
   if ! is_managed && [[ "${USE_VM}" -eq 1 && "${USE_HUB_WIP}" -eq 0 ]]; then
     fatal "Hub Workload Identity Pool is required to add VM workloads. Run the script with the -o hub-meshca option."
+  fi
+
+  if ! is_managed && [[ "${CA}" == "mesh_ca" && "${USE_HUB_WIP}" -eq 0 ]] && ( can_register_cluster && is_gcp || is_cluster_registered ) ; then
+    info "Fleet workload identity pool is used as default for Mesh CA. "
+    context_set-option "USE_HUB_WIP" 1
+    local OPTIONAL_OVERLAY; OPTIONAL_OVERLAY="$(context_get-option "OPTIONAL_OVERLAY")"
+    OPTIONAL_OVERLAY="hub-meshca,${OPTIONAL_OVERLAY}"
+    context_set-option "OPTIONAL_OVERLAY" "${OPTIONAL_OVERLAY}"
   fi
 }
 
--- a/asmcli/components/control-plane/in-cluster.sh
+++ b/asmcli/components/control-plane/in-cluster.sh
@@ -27,13 +27,13 @@ configure_in_cluster_control_plane() {
 
 init_meshconfig() {
   local USE_HUB_WIP; USE_HUB_WIP="$(context_get-option "USE_HUB_WIP")"
-  local HUB_MEMBERSHIP_ID; HUB_MEMBERSHIP_ID="$(context_get-option "HUB_MEMBERSHIP_ID")"
   local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
   local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
 
   info "Initializing meshconfig API..."
   if [[ "${USE_HUB_WIP}" -eq 1 ]]; then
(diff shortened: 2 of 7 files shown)


asmcli/commands/validate.sh
```diff
@@ -58,6 +52,12 @@ validate_dependencies() {
     fi
   fi
 
+  if can_register_cluster && is_gcp; then
```
jtrbs, line 55: Cluster registration depends on GKE workload identity enablement. Thus move this section after enable_workload_identity.

asmcli/components/control-plane/in-cluster.sh
```diff
@@ -27,13 +27,13 @@ configure_in_cluster_control_plane() {
 
 init_meshconfig() {
   local USE_HUB_WIP; USE_HUB_WIP="$(context_get-option "USE_HUB_WIP")"
-  local HUB_MEMBERSHIP_ID; HUB_MEMBERSHIP_ID="$(context_get-option "HUB_MEMBERSHIP_ID")"
```
jtrbs, line 30: HUB_MEMBERSHIP_ID in the context has not been set when cluster is registered with asmcli. Thus move this line after the populate_fleet_info.

asmcli/commands/validate.sh
```diff
@@ -119,6 +119,14 @@ validate_hub() {
   if ! is_managed && [[ "${USE_VM}" -eq 1 && "${USE_HUB_WIP}" -eq 0 ]]; then
     fatal "Hub Workload Identity Pool is required to add VM workloads. Run the script with the -o hub-meshca option."
   fi
+
+  if ! is_managed && [[ "${CA}" == "mesh_ca" && "${USE_HUB_WIP}" -eq 0 ]] && ( can_register_cluster && is_gcp || is_cluster_registered ) ; then
+    info "Fleet workload identity pool is used as default for Mesh CA. "
+    context_set-option "USE_HUB_WIP" 1
```
jtrbs, line 125: If the user does not specify `hub-meshca` option, `USE_HUB_WIP` will be set to 1 (use Hub WIP as default) when we know either the cluster has been registered into Hub or we can help user register the cluster. Otherwise, `USE_HUB_WIP` is still 0 and user can continue to use GKE WIP.

myidpt (approved):

asmcli/commands/validate.sh
```diff
@@ -119,6 +119,14 @@ validate_hub() {
...
+    local OPTIONAL_OVERLAY; OPTIONAL_OVERLAY="$(context_get-option "OPTIONAL_OVERLAY")"
+    OPTIONAL_OVERLAY="hub-meshca,${OPTIONAL_OVERLAY}"
```
zerobfd, line 127: Is there anything in this overlay that's still necessary that's not already included with MeshCA?

asmcli/commands/validate.sh
zerobfd: If you're on GCP and you have workload identity enabled and you register the cluster, what's the difference between GKE and Fleet WIP?

asmcli/commands/validate.sh
```diff
@@ -119,6 +119,14 @@ validate_hub() {
   if ! is_managed && [[ "${USE_VM}" -eq 1 && "${USE_HUB_WIP}" -eq 0 ]]; then
     fatal "Hub Workload Identity Pool is required to add VM workloads. Run the script with the -o hub-meshca option."
   fi
+
+  if ! is_managed && [[ "${CA}" == "mesh_ca" && "${USE_HUB_WIP}" -eq 0 ]] && ( can_register_cluster && is_gcp || is_cluster_registered ) ; then
+    info "Fleet workload identity pool is used as default for Mesh CA. "
```
zerobfd, line 124: Let's remove this info, I don't think most customers care.

asmcli/lib/parse.sh
```diff
@@ -70,6 +70,7 @@ parse_args() {
         OPTIONAL_OVERLAY="${2},${OPTIONAL_OVERLAY}"
         context_set-option "OPTIONAL_OVERLAY" "${OPTIONAL_OVERLAY}"
         if [[ "${2}" == "hub-meshca" ]]; then
+          info "Fleet workload identity pool is used as default for Mesh CA. No need to specify hub-meshca option."
           context_set-option "USE_HUB_WIP" 1
```
zerobfd, line 74: If there's no need to specify it, then we shouldn't need to set the USE_HUB_WIP context, right?

zerobfd (changes requested):

asmcli/commands/validate.sh
jtrbs: Hi Clay, in this case, there is no difference in WIP.

asmcli/lib/parse.sh
jtrbs: USE_HUB_WIP is set to 1 because the Hub configuration flow (reading Hub resources and parse Hub meta such as IdP) is controlled by this flag. Maybe as Hub is default,  USE_HUB_WIP can have the default value 1 and will only be set to 0 when user does not specify enable_all or enable_registration, and the cluster has not been registered into Hub before the installation. Basically when `! is_managed && [[ "${CA}" == "mesh_ca" ]] && ( is_gcp && ! can_register_cluster && ! is_cluster_registered )`. Do you think this is better?

asmcli/commands/validate.sh
jtrbs: You mean not covered by https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asm/istio/istio-operator.yaml? There are some setters which have different behavior, especially when it is multiproject. We can merge the istio-operator together with hub-meshca.yaml. In this case, some UG needs to be changed like the gke multiproject installation guide since setters will be changed in the istio operator. Maybe we can do this merge in 1.11 and hopefully at that time gke multiproject installation guide will be retired?

PR #816, opened 2021-07-16
add experimental install and align with AFC

ZhengzheYang:


asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+    exit 2
+  fi
+
+  # shellcheck disable=SC2064
+  trap "$(shopt -p nocasematch)" RETURN
+  shopt -s nocasematch
+
+  REVISION_LABEL_REGULAR="asm-managed"; readonly REVISION_LABEL_REGULAR
+  REVISION_LABEL_RAPID="asm-managed-rapid"; readonly REVISION_LABEL_RAPID
+  REVISION_LABEL_STABLE="asm-managed-stable"; readonly REVISION_LABEL_STABLE
```
zerobfd, line 489: These don't change depending on experimental or not, can we just add them to init()?

asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+        shift 2
+        ;;
+      -e | --enable_all | --enable-all)
+        context_set-option "ENABLE_ALL" 1
+        shift 1
+        ;;
+      --enable_cluster_roles | --enable-cluster-roles)
+        context_set-option "ENABLE_CLUSTER_ROLES" 1
+        shift 1
+        ;;
```
zerobfd, line 530: Cluster roles shouldn't be necessary anymore, since the installer isn't getting run as whoever is running the script.

asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+      --enable_registration | --enable-registration)
+        context_set-option "ENABLE_REGISTRATION" 1
+        shift 1
+        ;;
+      --enable_namespace_creation | --enable-namespace-creation)
+        context_set-option "ENABLE_NAMESPACE_CREATION" 1
+        shift 1
+        ;;
```
zerobfd, line 554: I believe AFC takes care of this, can you confirm?

asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+      --disable_canonical_service | --disable-canonical-service)
+        context_set-option "DISABLE_CANONICAL_SERVICE" 1
+        shift 1
+        ;;
```
zerobfd, line 558: I'm pretty sure we don't want to allow this when we start doing AFC provisioning, since it's not going to be in-cluster.

asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+      --print_config | --print-config)
+        context_set-option "PRINT_CONFIG" 1
+        shift 1
+        ;;
```
zerobfd, line 562: Different command

asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+      -s | --service_account | --service-account)
+        arg_required "${@}"
+        context_set-option "SERVICE_ACCOUNT" "${2}"
+        shift 2
+        ;;
+      -k | --key_file | --key-file)
+        arg_required "${@}"
+        context_set-option "KEY_FILE" "${2}"
+        shift 2
+        ;;
```
zerobfd, line 572: We shouldn't need this anymore

asmcli/lib/parse.sh
```diff
@@ -473,3 +473,157 @@ arg_required() {
...
+      --only_enable | --only-enable)
+        context_set-option "ONLY_ENABLE" 1
+        shift 1
+        ;;
+      --platform)
+        arg_required "${@}"
+        context_set-option "PLATFORM" "$(echo "${2}" | tr '[:upper:]' '[:lower:]')"
+        shift 2
+        ;;
```
zerobfd, line 594: None of these should be required

asmcli/lib/validate.sh
```diff
@@ -0,0 +1,153 @@
+x_validate_install_args() {
```
zerobfd, line 1: Question: does it make more sense to you to separate the experimental-only lib functions from the rest of them? I don't have a strong opinion yet.

zerobfd (changes requested):

asmcli/lib/parse.sh
ZhengzheYang: Confirmed with David, no, AFC does not create the `istio-system` namespace.

asmcli/lib/parse.sh
ZhengzheYang: Confirmed with David, the managed controller will run if AFC does not see an in-cluster one. Asked in the chat room but probably we don't need this flag.

asmcli/lib/parse.sh
ZhengzheYang: Synced offline that this is still needed to enable dependencies.

asmcli/lib/validate.sh
ZhengzheYang: Synced offline to have separate parse and validation for the experimental commands.

asmcli/lib/parse.sh
ZhengzheYang: Done

asmcli/lib/parse.sh
ZhengzheYang: Done

asmcli/lib/parse.sh
ZhengzheYang: Done

asmcli/lib/parse.sh
ZhengzheYang: Done

zerobfd (approved):

PR #826, opened 2021-07-21
Add kubeconfig (GKE only) support to create-mesh subcommand.

backward-compatible:
The subcommand can accept N kubeconfig files, each of which is expected
to contain a default context that will be the ONLY context used. The UX
for N clusters would be to supply N differennt files, each of which will
have a default context set to the corresponding target cluster.

### DIFF
--- a/asmcli/commands/create-mesh.sh
+++ b/asmcli/commands/create-mesh.sh
@@ -22,23 +22,46 @@ parse_cluster_args() {
   fi
 
   while [[ $# != 0 ]]; do
-    if [[ "$1" =~ .+/.+/.+ ]]; then
-      local CLUSTER; CLUSTER="${1}"
-      context_append "clustersInfo" "${CLUSTER//\// }"
-    else
+    if [ -e "$1" ]; then
       local KCF; KCF="${1}"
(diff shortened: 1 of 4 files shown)


asmcli/commands/create-mesh.sh
```diff
@@ -22,8 +22,13 @@ parse_cluster_args() {
   fi
 
   while [[ $# != 0 ]]; do
-    local CLUSTER; CLUSTER="${1}"
-    context_append "clustersInfo" "${CLUSTER//\// }"
+    if [[ "$1" =~ .+/.+/.+ ]]; then
```
zerobfd, line 25: I think we should check the opposite thing: if `$1` is a file, then we treat it like a kubeconfigFile, otherwise assume it's a clustersInfo. I don't want to mistake `asmcli create-mesh path/to/kubeconfig` as the wrong thing.

zerobfd (changes requested):

asmcli/commands/create-mesh.sh
backward-compatible: ah good point, thanks!

PR #827, opened 2021-07-21
 Add kubeconfig (GKE only) support to create-mesh subcommand

backward-compatible:
The subcommand can accept N kubeconfig files, each of which is expected
to contain a default context that will be the ONLY context used. The UX
for N clusters would be to supply N differennt files, each of which will
have a default context set to the corresponding target cluster.

### DIFF
--- a/asmcli/commands/create-mesh.sh
+++ b/asmcli/commands/create-mesh.sh
@@ -22,23 +22,46 @@ parse_cluster_args() {
   fi
 
   while [[ $# != 0 ]]; do
-    if [[ "$1" =~ .+/.+/.+ ]]; then
-      local CLUSTER; CLUSTER="${1}"
-      context_append "clustersInfo" "${CLUSTER//\// }"
-    else
+    if [ -e "$1" ]; then
       local KCF; KCF="${1}"
       context_append "kubeconfigFiles" "${KCF}"
+    else
+      local CLUSTER; CLUSTER="${1}"
+      context_append "clustersInfo" "${CLUSTER//\// }"
     fi
     shift 1
   done
 }
 
 validate_cluster_args() {
+  local KCF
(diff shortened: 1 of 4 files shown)


asmcli/commands/create-mesh.sh
```diff
@@ -22,18 +22,46 @@ parse_cluster_args() {
   fi
 
   while [[ $# != 0 ]]; do
-    local CLUSTER; CLUSTER="${1}"
-    context_append "clustersInfo" "${CLUSTER//\// }"
+    if [ -e "$1" ]; then
```
zerobfd, line 25: We should use `-f` instead of `-e`. (see https://tldp.org/LDP/abs/html/fto.html)

asmcli/commands/create-mesh.sh
```diff
@@ -22,18 +22,46 @@ parse_cluster_args() {
...
+  local PROJECT_ID
+  local CLUSTER_LOCATION
+  local CLUSTER_NAME
+  local CTX_CLUSTER
+  local GKE_CLUSTER_URI
+
   # validate fleet id is valid
   get_project_number
 
+  # flatten any kubeconfig files into cluster P/L/C
```
zerobfd, line 47: Can you add "this is GCP-only and will need to be reworked for other platforms" in a comment here as well? Just to make it more obvious.

zerobfd (approved):

asmcli/commands/create-mesh.sh
backward-compatible: Ack

asmcli/commands/create-mesh.sh
backward-compatible: yes, thanks!

PR #828, opened 2021-07-21
deprecate --only-validate and fix parse bug 

ZhengzheYang:
this PR does
1. show deprecation message for `--only-validate`
2. fix the parsing bug where user gets no meaningful info after the parsing is delegated to each subcommand.

### DIFF
--- a/asmcli/lib/parse.sh
+++ b/asmcli/lib/parse.sh
@@ -148,9 +148,9 @@ parse_args() {
         shift 1
         ;;
       --only_validate | --only-validate)
-        warn "As of version 1.10 the --only_validate flag is deprecated and will be ignored."
+        warn "As of version 1.11 the --only_validate flag is deprecated and will be ignored."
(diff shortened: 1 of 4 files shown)


ZhengzheYang:
@zerobfd I deleted the lines that directly `exit` if the user does `asmcli install`. Instead it will go to validation to at least show some meaningful information.

asmcli/lib/parse.sh
```diff
@@ -148,6 +143,8 @@ parse_args() {
         shift 1
         ;;
       --only_validate | --only-validate)
+        warn "As of version 1.11 the --only_validate flag is deprecated and will be ignored."
```
zerobfd, line 146: I think "In version 1.11" is a bit more clear than "As of", since the latter implies that it might be already released.

zerobfd (approved):

asmcli/lib/parse.sh
ZhengzheYang: Done

PR #830, opened 2021-07-22
add usage and refactor create-mesh

ZhengzheYang:
added usage for create-mesh and refactor while we are at it

### DIFF
--- a/asmcli/cloudbuild-main.yaml
+++ b/asmcli/cloudbuild-main.yaml
@@ -135,6 +135,24 @@ steps:
     #############
   - name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
     dir: 'asmcli'
+    id: 'run-create-mesh'
+    entrypoint: '/bin/bash'
+    args:
+      - '-c'
+      - >
+        ./tests/run_create_mesh
+        --PROJECT_ID "${PROJECT_ID}"
+        --BUILD_ID "${BUILD_ID}"
+    env:
+      - 'SERVICE_ACCOUNT=${_SERVICE_ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com'
+      - 'KEY_FILE=${_KEY_FILE}'
+      - '_CI_ASM_IMAGE_LOCATION=${_IMAGE_LOCATION}'
+      - '_CI_ASM_PKG_LOCATION=${_PKG_LOCATION}'
+      - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
+    timeout: 600s
+    
+  - name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
+    dir: 'asmcli'
     id: 'run-basic-suite-managed'
     entrypoint: '/bin/bash'
     args:
--- a/asmcli/cloudbuild-main.yaml
+++ b/asmcli/cloudbuild-main.yaml
@@ -135,24 +135,6 @@ steps:
     #############
   - name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
     dir: 'asmcli'
-    id: 'run-create-mesh'
-    entrypoint: '/bin/bash'
-    args:
-      - '-c'
-      - >
-        ./tests/run_create_mesh
-        --PROJECT_ID "${PROJECT_ID}"
-        --BUILD_ID "${BUILD_ID}"
-    env:
-      - 'SERVICE_ACCOUNT=${_SERVICE_ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com'
-      - 'KEY_FILE=${_KEY_FILE}'
-      - '_CI_ASM_IMAGE_LOCATION=${_IMAGE_LOCATION}'
-      - '_CI_ASM_PKG_LOCATION=${_PKG_LOCATION}'
-      - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
-    timeout: 600s
-    
-  - name: 'gcr.io/$PROJECT_ID/${_IMAGE_NAME}'
-    dir: 'asmcli'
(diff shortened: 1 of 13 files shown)


asmcli/commands/help.sh
```diff
@@ -249,3 +249,52 @@ FLAGS:
...
+create-mesh_usage() {
+  cat << EOF
+${SCRIPT_NAME} $(version_message)
+usage: ${SCRIPT_NAME} create-mesh FLEET_ID (PROJECT_ID/CLUSTER_LOCATION/CLUSTER_NAME | KUBECONFIG_PATH) ... 
+
+Create a service mesh using the GKE clusters provided by registering them to the fleet and install remote secrets on each
+pair of clusters.
+Cluster information and KUBECONFIG could be passed in interchangeably.
+You could pass in as many clusters as you want, as long as they are not already registered to other fleets.
+Clusters already registered to the specified fleet will be skipped, whereas those registered to other fleets will fail.
```
zerobfd, line 262:
```
Create a multi-cluster service mesh and allow cross-cluster service discovery. This is done by
registering the clusters to the specified fleet and installing the necessary Kubernetes cluster
secrets to allow ASM to communicate securely.

You can pass in an arbitrary number of clusters at once, either by specifying the project ID,
cluster location, and cluster name (for GKE on GCP clusters only) or the location of a
kubeconfig file that has the default context set to the desired cluster.

When adding/removing clusters to an existing multi-cluster fleet, you must specify the entire set
of clusters for every invocation. You cannot add or remove clusters one at a time.

This command will fail if any of the specified clusters are registered to a fleet other than the one
specified by FLEET_ID.
```

asmcli/commands/help.sh
```diff
@@ -249,3 +249,52 @@ FLAGS:
...
+
+  $> ${SCRIPT_NAME} \\
+      create-mesh \\
+      my-fleet \\
+      my-project/us-central1-c/my-cluster1 \\
+      my-project/us-central1-c/my-cluster2 \\
+      my-project/us-central1-c/my-cluster3 \\
+      path/to/kubeconfig1 \\
+      path/to/kubeconfig2 \\ 
+      path/to/kubeconfig3 \\
```
zerobfd, line 282: I think we could probably just get away with

```
my-project/us-central1-c/my-cluster1 \\
path/to/kubeconfig2 \\ 
my-project/us-central1-c/my-cluster3
```

(Also note that the command would fail if you had trailing `\\` on the last line.)

asmcli/commands/help.sh
```diff
@@ -249,3 +249,52 @@ FLAGS:
...
+
+EOF
+}
+
+create-mesh_usage_short() {
+  cat << EOF
+${SCRIPT_NAME} $(version_message)
+usage: ${SCRIPT_NAME} create-mesh FLEET_ID (PROJECT_ID/CLUSTER_LOCATION/CLUSTER_NAME | KUBECONFIG_PATH) ... 
+
+Create a service mesh using the GKE clusters provided by registering them to the fleet.
```
zerobfd, line 292: Create a multi-cluster service mesh and allow cross-cluster service discovery.

asmcli/lib/util.sh
```diff
@@ -244,6 +244,26 @@ prepare_environment() {
   fi
 }
 
+# Need to prepare differently under multicluster environment
+# validate_cluster and configure_kubectl will be called in validation
+# for each cluster
+prepare_create_mesh_environment() {
```
zerobfd, line 250: If this is create_mesh specific I think we can leave it in the command file for now.

asmcli/lib/validate.sh
```diff
@@ -694,6 +712,59 @@ arg_required() {
   fi
 }
 
+validate_create_mesh_args() {
```
zerobfd, line 715: If this is create_mesh specific I think we can leave it in the command file for now.

zerobfd (changes requested):
As you're refactoring, see if you can find any opportunities to change functions that use `context_get-option` at the top and instead change them to take in arguments to the function.

asmcli/commands/help.sh
ZhengzheYang: Done

asmcli/commands/help.sh
ZhengzheYang: Done

asmcli/commands/help.sh
ZhengzheYang: Done

asmcli/lib/util.sh
ZhengzheYang: Done

asmcli/lib/validate.sh

asmcli/cloudbuild-main.yaml
```diff
@@ -133,6 +133,24 @@ steps:
...
+        --PROJECT_ID "${PROJECT_ID}"
+        --BUILD_ID "${BUILD_ID}"
+    env:
+      - 'SERVICE_ACCOUNT=${_SERVICE_ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com'
+      - 'KEY_FILE=${_KEY_FILE}'
+      - '_CI_ASM_IMAGE_LOCATION=${_IMAGE_LOCATION}'
+      - '_CI_ASM_PKG_LOCATION=${_PKG_LOCATION}'
+      - '_CI_ASM_KPT_BRANCH=${_ASM_PKG_BRANCH}'
+    timeout: 600s
+
```
zerobfd, line 153: It's probably a good idea to run it on this PR since it's refactoring, but let's not add another slow test to the `main` file. Can you put this in the `staging` version before merging?

zerobfd (approved):
Looks great! Thanks!

asmcli/cloudbuild-main.yaml
ZhengzheYang: yes that was my plan. Just want to wait till it's approved. 😉

PR #837, opened 2021-07-23
GCP GA integration changes

shankgan:


zerobfd:
You should run 'scripts/release-asm/precommit' to re-generate `asmcli` and also commit that.

asmcli/commands/install.sh
```diff
@@ -83,21 +83,20 @@ install_private_ca() {
...
-  local ISTIOD_SERVICE_ACCOUNT
-  ISTIOD_SERVICE_ACCOUNT="istiod-${REVISION_LABEL}"
-  if [[ "${_CI_NO_REVISION}" -eq 1 ]]; then
-    ISTIOD_SERVICE_ACCOUNT="istiod"
-  fi
-  local WORKLOAD_IDENTITY; WORKLOAD_IDENTITY="${WORKLOAD_POOL}[istio-system/${ISTIOD_SERVICE_ACCOUNT}]"
-  local NAME; NAME=$(echo "${CA_NAME}" | cut -f6 -d/)
+  local CA_NAME; CA_NAME="$(context_get-option "CA_NAME")"
+
+  local WORKLOAD_IDENTITY; WORKLOAD_IDENTITY="${WORKLOAD_POOL}[-/-]"
```
shankgan, line 88: Unsure if WORKLOAD_POOL needs to be retrieved from context. It doesn't appear to be set anywhere

myidpt (approved):

asmcli/commands/install.sh
zerobfd: Does everything work correctly when you test locally?

shankgan: yes

PR #840, opened 2021-07-23
Set env variables explicitly

zerobfd:
kpt doesn't merge the maps, so we should set all of the env variable for pilot explicitly

### DIFF
--- a/asmcli/lib/config.sh
+++ b/asmcli/lib/config.sh
@@ -26,6 +26,7 @@ configure_package() {
     kpt cfg set asm gcloud.compute.network "${GCE_NETWORK_NAME}"
   else
     kpt cfg set asm gcloud.core.project "${FLEET_ID}"
+    kpt cfg set asm anthos.servicemesh.controlplane.monitoring.enabled "false"
   fi
 
   kpt cfg set asm gcloud.project.environProjectNumber "${PROJECT_NUMBER}"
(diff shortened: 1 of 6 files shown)


asmcli/lib/config.sh
```diff
@@ -26,6 +26,7 @@ configure_package() {
     kpt cfg set asm gcloud.compute.network "${GCE_NETWORK_NAME}"
   else
     kpt cfg set asm gcloud.core.project "${FLEET_ID}"
+    kpt cfg set asm anthos.servicemesh.controlplane.monitoring.enabled "false"
```
jtrbs, line 29: This part may not be true. In the case ! is_gcp and "${CA}" = "meshca", kpt cfg set asm anthos.servicemesh.controlplane.monitoring.enabled "true". With Hub registration, for onprem, stackdriver is supported.

zerobfd: Thanks, I didn't know that! Changed to `if ! is_gcp && ! is_cluster_registered`

ZhengzheYang (approved):

asmcli/lib/config.sh
```diff
@@ -26,7 +25,9 @@ configure_package() {
     kpt cfg set asm gcloud.compute.network "${GCE_NETWORK_NAME}"
   else
     kpt cfg set asm gcloud.core.project "${FLEET_ID}"
-    kpt cfg set asm anthos.servicemesh.controlplane.monitoring.enabled "false"
+    if ! is_cluster_registered; then
```
jtrbs, line 28: Maybe checking CA is better? cluster registration will be required thus this condition will not be met. However, in the case of Citadel, within ASM source code, stackdriver will not be setup. Thus kpt cfg set asm anthos.servicemesh.controlplane.monitoring.enabled "true" in this case maybe misleading.

zerobfd: Changed, thanks again.

jtrbs: No problem!

jtrbs (approved):

PR #847, opened 2021-07-27
only-validate will not exit

### LINKED ISSUE #829: Don't exit on validation errors if this is explicitly a validation run
It makes sense to abort the process if a validation error is found during installation, but if a customer is explicitly asking us to validate via `--only-validate` or `asmcli validate`, then we should collect the issues as they occur and then print them and exit 1 at the end.

ZhengzheYang:
Fix #829

### DIFF
--- a/asmcli/commands/validate.sh
+++ b/asmcli/commands/validate.sh
@@ -16,8 +16,14 @@ validate() {
   validate_control_plane
 
   if [[ "${ONLY_VALIDATE}" -eq 1 ]]; then
-    info "Successfully validated all requirements to install ASM."
-    exit 0
+    local VALIDATION_ERROR; VALIDATION_ERROR="$(context_get-option "VALIDATION_ERROR")"
+    if [[ "${VALIDATION_ERROR}" -eq 0 ]]; then
+      info "Successfully validated all requirements to install ASM."
+      exit 0
+    else
+      info "Validation failed. Please see the errors for guidance to successfully install ASM."
+      exit 2
+    fi
   fi
 
   if only_enable; then
--- a/asmcli/commands/validate.sh
+++ b/asmcli/commands/validate.sh
@@ -15,13 +15,13 @@ validate() {
   validate_dependencies
   validate_control_plane
 
-  if [[ "${ONLY_VALIDATE}" -eq 1 ]]; then
+  if [[ "${ONLY_VALIDATE}" -ne 0 ]]; then
     local VALIDATION_ERROR; VALIDATION_ERROR="$(context_get-option "VALIDATION_ERROR")"
     if [[ "${VALIDATION_ERROR}" -eq 0 ]]; then
       info "Successfully validated all requirements to install ASM."
       exit 0
     else
-      info "Validation failed. Please see the errors for guidance to successfully install ASM."
+      warn "Please see the errors above."
       exit 2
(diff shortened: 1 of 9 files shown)


asmcli/commands/validate.sh
```diff
@@ -16,8 +16,14 @@ validate() {
...
 
   if [[ "${ONLY_VALIDATE}" -eq 1 ]]; then
-    info "Successfully validated all requirements to install ASM."
-    exit 0
+    local VALIDATION_ERROR; VALIDATION_ERROR="$(context_get-option "VALIDATION_ERROR")"
+    if [[ "${VALIDATION_ERROR}" -eq 0 ]]; then
+      info "Successfully validated all requirements to install ASM."
+      exit 0
+    else
+      info "Validation failed. Please see the errors for guidance to successfully install ASM."
```
zerobfd, line 24: I'd change "info" to "warn". I think we can also just say "Please see the errors above."

asmcli/lib/util.sh
```diff
@@ -117,6 +117,15 @@ info() {
   fi
 }
 
+validation_error() {
+  error "${1}"
+  if only_validate; then
+    context_set-option "VALIDATION_ERROR" 1
```
zerobfd, line 123: I think it would be nice if we could increment a counter here instead of just setting it to 1. That would let us print out the number of errors when we exit, and also make it easier to test.

zerobfd (changes requested):
Approach and code look good. Can you see if there's a way to add a unit test to make sure just validating doesn't exit early? If you change `validation_error` to count errors, I think it should be possible by asserting that the VALIDATION_ERROR context var is greater than 1.

asmcli/commands/validate.sh
ZhengzheYang: Done

asmcli/lib/util.sh
ZhengzheYang: Done

ZhengzheYang:
Couldn't find a clean/easy way to test the validation error logic. Manual test results attached here:
```
asmcli: Checking required APIs...
asmcli: Checking for project zhengzhe-anthos-test...
asmcli: Reading labels for us-central1-c/cluster-1...
asmcli: Checking for istio-system namespace...
asmcli: [ERROR]: The istio-system namespace doesn't exist.
Please create the "istio-system" and retry, or run the script with the
'--enable_namespace_creation' flag to allow the script to enable it on your behalf.
Alternatively, use --enable_all|-e to allow this tool to handle all dependencies.
asmcli: Confirming node pool requirements for zhengzhe-anthos-test/us-central1-c/cluster-1...
asmcli: Checking Istio installations...
asmcli: [WARNING]: There is no way to validate that the meshconfig API has been initialized.
asmcli: [WARNING]: This needs to happen once per GCP project. If the API has not been initialized
asmcli: [WARNING]: for zhengzhe-anthos-test, please re-run this tool with the --enable_gcp_components
asmcli: [WARNING]: flag. Otherwise, installation will succeed but Anthos Service Mesh
asmcli: [WARNING]: will not function correctly.
asmcli: [WARNING]: Please see the errors above.
```

Now even though `istio-system` validation has error, it will not exit but continue.

zerobfd (approved):

PR #848, opened 2021-07-28
allow override runistiod data in testing

richardwxn:
we should be able to override it in testing environment


richardwxn:
@zerobfd

scripts/asm-installer/install_asm
```diff
@@ -2728,7 +2729,9 @@ EOF
   fi
 
   local CR_IMAGE_JSON; CR_IMAGE_JSON="";
-  if [[ -n "${_CI_CLOUDRUN_IMAGE_HUB}" ]]; then
+  if [[ -n "${_CI_CLOUDRUN_RUNISTIOD_DATA}" ]]; then
+    CR_IMAGE_JSON="${_CI_CLOUDRUN_RUNISTIOD_DATA}"
+  elif [[ -n "${_CI_CLOUDRUN_IMAGE_HUB}" ]]; then
     CR_IMAGE_JSON=$(cat <<EOF
 {"image": "${_CI_CLOUDRUN_IMAGE_HUB}:${_CI_CLOUDRUN_IMAGE_TAG}"}
 EOF
```
zerobfd, line 2737: If we're going to add an option to replace the entire `CR_IMAGE_JSON` var, can we just get rid of `_CI_CLOUDRUN_IMAGE_HUB` and `_CI_CLOUDRUN_IMAGE_TAG` and use this one exclusively?

zerobfd (changes requested):

scripts/asm-installer/install_asm
richardwxn: I think we can continue keeping the options for `_CI_CLOUDRUN_IMAGE_HUB` and `_CI_CLOUDRUN_IMAGE_TAG` since for most cases we just need to set hub and tag value, we don't need to provide the whole json data.  The `CR_IMAGE_JSON` just provide a more advanced override that we can provide the json data directly.

Also the existing variables are used in multiple places like testing and doc, so it would be backward compatible to keep them.

zerobfd: I don't doubt that it's more convenient, but I don't think we need to introduce additional complexity into the code that the client runs in order to get it. It would be simple to have the logic for turning a hub/tag into the JSON blob outside of the script and then pass it in. That would make this PR remove one variable instead of add one as well as a new branch condition.

richardwxn: Yep, I agree that it would not be hard turning a hub/tag into the JSON blob outside of the script but it means that we have to do this logic everywhere, in the prow tests, doc, etc. And people are familiar with the way how we override hub/tag, which aligns with other Istio components. 
I believe the code added(a new variable and a branch condition) does not add much to complexity to the maintenance of the code since it does not change the structure and it is not a backward incompatible change.

richardwxn:
/CC @mandarjog this is needed for testing MDP with MCP in our integration test

scripts/asm-installer/install_asm
richardwxn: @zerobfd gentle ping~

zerobfd: Sorry for the delayed response, this fell off of my todo list somehow.

I'm strongly against having this level of complexity just to support internal testing unless absolutely necessary. If you come up with a way to get us from this PR to a state where we don't have separate variables for `HUB` and `TAG` then I'm fine with it. Otherwise, we should figure out how to make this work in one PR.

PR #850, opened 2021-07-29
Use Hub WIP as default

jtrbs:
1. Combine hub-meshca overlay together with istio-operator. 
2. Use Hub WIP as default.


jtrbs:
Take over https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/814

shankgan:
quick q: Does this assume that every cluster where this script installs ASM is registered to a fleet that has Hub workload Identity turned on?

jtrbs:
Hi @shankgan, 
For baremetal and onprem, the fleet registration is a prerequisite before executing asmcli.
For GKE-on-GCP
- If it can be registered (--enable-all or --enable-registration or --vm), asmcli will help register the cluster if it has not been registered. 
- If it cannot be registered with asmcli and it has no hub regisration, GKE WIP will be used.

zerobfd (approved):
LGTM. This change works with both MeshCA and Citadel off of GCP, right?

jtrbs:
Hi @zerobfd, yeah this does not impact off GCP installation. For of GCP, I have tested both MeshCA and Citadel.

PR #856, opened 2021-08-04
Use context cluster instead of name to detect GKE naming scheme

backward-compatible:


zerobfd (approved):
Looks good. If you have the time, I'd do this:

1) take the `cut -c2- | sed 's/\([[:blank:]]\)*\1/\1/g' | cut -d ' ' -f 3` and put it into a separate function
2) change the subshell call to `kubectl config get-contexts --no-headers | new_function`
3) add a unit test on new_function by doing something like asserting `echo '* name context extra' | new_function` is `context` (I did that from memory so double check the column order if that doesn't work)

This is more to catch regressions than test correctness for this PR.

ZhengzheYang (approved):

asmcli/lib/util.sh
```diff
@@ -565,3 +565,12 @@ EOF
...
 }
+
+# Reads in stdin, expected in the form of kubectl config get-contexts
+# output and isolates the default context's cluster value.
+get_context_cluster() {
+  local CONTEXT_ROW
+  while IFS= read -r CONTEXT_ROW; do
+    echo "${CONTEXT_ROW}" | grep '^\*' | cut -c2- | sed 's/\([[:blank:]]\)*\1/\1/g' | cut -d ' ' -f 3
+  done
+}
```
zerobfd, line 576: I feel like you can just make the entire thing one line

```
get_context_cluster() {
  grep '^\*' | cut -c2- | sed 's/\([[:blank:]]\)*\1/\1/g' | cut -d ' ' -f 3
}
```

backward-compatible: yes, done.

PR #872, opened 2021-08-10
Apply topology.istio.io/network label to istio-system for MN topology

backward-compatible:


asmcli/commands/install.sh
```diff
@@ -192,6 +192,7 @@ install_control_plane() {
   if [[ "${MANAGED}" -eq 1 ]]; then
     install_managed_control_plane
   else
+    label_istio_namespace
```
backward-compatible, line 195: @zerobfd should this call be outside the surrounding if/else ? (In order to apply it for both in-cluster and MCP.)

zerobfd: yep

asmcli/lib/context.sh
```diff
@@ -45,7 +46,6 @@ context_init() {
     "USE_HUB_WIP": ${USE_HUB_WIP:-1},
     "USE_VM": ${USE_VM:-0},
     "USE_MCP_CNI": ${USE_MCP_CNI:-0},
-    "FLEET_ID": "${FLEET_ID:-}",
```
zerobfd, line 48: was this removed on purpose?

backward-compatible: yea that line is duplicated above..

asmcli/commands/install.sh
backward-compatible: done

zerobfd (approved):

PR #873, opened 2021-08-10
Update MDP CRD to v1 (#871)

bianpengyuan:


zerobfd (approved):

zerobfd:
the yaml linter doesn't like cni-managed.yaml, the indentation is wrong. It's good to merge as soon as that's fixed.

PR #880, opened 2021-08-13
add retry to kubectl api-resources

ZhengzheYang:
currently the calls to the api-server experiences some flakes and give errors like:
```
error: unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```

this issue does not affect prod clusters but does introduce a lot of test flakes. This PR adds retries to the api-resources calls.

### DIFF
--- a/asmcli/commands/experimental/vm.sh
+++ b/asmcli/commands/experimental/vm.sh
@@ -321,7 +321,7 @@ is_google_identity_provider_installed() {
 }
 
 is_idp_crd_installed() {
-  if ! kubectl api-resources --api-group=security.cloud.google.com | grep -q identityproviders; then
+  if ! retry 2 kubectl api-resources --api-group=security.cloud.google.com | grep -q identityproviders; then
     false
   fi
 }
--- a/asmcli/commands/experimental/vm.sh
+++ b/asmcli/commands/experimental/vm.sh
@@ -321,7 +321,8 @@ is_google_identity_provider_installed() {
 }
 
 is_idp_crd_installed() {
-  if ! retry 2 kubectl api-resources --api-group=security.cloud.google.com | grep -q identityproviders; then
+  if [[ "$(retry 2 kubectl get crd identityproviders.security.cloud.google.com -ojsonpath="{..metadata.name}" \
+    | grep -w -c identityproviders || true)" -eq 0 ]]; then
     false
   fi
 }
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -746,7 +746,7 @@ is_google_identity_provider_installed() {
 }
 
 is_idp_crd_installed() {
-  if ! kubectl api-resources --api-group=security.cloud.google.com | grep -q identityproviders; then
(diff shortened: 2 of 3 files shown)


asmcli/commands/experimental/vm.sh
```diff
@@ -321,7 +321,7 @@ is_google_identity_provider_installed() {
 }
 
 is_idp_crd_installed() {
-  if ! kubectl api-resources --api-group=security.cloud.google.com | grep -q identityproviders; then
+  if ! retry 2 kubectl api-resources --api-group=security.cloud.google.com | grep -q identityproviders; then
```
jasonwzm, line 324: If we use `kubectl get crd`, can we potentially avoid the error?

Also do we need a retry on the entirety of cluster registration verification?

zerobfd, line 324: If this can't use `get crd` and requires a `| grep` then let's split it into two steps.

```
local OUTPUT; OUTPUT="$(retry 2 kubectl api-resources...)"
if ! echo "${OUTPUT}" | grep -q identityproviders; then
```

asmcli/lib/installation_dependencies.sh
```diff
@@ -265,7 +265,7 @@ generate_secret_name() {
 }
 
 register_cluster() {
-  if is_cluster_registered; then return; fi
+  if retry 2 is_cluster_registered; then return; fi
```
zerobfd, line 268: Let's not retry entire functions unless it's absolutely necessary.

zerobfd (changes requested):

asmcli/commands/experimental/vm.sh
ZhengzheYang: can I use https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asmcli/asmcli#L1914-L1915? like

```
is_membership_crd_installed() {
  if [[ "$(retry 2 kubectl get crd memberships.hub.gke.io -ojsonpath="{..metadata.name}" \
    | grep -w -c memberships || true)" -eq 0 ]]; then
    false
    return
  fi

  if [[ "$(retry 2 kubectl get memberships.hub.gke.io -ojsonpath="{..metadata.name}" \
    | grep -w -c membership || true)" -eq 0 ]]; then
    false
  fi
}
```

zerobfd: That should also work, yea.

ZhengzheYang: Updated to `kubectl get crd`.  Clay advised against retry the entirety so I am leaving as is.

ZhengzheYang: Done

asmcli/lib/installation_dependencies.sh
ZhengzheYang: Done

zerobfd (approved):

jasonwzm (approved):

PR #882, opened 2021-08-16
Add better limits for canonical service controller

zerobfd:


zerobfd:
This also fixes an issue with MCP provisioning when not using Hub WIP.

scripts/asm-installer/install_asm
```diff
@@ -2135,7 +2135,12 @@ init_meshconfig() {
 
 init_meshconfig_managed() {
   info "Initializing meshconfig managed API..."
-  local POST_DATA; POST_DATA='{"workloadIdentityPools":["'${ENVIRON_PROJECT_ID}'.hub.id.goog","'${ENVIRON_PROJECT_ID}'.svc.id.goog"], "prepare_istiod": true}'
+  local POST_DATA
+  if [[ "${USE_HUB_WIP}" -eq 1 && -n "${ENVIRON_PROJECT_ID}" ]]; then
+    POST_DATA='{"workloadIdentityPools":["'${ENVIRON_PROJECT_ID}'.hub.id.goog","'${ENVIRON_PROJECT_ID}'.svc.id.goog"], "prepare_istiod": true}'
+  else
+    POST_DATA='{"workloadIdentityPools":["'${PROJECT_ID}'.hub.id.goog","'${PROJECT_ID}'.svc.id.goog"], "prepare_istiod": true}'
```
ZhengzheYang, line 2142: why we still pass the workload identity pool even when `USE_HUB_WIP` is false?

ZhengzheYang (approved):

scripts/asm-installer/install_asm
zerobfd: Forward compatibility. This just makes it easier in the future to do migrations in case the format changes.

PR #883, opened 2021-08-17
initialie meshconfig on fleet_id

ZhengzheYang:


asmcli/lib/util.sh
```diff
@@ -579,3 +579,13 @@ EOF
...
   grep '^\*' | cut -c2- | sed 's/\([[:blank:]]\)*\1/\1/g' | cut -d ' ' -f 3
 }
+
+init_meshconfig_curl() {
+  local POST_DATA; POST_DATA="${1}"
+  local ID; ID="${2}"
+  run_command curl --request POST --fail \
+    --data "${POST_DATA}" -o /dev/null \
+    "https://meshconfig.googleapis.com/v1alpha1/projects/${ID}:initialize" \
+    --header "Content-Type: application/json" \
```
zerobfd, line 589: I think we also need the `--header "X-Server-Timeout: 600"` here as well

zerobfd (approved):
LGTM except for the one issue, thanks!

asmcli/lib/util.sh
ZhengzheYang: Done

PR #892, opened 2021-08-20
Migration script tweaks for MCP

zerobfd:


scripts/migration/migrate-to-asm
```diff
@@ -1360,13 +1595,20 @@ main() {
     download_asm
     configure_kubectl
     pre_migrate
+    if should_use_mcp; then
+      run_mcp_fit_assessment
```
howardjohn, line 1599: `should_use_mcp` already calls `run_mcp_fit_assessment` so seems weird to run it here as well?

scripts/migration/migrate-to-asm
```diff
@@ -1337,13 +1495,90 @@ apply_backup() {
...
+    "${CLUSTER_NAME}" --format json \
+    | grep -q '"enablePrivateEndpoint": true,'
+}
+
+is_autopilot() {
+  gcloud container clusters describe \
+    --region "${CLUSTER_LOCATION}" \
+    --project "${PROJECT_ID}" \
+    "${CLUSTER_NAME}" --format json \
+    | grep -q 'autopilot'
```
howardjohn, line 1526: this doesn't seem that reliable, the name or anything else in the output could have "autopilot".

if we are avoiding jq, `--format` can take complex expressions like `--format='value(labels.instanceId)'`, etc that may be more precise

scripts/migration/migrate-to-asm
```diff
@@ -1337,13 +1495,90 @@ apply_backup() {
...
+}
+
+is_supported_location() {
+  while read -r region; do
+    if [[ "${CLUSTER_LOCATION}" = "${region}"* ]]; then
```
howardjohn, line 1531: is CLUSTER_LOCATION always region? never something like us-central1-c?

scripts/migration/migrate-to-asm
```diff
@@ -1337,13 +1495,90 @@ apply_backup() {
...
+    fi
+  done <<< "$(supported_mcp_regions)"
+  false
+}
+
+run_mcp_fit_assessment() {
+  echo "Analyzing cluster configuration to determine optimal control plane..."
+
+  if ! is_supported_location; then false; return; fi
+  if is_private_cluster; then false; return; fi
```
howardjohn, line 1542: should we output something here? So we understand why it decided what it did (at the very least for debugging tickets)

scripts/migration/migrate-to-asm
```diff
@@ -1337,13 +1495,90 @@ apply_backup() {
...
+  if is_autopilot; then false; return; fi
+
+  if [[ ! -f ./configuration_backup/default_iop ]]; then
+    istioctl profile dump >./configuration_backup/default_iop 2>/dev/null
```
howardjohn, line 1546: very nit: it feels safer to use an empty IOP than default profile, although functionally it does appear to generate the same result. But less that can go wrong with an empty one.

scripts/migration/migrate-to-asm
```diff
@@ -1337,13 +1495,90 @@ apply_backup() {
...
+  fi
+
+  local ARGS; ARGS="-f ./configuration_backup/default_iop"
+  if [[ -f ./configuration_backup/iop_backup ]]; then
+    ARGS="-f ./configuration_backup/iop_backup"
```
howardjohn, line 1551: is there only one CUSTOM_OVERLAY allowed? I thought I had seen folks passing multiple?

howardjohn (commented):
LGTM at a high level, few small comments

scripts/migration/migrate-to-asm
```diff
@@ -1360,13 +1595,20 @@ main() {
     download_asm
     configure_kubectl
     pre_migrate
+    if should_use_mcp; then
```
howardjohn, line 1598: I vaguely recall discussion about having a simple way to just run the check without needing to jump through a bunch of hoops. Are we still planning that? Or is it mostly a docs thing?

scripts/migration/migrate-to-asm
zerobfd: I didn't notice that, I just thought it was weird to just call "should_use_mcp". Thanks for pointing that out.

scripts/migration/migrate-to-asm
zerobfd: Nope, I'll change the variable name. Lazy copy/paste.

zerobfd: Nevermind, actually it is always region--these are the supported regions for MCP, not anything the user input.

scripts/migration/migrate-to-asm
zerobfd: Good idea.

scripts/migration/migrate-to-asm
zerobfd: I'll change it, no real reason to use default.

scripts/migration/migrate-to-asm
zerobfd: For this tool only one. Do Istio users generally keep their custom IOPs separate?

scripts/migration/migrate-to-asm
zerobfd: You can do it with pre-migrate, although that will also validate all of the other dependencies. Not sure if it's worth adding another option to just run the mcp fit assessment for this tool when peopel can just use it directly or use `asmcli`.

scripts/migration/migrate-to-asm
howardjohn: I think https://cloud.google.com/service-mesh/docs/enable-optional-features recommends it. I have seen it used a few times line that doc

zerobfd: Ah, I forgot we changed this to also target ASM customers and not just OSS. I'll see how to update this.

PR #923, opened 2021-08-30
CAS: use --ca-pool instead of --ca-name

shankgan:


asmcli/lib/parse.sh
```diff
@@ -53,7 +53,7 @@ parse_args() {
         context_set-option "CA" "$(echo "${2}" | tr '[:upper:]' '[:lower:]')"
         shift 2
         ;;
-      --ca_name | --ca-name)
+      --ca_name | --ca-name | --ca_pool | --ca-pool)
         arg_required "${@}"
```
zerobfd, line 57: If you plan to remove --ca_name, can you do something like

```
--ca_name | --ca-name)
  warn "The --ca_name argument is deprecated and will be removed in the next minor release, please use --ca_pool instead."
  warn "Execution will continue normally."
  sleep 1
  ;&
--ca_pool | --ca-pool)
  arg_required "${@}"
  ...
```

zerobfd (approved):

PR #929, opened 2021-08-31
add --managed flag to x install

ZhengzheYang:


asmcli/lib/validate.sh
```diff
@@ -760,6 +761,10 @@ CLUSTER_LOCATION
 CLUSTER_NAME
 EOF
 
+  if [[ "${MANAGED}" -eq 0 ]]; then
+    fatal "Only managed control plane installation is supported by experimental install."
```
zerobfd, line 765: Can you add "currently" so no one thinks that it will never be supported?

zerobfd (approved):

asmcli/lib/validate.sh

PR #933, opened 2021-09-01
enable mesh feature for mcp

ZhengzheYang:


zerobfd (approved):
We should go through and simplify all of the logic around the feature and fleets at some point since we're requiring it all the time.

PR #936, opened 2021-09-03
wait for crd to be installed by AFC

ZhengzheYang:


asmcli/commands/experimental/install.sh
```diff
@@ -17,6 +17,15 @@ x_install() {
 
   # TODO: MANAGED_CNI install should go here, if necessary
 
+  # `kubectl wait` for non-existent resources will return error directly so we wait in loop
+  info "Wait for the controlplanerevisions CRD to be installed by AFC. This should take a few minutes if cluster is newly registered."
+  for i in {1..10}; do
+    if kubectl wait --for condition=established --timeout=10s crd/controlplanerevisions.mesh.cloud.google.com 2>/dev/null; then
```
zerobfd, line 23: How did we pick 10s? Could we change it to 20s and remove the sleep 10 below?

zerobfd (approved):

asmcli/commands/experimental/install.sh
ZhengzheYang: this is taken directly from the AFC e2e test. The `--timeout=10s` doesn't do anything if the resource does not exist yet so the wait will return directly with an error. The actual wait is the `sleep 10` call. Let me know if there are better ways.

zerobfd: Ah, makes sense. I can't think of any better ways then.

PR #937, opened 2021-09-07
Make CNI always on at managed ASM.

bianpengyuan:
Remove the necessities of setting `cni-managed` option when installing managed ASM.

cc @mandarjog


zerobfd (approved):

mandarjog:
Is this for 1.11 ?

bianpengyuan:
Yes

PR #938, opened 2021-09-07
Update kpt value (gcloudl.container.cluster) for off-GCP unregistered cluster

backward-compatible:


zerobfd (approved):
LGTM as long as we verified that CLUSTER_NAME is always populated.

PR #947, opened 2021-09-08
Make CNI always on at managed ASM.

bianpengyuan:
Similar change as https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/937. Not sure if this is needed for install_asm though..


zerobfd (approved):
Can't hurt.

PR #951, opened 2021-09-08
Fix the ASM CAS installation

lei-tang:
This PR fixes the following problems of ASM CAS installation:
- add-iam-policy-binding should use the CA pool as its argument, instead of the CA name (in the form of "projects/project_name/locations/ca_region/caPools/ca_pool").
- The role "roles/privateca.auditor" should be added to the member "group:${WORKLOAD_IDENTITY}", instead of "serviceAccount:${WORKLOAD_IDENTITY}". When the role "roles/privateca.auditor" is added to "serviceAccount:${WORKLOAD_IDENTITY}", an error occurs.
- The --project is needed to specify the project ID.


zerobfd (approved):
LGTM but let's wait for @myidpt before merging

myidpt (approved):

PR #961, opened 2021-09-15
Add another condition to detect workload identity enablement

ZhiminXiang:
This condition could happen when enabling WI, and then disabling it, and re-enabling it.


zerobfd:
Thanks for the PR! The file you edited is actually generated and not the source file. If you look at https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/docs/developer-guide.md it should give a better idea.

This specific check should be in asmcli/lib/checks.sh

PR #962, opened 2021-09-16
Update create-mesh to support off-GKE clusters

backward-compatible:


asmcli/commands/create-mesh.sh
```diff
@@ -70,105 +76,76 @@ create-mesh_validate_args() {
...
-      if [[ "${CTX_CLUSTER1}" != "${CTX_CLUSTER2}" ]]; then
-        install_one_remote_secret "${CTX_CLUSTER1}" "${CTX_CLUSTER2}"
+  while read -r KCF1; do
+    while read -r KCF2; do
+      if [[ "${KCF1}" != "${KCF2}" ]]; then
+        install_one_remote_secret "${KCF1}" "${KCF2}"
       fi
     done <<EOF
-$(context_list "clusterContexts")
+$(m1context_list "kubeconfigFiles")
```
zerobfd, line 131: What's the `m1` here?

asmcli/lib/config.sh
```diff
@@ -66,14 +66,15 @@ configure_package() {
...
-  local CLUSTER_NAME; CLUSTER_NAME="${3}"
   local CONTEXT; CONTEXT="$(context_get-option "CONTEXT")"
   local KUBECONFIG; KUBECONFIG="$(context_get-option "KUBECONFIG")"
   local KUBECONFIG_SUPPLIED; KUBECONFIG_SUPPLIED="$(context_get-option "KUBECONFIG_SUPPLIED")"
 
   if [[ "${KUBECONFIG_SUPPLIED}" -eq 0 ]]; then
+    local PROJECT_ID; PROJECT_ID="${1}"
+    local CLUSTER_LOCATION; CLUSTER_LOCATION="${2}"
+    local CLUSTER_NAME; CLUSTER_NAME="${3}"
+
```
zerobfd, line 77: Was the intention here to be able to use `configure_kubectl` without passing in arguments?

asmcli/commands/create-mesh.sh
```diff
@@ -1,11 +1,17 @@
 create-mesh_subcommand() {
   ### Preparation ###
+
+  # using kubeconfig globally for create-mesh sub-command
+  context_set-option "KUBECONFIG_SUPPLIED" 1
```
zerobfd, line 5: I'd like to either change the name of this context variable, or change the logic that depends on it.

asmcli/commands/create-mesh.sh
```diff
@@ -1,11 +1,17 @@
...
+  # setting same behavior for all environments
+  context_set-option "PLATFORM" "multicloud"
```
zerobfd, line 7: Same as above

asmcli/commands/create-mesh.sh
```diff
@@ -70,105 +76,76 @@ create-mesh_validate_args() {
...
 
 install_all_remote_secrets() {
-  local CTX_CLUSTER1 CTX_CLUSTER2
-
-  while read -r CTX_CLUSTER1; do
-    while read -r CTX_CLUSTER2; do
-      if [[ "${CTX_CLUSTER1}" != "${CTX_CLUSTER2}" ]]; then
-        install_one_remote_secret "${CTX_CLUSTER1}" "${CTX_CLUSTER2}"
+  while read -r KCF1; do
+    while read -r KCF2; do
```
zerobfd, line 126: I like the shorter variable names, but could we pull the no-logic-change refactors into another PR that I can stamp and then rebase this PR on that commit? It makes it a bit difficult to review just the logic changes.

zerobfd (changes requested):
The overall logic flow makes sense, just a few discrete issues.

asmcli/commands/create-mesh.sh
backward-compatible: sorry that's remnant from my local tests..

asmcli/lib/config.sh
backward-compatible: yes

asmcli/commands/create-mesh.sh
backward-compatible: why/to what?

asmcli/commands/create-mesh.sh
backward-compatible: i'm going to move the PLATFORM set call more locally but anyway why/to what would you like it to be renamed?

asmcli/commands/create-mesh.sh
backward-compatible: this specific one is more than just renaming, i'm changing the usage from **context** to **kubeconfig** so had rename accordingly.

asmcli/lib/config.sh
backward-compatible: looking at the code as it is rn, i feel like this is no longer needed

backward-compatible:
merge conflict need to create a new PR

PR #969, opened 2021-09-17
CAS: Fixing Project ID

shankgan:
Derive Project ID from CA_NAME in case ca pool is setup in a different project

### DIFF
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -1348,16 +1348,16 @@ install_private_ca() {
   local WORKLOAD_IDENTITY; WORKLOAD_IDENTITY="${WORKLOAD_POOL}:/allAuthenticatedUsers/"
   local CA_LOCATION; CA_LOCATION=$(echo "${CA_NAME}" | cut -f4 -d/)
   local CA_POOL; CA_POOL=$(echo "${CA_NAME}" | cut -f6 -d/)
-  local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
+  local PROJECT; PROJECT=$(echo "${CA_NAME}" | cut -f2 -d/)
 
   retry 3 gcloud privateca pools add-iam-policy-binding "${CA_POOL}" \
-    --project "${PROJECT_ID}" \
+    --project "${PROJECT}" \
     --location "${CA_LOCATION}" \
     --member "group:${WORKLOAD_IDENTITY}" \
     --role "roles/privateca.workloadCertificateRequester"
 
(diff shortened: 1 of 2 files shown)


shankgan:
cherry-pick into release-1.11-asm branch if one exists

asmcli/asmcli
```diff
@@ -1348,7 +1348,7 @@ install_private_ca() {
   local WORKLOAD_IDENTITY; WORKLOAD_IDENTITY="${WORKLOAD_POOL}:/allAuthenticatedUsers/"
   local CA_LOCATION; CA_LOCATION=$(echo "${CA_NAME}" | cut -f4 -d/)
   local CA_POOL; CA_POOL=$(echo "${CA_NAME}" | cut -f6 -d/)
-  local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
+  local PROJECT_ID; PROJECT_ID=$(echo "${CA_NAME}" | cut -f2 -d/)
```
lei-tang, line 1351: This PR changes how to obtain PROJECT_ID. Will this PR impact the integration test and user guide?

asmcli/commands/install.sh
```diff
@@ -73,7 +73,7 @@ install_private_ca() {
   local WORKLOAD_IDENTITY; WORKLOAD_IDENTITY="${WORKLOAD_POOL}:/allAuthenticatedUsers/"
   local CA_LOCATION; CA_LOCATION=$(echo "${CA_NAME}" | cut -f4 -d/)
   local CA_POOL; CA_POOL=$(echo "${CA_NAME}" | cut -f6 -d/)
-  local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
+  local PROJECT_ID; PROJECT_ID=$(echo "${CA_NAME}" | cut -f2 -d/)
```
lei-tang, line 76: Ditto the comment above.

asmcli/asmcli
shankgan: It will affect integration test - previously, behavior was incorrect. No change to user guide

asmcli/commands/install.sh
shankgan: It will affect integration test - previously, behavior was incorrect. No change to user guide

lei-tang (approved):

zerobfd (approved):
Can you change the variable name to something like CA_PROJECT_ID? Otherwise it's going to be different from every other PROJECT_ID in the code.

PR #970, opened 2021-09-17
masterCas master project

shankgan:
Fixing where we get project ID from


lei-tang (approved):
Approve, same as https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/969/

zerobfd (approved):

PR #971, opened 2021-09-17
Update create-mesh to support off-GKE clusters

backward-compatible:


asmcli/commands/create-mesh.sh
```diff
@@ -70,105 +74,94 @@ create-mesh_validate_args() {
...
+create-mesh_set_platform() {
+  while read -r KCF; do
+    context_set-option "KUBECONFIG" "${KCF}"
+    if [[ "$(kubectl config current-context)" =~ "gke_" ]]; then
+      context_set-option "PLATFORM" "gcp"
+    else
+      context_set-option "PLATFORM" "multicloud"
+    fi
+    break
+  done <<EOF
```
zerobfd, line 121: If the intention here is to just get the first line, you can do it with head.
```
KFC="$(context_list "kubeconfigFiles" | head -n 1)"
if [[ -z "${KFC}" ]]; then fatal "Context is empty? Please file a GitHub issue."; fi
context_set-option "KUBECONFIG" ....
```

zerobfd (approved):

asmcli/commands/create-mesh.sh
backward-compatible: Yes, thanks!

asmcli/commands/create-mesh.sh
```diff
@@ -70,105 +74,94 @@ create-mesh_validate_args() {
...
-    --gke-uri="${GKE_CLUSTER_URI}" \
-    --enable-workload-identity
+create-mesh_register() {
+  while read -r KCF; do
+    context_set-option "KUBECONFIG" "${KCF}"
+    context_set-option "CONTEXT" "$(kubectl config current-context)"
+    is_cluster_registered
+    if ! is_membership_crd_installed; then
+      register_cluster
+    fi
```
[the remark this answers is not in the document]
backward-compatible: Thanks for catching that. I didn't realize that register call takes care of that anyway. Both the checks were redundant so removed them both.

PR #987, opened 2021-09-30
Check return code from runIstiod call

samertm:
Check the return code when calling runIstiod, fail if it's not a 200.

When successful, it logs this:
```
2021-09-30T20:36:30.992145 asmcli: Running: 'curl --write-out %{http_code} --silent --show-error --output /tmp/tmp.uMF7JCraIx --request POST https://meshconfig.googleapis.com/v1alpha1/projects/samer-asm-test2/locations/us-central1-b/clusters/somecluster:runIstiod --data {} --header X-Server-Timeout: 600 --header Content-Type: application/json -K /dev/fd/63'
2021-09-30T20:36:31.022427 asmcli: Running: '/usr/bin/gcloud --project=samer-asm-test2 auth print-access-token'
2021-09-30T20:36:31.041526 asmcli: -------------
2021-09-30T20:36:31.073035 asmcli: -------------
{}
```

This is what it looks like when it fails:

```
2021-09-30T20:40:35.943337 asmcli: Provisioning control plane...
2021-09-30T20:40:36.081721 asmcli: Running: 'curl --write-out %{http_code} --silent --show-error --output /tmp/tmp.4CDgHkLmOv --request POST localhost:8000 --data {} --header X-Server-Timeout: 600 --header Content-Type: application/json -K /dev/fd/63'
2021-09-30T20:40:36.111154 asmcli: Running: '/usr/bin/gcloud --project=samer-asm-test2 auth print-access-token'
2021-09-30T20:40:36.141329 asmcli: -------------
2021-09-30T20:40:36.178882 asmcli: -------------
HTTP response code: 500
some important error message
2021-09-30T20:40:37.133772 asmcli: [WARNING]: Failed, retrying...(1 of 2)
2021-09-30T20:40:39.263530 asmcli: Running: 'curl --write-out %{http_code} --silent --show-error --output /tmp/tmp.Sr5aSuZjoX --request POST localhost:8000 --data {} --header X-Server-Timeout: 600 --header Content-Type: application/json -K /dev/fd/63'
2021-09-30T20:40:39.311517 asmcli: -------------
HTTP response code: 500
some important error message
2021-09-30T20:40:39.371207 asmcli: [WARNING]: Failed, retrying...(2 of 2)
2021-09-30T20:40:41.440905 asmcli: [WARNING]: Command 'check_curl --request POST localhost:8000 --data {} --header X-Server-Timeout: 600 --header Content-Type: application/json -K /dev/fd/63' failed.
```

### DIFF
--- a/asmcli/lib/util.sh
+++ b/asmcli/lib/util.sh
@@ -42,6 +42,30 @@ run_command() {
   return $RETVAL
 }
 
+######
+# check_curl calls curl and returns non-zero if the http code is not 200 or the
+# curl command fails.
+######
+check_curl() {
+  local TMPFILE; TMPFILE=$(mktemp)
+
+  local HTTPCODE;
+  local RETVAL;
+  HTTPCODE=$(run_command curl --write-out '%{http_code}' --silent --show-error --output "$TMPFILE" "${@}")
+  RETVAL="$?"
+  if [[ "$RETVAL" != "0" ]] ; then
+    return "$RETVAL"
+  fi
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    warn "HTTP response code: ${HTTPCODE}"
+  fi
+  cat "$TMPFILE"
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    false
+    return
+  fi
+}
+
 #######
 # retry takes an integer N as the first argument, and a list of arguments
 # representing a command afterwards. It will retry the given command up to N
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -1917,7 +1917,7 @@ install_managed_control_plane() {
   fi
 
   info "Provisioning control plane..."
-  retry 2 run_command curl --request POST \
+  retry 2 check_curl --request POST \
     "https://meshconfig.googleapis.com/v1alpha1/projects/${PROJECT_ID}/locations/${CLUSTER_LOCATION}/clusters/${CLUSTER_NAME}:runIstiod" \
     --data "${POST_DATA}" \
     --header "X-Server-Timeout: 600" \
@@ -3396,6 +3396,30 @@ run_command() {
   local RETVAL
   { "${@}"; RETVAL="$?"; } || true
   return $RETVAL
+}
+
+######
+# check_curl calls curl and returns non-zero if the http code is not 200 or the
+# curl command fails.
+######
+check_curl() {
+  local TMPFILE; TMPFILE=$(mktemp)
+
+  local HTTPCODE;
+  local RETVAL;
+  HTTPCODE=$(run_command curl --write-out '%{http_code}' --silent --show-error --output "$TMPFILE" "${@}")
+  RETVAL="$?"
+  if [[ "$RETVAL" != "0" ]] ; then
+    return "$RETVAL"
+  fi
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    warn "HTTP response code: ${HTTPCODE}"
+  fi
+  cat "$TMPFILE"
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    false
+    return
+  fi
 }
 
 #######
--- a/asmcli/components/control-plane/managed.sh
+++ b/asmcli/components/control-plane/managed.sh
@@ -26,7 +26,7 @@ install_managed_control_plane() {
   fi
 
   info "Provisioning control plane..."
-  retry 2 run_command curl --request POST \
+  retry 2 check_curl --request POST \
     "https://meshconfig.googleapis.com/v1alpha1/projects/${PROJECT_ID}/locations/${CLUSTER_LOCATION}/clusters/${CLUSTER_NAME}:runIstiod" \
     --data "${POST_DATA}" \
     --header "X-Server-Timeout: 600" \


asmcli/lib/util.sh
```diff
@@ -42,6 +42,28 @@ run_command() {
   return $RETVAL
 }
 
+######
+# check_curl calls curl and returns non-zero if the http code is not 200 or the
+# curl command fails.
+######
+check_curl() {
+  local TMPFILE; TMPFILE=$(mktemp)
```
zerobfd, line 50: Before we start the process we `cd` into a working directory, it's probably useful to make this a named file and leave it around instead of `mktemp` and `rm`.

asmcli/lib/util.sh
```diff
@@ -42,6 +42,28 @@ run_command() {
...
+  local HTTPCODE; HTTPCODE=$(run_command curl --write-out '%{http_code}' --silent --show-error --output "$TMPFILE" "${@}")
+  local RETVAL; RETVAL="$?"
```
zerobfd, line 52: If you do it like this then `$?` is going to have the return code from `local RETVAL` and not `curl`. I probably just checking if $HTTPCODE is empty might be what you want.

asmcli/lib/util.sh
```diff
@@ -42,6 +42,28 @@ run_command() {
...
+    return "$RETVAL"
+  fi
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    echo "HTTP response code: ${HTTPCODE}"
+  fi
+  cat "$TMPFILE" && rm "$TMPFILE"
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    return 1
+  fi
+  return 0
```
zerobfd, line 64: No need to return at the end, by default it will return the return code of the last command run.

asmcli/lib/util.sh
```diff
@@ -42,6 +42,28 @@ run_command() {
...
+  if [[ "$RETVAL" != "0" ]] ; then
+    rm "$TMPFILE"
+    return "$RETVAL"
+  fi
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    echo "HTTP response code: ${HTTPCODE}"
+  fi
+  cat "$TMPFILE" && rm "$TMPFILE"
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    return 1
```
zerobfd, line 62: Returning a specific error code implies that there's some meaning other than just "it didn't work". The canonical(?) way is to do `false; return`, which doesn't imply anything other than "not good".

zerobfd (changes requested):
Approach is good, a few style issues and I think we should keep the log file around. I'm not sure what `curl` will do if the output file already exists since customers might reuse an output directory, so either verify that it will overwrite the file or use a timestamp or something as part of the log name.

asmcli/lib/util.sh
samertm: Done

asmcli/lib/util.sh
samertm: Done

asmcli/lib/util.sh
samertm: You want to check both because if curl fails, HTTPCODE will be "000", which is a bit confusing, so it's probably better to just pass along the curl error code instead. Moved the `local` calls to above the curl command.

asmcli/lib/util.sh
samertm: So if we want to capture the output, the caller of check_curl could also just assign the output to a variable and write it to a file. I think using `check_curl` as a general utility function might be trickier if you need to assign a file for output every time you use it.

samertm:
> I think we should keep the log file around.

So runIstiod outputs `{}` on success, which isn't super useful, and on failures it'll write the error response to stdout (which should maybe go to stderr instead?). Is that useful to keep around? Seems like it would be more confusing than anything, especially since the runIstiod call is eventually going to go away in favor of AFC.

> I'm not sure what curl will do if the output file already exists since customers might reuse an output directory, so either verify that it will overwrite the file or use a timestamp or something as part of the log name.

`tempfile` should always create a uniquely named temporary file in the user's temp directory, which should sidestep that issue, right?

asmcli/lib/util.sh
```diff
@@ -42,6 +42,31 @@ run_command() {
...
+  local HTTPCODE;
+  local RETVAL;
+  HTTPCODE=$(run_command curl --write-out '%{http_code}' --silent --show-error --output "$TMPFILE" "${@}")
+  RETVAL="$?"
+  if [[ "$RETVAL" != "0" ]] ; then
+    rm "$TMPFILE"
+    return "$RETVAL"
+  fi
+  if [[ "$HTTPCODE" != "200" ]] ; then
+    echo "HTTP response code: ${HTTPCODE}"
```
samertm, line 61: Should I use `warn` here, and send the output of $TMPFILE to `warn` as well somehow?

zerobfd: For the response code yes. Although now that I'm thinking about it, if it's not 200 do we want to continue the installation? Maybe we should just `fatal` here and call the function `curl_or_fatal` or something?

zerobfd:
The majority of times the working folder is already in `/tmp`, so I'm not too worried about keeping files around unnecessarily. We assume that they'll get cleaned up by the OS if the customer doesn't want to keep them around. {} as output isn't super helpful, but capturing the error *is* super useful. On success we won't print anything, so there's a useless 2 byte file silently created in what's probably `tmpfs`, which I think is fine. On failure, having the log available makes debugging *a lot* easier, and also allows customers to attach it to support tickets etc.

asmcli/lib/util.sh
samertm: > For the response code yes.
Fixed.

> Although now that I'm thinking about it, if it's not 200 do we want to continue the installation? Maybe we should just fatal here and call the function curl_or_fatal or something?

If all the runIstiod calls fail (since it's called by retry), we will fail the installation. Re: changing this to `curl_or_fatal`, that would make this function harder to use with `retry`. The caller will have enough info to decide to exit the script if that's desired.

zerobfd (approved):

PR #988, opened 2021-09-30
Use Cluster Local Project P4SA with Fleet WIP

jtrbs:
This PR implements

- Set PROJECT_NUMBER within GCP_METADATA with cluster local project gcloud.project.projectNumber. This implies the mesh use cluster local project P4SA. 
- Initialize cluster local project with both Fleet WIP and GKE WIP. This will allow Fleet WI impersonate Cluster Local Project P4SA.

### DIFF
--- a/asm/Kptfile
+++ b/asm/Kptfile
@@ -173,7 +173,7 @@ openAPI:
           - marker: PROJECT_ID
             ref: '#/definitions/io.k8s.cli.setters.gcloud.core.project'
           - marker: PROJECT_NUMBER
-            ref: '#/definitions/io.k8s.cli.setters.gcloud.project.environProjectNumber'
+            ref: '#/definitions/io.k8s.cli.setters.gcloud.project.projectNumber'
           - marker: asm-cluster
             ref: '#/definitions/io.k8s.cli.setters.gcloud.container.cluster'
           - marker: us-central1-c
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -1879,11 +1879,13 @@ init_meshconfig() {
     info "Cluster has Membership ID ${HUB_MEMBERSHIP_ID} in the Hub of project ${FLEET_ID}"
     # initialize replaces the existing Workload Identity Pools in the IAM binding, so we need to support both Hub and GKE Workload Identity Pools
     local POST_DATA
-    POST_DATA='{"workloadIdentityPools":["'${PROJECT_ID}'.hub.id.goog","'${PROJECT_ID}'.svc.id.goog"]}'
-    init_meshconfig_curl "${POST_DATA}" "${PROJECT_ID}"
+    # Initialize the Hub Hosting project with Hub WIP
+    POST_DATA='{"workloadIdentityPools":["'${FLEET_ID}'.hub.id.goog","'${FLEET_ID}'.svc.id.goog"]}'
+    init_meshconfig_curl "${POST_DATA}" "${FLEET_ID}"
     if [[ "${FLEET_ID}" != "${PROJECT_ID}" ]]; then
-      POST_DATA='{"workloadIdentityPools":["'${FLEET_ID}'.hub.id.goog","'${FLEET_ID}'.svc.id.goog"]}'
-      init_meshconfig_curl "${POST_DATA}" "${FLEET_ID}"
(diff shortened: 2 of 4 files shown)


jtrbs:
Hi @jasonwzm could you help check if VM would be impacted by the kpt change? I think in the case of VM installation, environProjectNumber is the same as the cluster local project number. Just to double check if this is the case.

asm/Kptfile
```diff
@@ -173,7 +173,7 @@ openAPI:
           - marker: PROJECT_ID
             ref: '#/definitions/io.k8s.cli.setters.gcloud.core.project'
           - marker: PROJECT_NUMBER
-            ref: '#/definitions/io.k8s.cli.setters.gcloud.project.environProjectNumber'
+            ref: '#/definitions/io.k8s.cli.setters.gcloud.project.projectNumber'
```
zerobfd, line 176: It doesn't look like you actually use this, do you?

zerobfd (approved):
LGTM as long as tests pass. Please revert the Kptfile change if it's not necessary, or add a comment explaining what it does if it is necessary.

asm/Kptfile
jtrbs: You mean within the asmcli kpt commands? gcloud.project.projectNumber is set automatically after the package is downloaded (not quite clear about the logic behind it, it is similar to gcloud.core.project).

jtrbs: Let me update this PR to explicitly set this field.

jtrbs: Hi @zerobfd, I retested with the current PR. According to the following log, ``` 2021-09-30T22:46:14.170951 asmcli: Running: '/usr/local/google/home/tairan/go/src/asm/asm-helper/test/gcp/fleet/tmp1/kpt cfg set asm gcloud.core.project test-multiproj1'
2021-09-30T22:46:14.239829 asmcli: -------------
asm/
set 19 field(s) of setter "gcloud.core.project" to value "test-multiproj1"
asm/
set 2 field(s) of setter "gcloud.project.projectNumber" to value "416604968218"
```, looks like when `kpt cfg set asm gcloud.core.project test-multiproj1` is executed, `kpt cfg set asm gcloud.project.projectNumber` is executed automatically. Thus I do not need to call it explicitly one more time.

Comment (zerobfd) [reply asm/Kptfile:176]: Huh, I learned something today. Thanks for looking into it!

**Review by zerobfd — APPROVED** [review]

**Review by jasonwzm — APPROVED** [review]
VM is using fleet WIP, so it does not affect VM.

File: asmcli/asmcli [hunk 2]
```diff
@@ -1990,11 +1992,13 @@ init_meshconfig_managed() {
 
   info "Initializing meshconfig managed API..."
   local POST_DATA
-  POST_DATA='{"workloadIdentityPools":["'${PROJECT_ID}'.hub.id.goog","'${PROJECT_ID}'.svc.id.goog"], "prepare_istiod": true}'
-  init_meshconfig_curl "${POST_DATA}" "${PROJECT_ID}"
+  # Initialize the Hub Hosting project with Hub WIP
+  POST_DATA='{"workloadIdentityPools":["'${FLEET_ID}'.hub.id.goog","'${FLEET_ID}'.svc.id.goog"], "prepare_istiod": true}'
```
Aakash2017, line 1996: I think this is actually switched (we don't need to prepare istiod's for the fleet project, but we do for the cluster project).

asmcli/components/control-plane/managed.sh
```diff
@@ -99,10 +99,12 @@ init_meshconfig_managed() {
 
   info "Initializing meshconfig managed API..."
   local POST_DATA
-  POST_DATA='{"workloadIdentityPools":["'${PROJECT_ID}'.hub.id.goog","'${PROJECT_ID}'.svc.id.goog"], "prepare_istiod": true}'
-  init_meshconfig_curl "${POST_DATA}" "${PROJECT_ID}"
+  # Initialize the Hub Hosting project with Hub WIP
+  POST_DATA='{"workloadIdentityPools":["'${FLEET_ID}'.hub.id.goog","'${FLEET_ID}'.svc.id.goog"], "prepare_istiod": true}'
```
Aakash2017, line 103: same comment as above, prepare_istiod should be set at cluster project

Aakash2017:
LGTM

PR #990, opened 2021-10-01
Main 1112

zerobfd:


ZhengzheYang (approved):
Do we update Kptfile as well?

zerobfd:
Yes, you're just faster than me reacting to the linter :P

PR #1001, opened 2021-10-06
Always populate cluster values

zerobfd:


asmcli/lib/context.sh
```diff
@@ -5,6 +5,7 @@ context_init() {
     "PROJECT_ID": "${PROJECT_ID:-}",
     "CLUSTER_NAME": "${CLUSTER_NAME:-}",
     "CLUSTER_LOCATION": "${CLUSTER_LOCATION:-}",
+    "GKE_CLUSTER_URI": "${GKE_CLUSTER_URI:-}",
```
ZhengzheYang, line 8: do we also need `NETWORK_ID` here?

ZhengzheYang (approved):

PR #1009, opened 2021-10-07
Use absolute paths for kubeconfig files in create-mesh

zerobfd:


zerobfd:
This was tested manually and fixes the reported issue. (And prints the error message properly on failure to find the file.)

PR #1016, opened 2021-10-13
Fix retry logic for runIstiod

zerobfd:


samertm:
That was fast haha, can you fix the `initialize` call too?

PR #1017, opened 2021-10-19
Add VPCSC support for MCP in asmcli experiment

ruigulala:
An user input argument is required to install MCP in a
VPSC restricted environment.


zerobfd (changes requested):
Two things, overall looks fine.

1) `enable` has a specific meaning for the other flags for things that `asmcli` can enable on behalf of the user. This just specifies whether the mesh is on VPC-SC or not. Can we change `enable_vpcsc` to `use_vpcsc`?

2) Can you also add this flag to the usage printed with the `--help` flag?

zerobfd (approved):

PR #1018, opened 2021-10-20
Respect --network_id for non-GCP platforms

Monkeyanator:
Before, when not using a non-GCP platform, we'd ignore what's passed in for --network_id and use "default". After this we'll use what's set in the flag and leave unset otherwise.


zerobfd (changes requested):
There seems to be an extra file that made it into the repo and got included in the generated `asmcli` output, can you take a look?

PR #1024, opened 2021-10-21
Fix USE_VPCSC value check

ruigulala:


asmcli/commands/experimental/install.sh
```diff
@@ -50,7 +50,7 @@ x_configure_package() {
   if [[ "${USE_MANAGED_CNI}" -eq 1 ]]; then
     kpt cfg set asm anthos.servicemesh.use-managed-cni "true"
   fi
-  if [[ -n "${USE_VPCSC}" ]]; then
+  if [[ -n "${USE_VPCSC}" -eq 1 ]]; then
```
zerobfd, line 53: You can either do `-n $VAR` or `$VAR -eq 1` but not both.

PR #1033, opened 2021-11-02
use v1beta1 for CPR CR

yanyuan-meiyi:


zerobfd (approved):

ZhengzheYang (approved):

asm/control-plane-revision/crd.yaml
```diff
@@ -15,88 +15,150 @@ spec:
...
+                      type: string
+                    type:
+                      description: Type is the type of the condition.
+                      type: string
+                  type: object
+                type: array
+            type: object
+        type: object
+    served: true
+    storage: true
```
djazayeri, line 87: Just to verify, this is consistent with what AFC installs?

djazayeri (approved):

asm/control-plane-revision/crd.yaml
yanyuan-meiyi: Yes. This is copied from the file generated using the blaze command.

PR #1034, opened 2021-11-03
Adding Kpt to samples

shankgan:


zerobfd:
I'd rather not introduce more kpt if possible. What problem is this solving? Can we use, say, Helm?

shankgan:
> I'd rather not introduce more kpt if possible. What problem is this solving? Can we use, say, Helm?

uggh...just got done configuring everything with kpt 0.39.2. Can achieve the same with helm if needed.

Goal is to easily configure custom revision based ingress-gateways - especially needed when migrating workloads between control plane revisions with zero downtime

howardjohn:
I recommend doing none of the above.

These are _samples_. We want users to be customizing up to the entire spec of Service + Pod - over 1000 different API settings are available. They should check these into their repos and customize how they like (helm ,kpt, kustomize, directly modifying - we are not opinionated).

If they do want a helm chart instead they can use https://artifacthub.io/packages/helm/istio-official/gateway (OSS gateway chart). It is almost identical to the sample.

shankgan:
> I recommend doing none of the above.
> 
> These are _samples_. We want users to be customizing up to the entire spec of Service + Pod - over 1000 different API settings are available. They should check these into their repos and customize how they like (helm ,kpt, kustomize, directly modifying - we are not opinionated).
> 
> If they do want a helm chart instead they can use https://artifacthub.io/packages/helm/istio-official/gateway (OSS gateway chart). It is almost identical to the sample.

Ack. Closing this PR

PR #1035, opened 2021-11-03
Update migrate-addon script

howardjohn:
Internal bug: 204365008

### DIFF
--- a/scripts/migration/migrate-addon
+++ b/scripts/migration/migrate-addon
@@ -31,6 +31,7 @@ set -o pipefail
 # can be cleaned up using the cleanup() routine
 
 TMP_DIR="$(mktemp -d)"
+
 ISTIOCTL_BIN="istioctl"
 SKIP_CONFIRM=""
 CONTEXT=""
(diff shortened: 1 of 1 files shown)


zerobfd:
Can you do the PR against main instead?

howardjohn:
> Can you do the PR against main instead?

the script is only in `master` for whatever reason

zerobfd:
You're right, I just can't read.

scripts/migration/migrate-addon
```diff
@@ -203,7 +241,90 @@ EOF
...
+                sdsConfig:
+                  initialFetchTimeout: 1s
+                  apiConfigSource:
+                    apiType: GRPC
+                    grpcServices:
+                    - envoyGrpc:
+                        clusterName: sds-grpc
+EOF
+  echo "Waiting for proxies to pick up the new root certificate..."; sleep 15
+  kube delete envoyfilter trigger-root-cert -n istio-system
```
zerobfd, line 313: I guess there's not an easy way to actually check?

zerobfd (approved):
LGTM.

scripts/migration/migrate-addon
howardjohn: Unfortunately not really. We can go exec into every pod but... not ideal

PR #1038, opened 2021-11-10
Use proper configmap name for trust domain logic

zerobfd:


asmcli/commands/create-mesh.sh
```diff
@@ -256,7 +261,7 @@ EOF
 has_fleet_alias() {
   local FLEET_ID; FLEET_ID="${1}"
   local REVISION; REVISION="${2}"
-  local RAW_TRUST_DOMAIN_ALIASES; RAW_TRUST_DOMAIN_ALIASES="$(retry 2 kubectl -n istio-system get configmap istio-"${REVISION}" \
+  local RAW_TRUST_DOMAIN_ALIASES; RAW_TRUST_DOMAIN_ALIASES="$(retry 2 kubectl -n istio-system get configmap "${REVISION}" \
```
ZhengzheYang, line 264: why we are using `$REVISION` here when we patch the cm with `$REV_NAME` in line 250?

asmcli/components/ca/mesh-ca.sh
```diff
@@ -45,6 +45,10 @@ EOF
   if [[ "$ISTIOD_COUNT" -ne 0 ]]; then
     local REVISION; REVISION="$(retry 2 kubectl -n istio-system get pod -l app=istiod \
       -o jsonpath='{.items[].spec.containers[].env[?(@.name=="REVISION")].value}')"
+    local REV_NAME; REV_NAME="istio-${REVISION}"
+    if [[ "${REVISION}" = default ]]; then
+      REV_NAME="istio"
+    fi
     local RAW_TRUST_DOMAINS_ALIASES; RAW_TRUST_DOMAINS_ALIASES="$(retry 2 kubectl -n istio-system get configmap istio-"${REVISION}" \
```
ZhengzheYang, line 52: do we need to change this to use `kubectl -n istio-system get configmap "${REV_NAME}"`?

howardjohn (approved):

asmcli/commands/create-mesh.sh
zerobfd: I just didn't think changing the variable name was super useful since we're passing it in, but you're right, I should probably be consistent.

asmcli/components/ca/mesh-ca.sh
zerobfd: Yes.

PR #1040, opened 2021-11-12
create cross-project mesh service account

davidhauck:
This will do two things:
1) Use the --enable-gcp-iam-roles to automatically create cross project IAM bindings if needed
2) Create the mesh service account binding cross project if needed, also respecting the --enable-gcp-iam-roles flag

### DIFF
--- a/asmcli/lib/util.sh
+++ b/asmcli/lib/util.sh
@@ -754,52 +754,80 @@ get_cr_yaml() {
   echo "${CR} ${REVISION}"
 }
 
-ensure_cross_project_fleet_sa() {
+ensure_cross_project_service_accounts() {
   local FLEET_ID; FLEET_ID="${1}"
   local PROJECT_ID; PROJECT_ID="${2}"
 
   if ! is_gcp; then return; fi
 
+  local FLEET_HOST_PROJECT_NUMBER
+  FLEET_HOST_PROJECT_NUMBER="$(gcloud projects describe "${FLEET_ID}" --format "value(projectNumber)")"
+
+  local FLEET_SA
+  local MESH_SA
+  FLEET_SA="serviceAccount:service-${FLEET_HOST_PROJECT_NUMBER}@gcp-sa-gkehub.iam.gserviceaccount.com"
+  MESH_SA="serviceAccount:service-${FLEET_HOST_PROJECT_NUMBER}@gcp-sa-servicemesh.iam.gserviceaccount.com"
+
+  if ! ensure_cross_project_sa "${FLEET_ID}" "${PROJECT_ID}" "${FLEET_SA}" "roles/gkehub.serviceAgent"; then
+    warn "The Fleet service account may not have been created for the Fleet hosted in ${FLEET_ID}."
+    warn "Please refer to https://cloud.google.com/anthos/multicluster-management/connect/prerequisites#gke-cross-project"
+    warn "for information on how to create the identity and grant permissions. You may also re-run this command"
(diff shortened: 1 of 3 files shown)


asmcli/lib/util.sh
```diff
@@ -754,52 +754,89 @@ get_cr_yaml() {
...
   if [[ -z "${PROJECT_POLICY_MEMBERS}" ]]; then
-    warn "Unable to verify cross-project Fleet permissions. Installation will continue."
-    warn "Please refer to https://cloud.google.com/anthos/multicluster-management/connect/prerequisites#gke-cross-project"
-    warn "if cross-project traffic doesn't work."
+    false
     return
   fi
 
-  PROJECT_POLICY_MEMBERS="$(echo "${PROJECT_POLICY_MEMBERS}" | jq '.bindings[] | select(.role == "roles/gkehub.serviceAgent")')"
+  PROJECT_POLICY_MEMBERS="$(echo "${PROJECT_POLICY_MEMBERS}" | jq ".bindings[] | select(.role == \"${ROLE}\")")"
```
zerobfd, line 815: This should still use single quotes. Do

```
'.bindings[] | select(.role == "'"${ROLE}"'")'
```

asmcli/lib/util.sh
```diff
@@ -754,52 +754,89 @@ get_cr_yaml() {
...
 
   if [[ -n "${PROJECT_POLICY_MEMBERS}" ]]; then
     PROJECT_POLICY_MEMBERS="$(echo "${PROJECT_POLICY_MEMBERS}" | jq '.members[]')"
   fi
 
-  if [[ "${PROJECT_POLICY_MEMBERS}" = *"${FLEET_SA}"* ]]; then return; fi
+  if [[ "${PROJECT_POLICY_MEMBERS}" == *"${SA_NAME}"* ]]; then
+    true
+    return
```
zerobfd, line 823: just return is sufficient

zerobfd (changes requested):
Why are we also changing the warnings to fatals here?

asmcli/lib/util.sh
davidhauck: Done

asmcli/lib/util.sh
davidhauck: Done

davidhauck:
I'm changing them to fatal since there was a case where we wanted to fail. Should I keep the exact same logic and fail only if the service account is not in the policies? Or should we just warn for all cases?

zerobfd:
We can't fail because in multi-project scenarios, the person installing ASM might not have permission to even *view* the policies in the other project. However, a security admin could have already created the required bindings.

davidhauck:
Got it. Updated to warn instead

asmcli/lib/util.sh
```diff
@@ -754,52 +754,88 @@ get_cr_yaml() {
...
-  fatal "for information on how to grant permissions."
+  if [[ "${ENABLE_ALL}" -eq 1 || "${ENABLE_GCP_IAM_ROLES}" -eq 1 ]]; then
+    create_sa_binding "${PROJECT_ID}" "${SA_NAME}" "${ROLE}"
+    true
+    return
+  fi
+  false
 }
+
+create_sa_binding() {
```
zerobfd, line 834: There's already a "bind_user_to_iam_policy" function, re-use that one.

zerobfd (approved):
LGTM after comment resolved.

asmcli/lib/util.sh
davidhauck: Done.

PR #1045, opened 2021-11-17
move afc install to main command

davidhauck:


asmcli/components/control-plane/managed.sh
```diff
@@ -24,49 +24,63 @@ call_runIstiod() {
...
+  else
+    # `kubectl wait` for non-existent resources will return error directly so we wait in loop
+    info "Wait for the controlplanerevisions CRD to be installed by AFC. This should take a few minutes if cluster is newly registered."
+    for i in {1..10}; do
+      if kubectl wait --for condition=established --timeout=10s crd/controlplanerevisions.mesh.cloud.google.com 2>/dev/null; then
+        break
+      fi
+      sleep 10
+    done
+  fi
```
zerobfd, line 79: Can you split the if/else logic into separate functions? e.g. `if is_legacy; then install_legacy_mcp; else install_mcp; fi`

asmcli/commands/help.sh
```diff
@@ -175,6 +175,8 @@ FLAGS:
 
      --managed                        Provision a remote, managed control plane
                                       instead of installing one in-cluster.
+     --legacy                         Provision a remote, managed control plane
+                                      with the legacy asmcli installer.
```
zerobfd, line 179: Can you add "that runs client-side" somewhere in the description?

asmcli/lib/parse.sh
```diff
@@ -284,6 +288,10 @@ x_parse_install_args() {
         context_set-option "MANAGED" 1
         shift 1
         ;;
+      --legacy)
+        context_set-option "LEGACY" 1
+        shift 1
+        ;;
```
zerobfd, line 294: I don't think we need to enable client-side installs for `x install`. In fact we might want to just print a warning that `x install` is included in `--managed` now.

asmcli/lib/util.sh
```diff
@@ -723,27 +723,27 @@ get_cr_channels() {
 
 get_cr_yaml() {
   local CHANNEL; CHANNEL="${1}"
-  local EXPERIMENTAL; EXPERIMENTAL="$(context_get-option "EXPERIMENTAL")"
+  local LEGACY; LEGACY="$(context_get-option "LEGACY")"
   local CR REVISION
   case "${CHANNEL}" in
     regular)
-      if [[ "${EXPERIMENTAL}" -eq 1 ]]; then
+      if [[ "${LEGACY}" -eq 0 ]]; then
```
zerobfd, line 730: You made the `is_legacy` function so use that here instead of checking the variable explicitly.

asmcli/commands/help.sh
```diff
@@ -264,6 +266,7 @@ FLAGS:
      --use_managed_cni
 
      --managed
+     --legacy
```
zerobfd, line 269: We should also make sure that if `legacy` is passed without `managed` we fail installation and let the user know that it only makes sense in the managed context.

zerobfd (changes requested):

asmcli/commands/help.sh
davidhauck: Done.

asmcli/commands/help.sh
davidhauck: Done.

asmcli/components/control-plane/managed.sh
davidhauck: Done.

asmcli/lib/parse.sh
davidhauck: Done.

asmcli/lib/util.sh
davidhauck: Done.

asmcli/commands/help.sh
```diff
@@ -175,6 +175,8 @@ FLAGS:
 
      --managed                        Provision a remote, managed control plane
                                       instead of installing one in-cluster.
+     --legacy                         Provision a remote, managed control plane with
+                                      the legacy asmcli installer the runs client-side.
```
zerobfd, line 179: s/the runs/that runs

asmcli/components/control-plane/managed.sh
```diff
@@ -65,14 +84,17 @@ install_managed_control_plane() {
...
-  install_mananged_cni_static
-
-  if [[ "${CA}" = "gcp_cas" ]]; then
-    install_managed_privateca
-  fi
-
-  install_managed_startup_config
+wait_for_cpr_crd() {
+  # `kubectl wait` for non-existent resources will return error directly so we wait in loop
+  info "Wait for the controlplanerevisions CRD to be installed by AFC. This should take a few minutes if cluster is newly registered."
```
zerobfd, line 91: s/Wait/Waiting
s/should/could

zerobfd (approved):
Some wording nits, otherwise if it works and tests pass then LGTM.

asmcli/commands/help.sh
davidhauck: Done.

asmcli/components/control-plane/managed.sh
davidhauck: Done.

zerobfd:
You'll also need to update the test for experimental managed install to instead test the legacy one.

PR #1048, opened 2021-11-17
Add CI variables to override istio folder

howardjohn:
Will send to install_asm once validated I didn't do things horribly wrong


zerobfd:
Fixed in #1049

PR #1057, opened 2021-11-22
Use only mesh.googleapis.com API instead of explicit list

djazayeri:


djazayeri:
I ran the presubmit command as specified in the failed cloud build, but this undid most of my changes...

zerobfd (approved):

PR #1060, opened 2021-11-24
Add created-by label to CPRs

jasonwzm:


asmcli/commands/help.sh
```diff
@@ -36,7 +36,7 @@ help_subcommand() {
 }
 
 version_message() {
-  local VER; VER="${MAJOR}.${MINOR}.${POINT}-asm.${REV}+config${CONFIG_VER}";
```
zerobfd, line 39: Changing it here is going to change it everywhere which will break a ton of things. Set a variable to $(version_message) before the kpt setter and then do `var="asmcli-${mess_var/+/.}"`` or something similar

zerobfd (changes requested):

asmcli/commands/help.sh
jasonwzm: Done.

zerobfd (approved):
If you tested it locally and it works then LGTM

jasonwzm:
Manually tested. Label is added with expected value.

PR #1063, opened 2021-11-30
fix create-mesh cluster names

stevenctl:
The `--name` flag should match the cluster name used for install.


zerobfd:
Unit test is failing, but the change makes sense. I'll LGTM it once it works.

zerobfd (approved):

PR #1079, opened 2021-12-15
Bump 1.12 revision

sergii-ssh:


zerobfd:
Can you make the change to the main branch and then we can cherry-pick it to the staging/release branches?

sergii-ssh:
right, 1.12 is the current release.. Will do

PR #1083, opened 2021-12-16
Update crd.yaml

chengsiy:
Show status in extra column printers


zerobfd:
I merged this manually from main.

PR #1084, opened 2021-12-17
Rename distroless-proxy to distroless-proxy.yaml

mathieu-benoit:
Sorry, I missed the file extension


zerobfd:
The doc blocks are also gone, was that intentional?

mathieu-benoit:
Yep, I think they are not mandatory, right?
There is couple examples with or without, this one https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asm/istio/options is without for example. Let me know if you see any issue with that, I will put them back. Thanks.

zerobfd:
No issue, just wasn't in the title/description of the PR so I wanted to double check.

zerobfd (approved):

PR #1097, opened 2022-01-13
sync migrate-addon script

richardwxn:


richardwxn:
@howardjohn

zerobfd:
Can we put this in main instead of master? Nothing uses this branch anymore.

PR #1100, opened 2022-01-18
Patch 1.12.2

sergii-ssh:


jasonwzm:
I believe we update the `main` branch first and cherrypick to `staging-1.12`.

PR #1111, opened 2022-02-02
Unset which

zerobfd:
RHEL for whatever reason now includes this by default and it breaks a lot of stuff, including asmcli.


ZhengzheYang (approved):
this only unsets the function definitions? Neat.

PR #1114, opened 2022-02-05
add certificate template support for google-cas installation

liwenhao0810:
Add certificate template support by appending the CT to the --ca_pool param.

### DIFF
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -1241,7 +1241,11 @@ OPTIONS:
                                       Name of the ca pool in the GCP CAS service used to
                                       sign certificates in the format
                                       'projects/project_name/locations/ \
-                                      ca_region/caPools/ca_pool'.
+                                      ca_region/caPools/ca_pool'. If certificate template
+                                      needs to be specified, the format is 'projects/ \
+                                      project_name/locations/ca_region/caPools/ \
+                                      ca_pool:projects/project_name/locations/ca_region/ \
+                                      certificateTemplates/cert_template'.
   -r|--revision_name <REVISION NAME>  Custom revision label. Label needs to follow DNS
                                       label formats (re: RFC 1123). Not supported if
                                       control plane is managed. Prefixing the revision
@@ -1597,11 +1601,12 @@ install_private_ca() {
   # If modify_gcp_component permissions are not granted, it is assumed that the
   # user has taken care of this, else Istio setup will fail
   local CA_NAME; CA_NAME="$(context_get-option "CA_NAME")"
+  local CA_POOL_URL; CA_POOL_URL=$(echo "${CA_NAME}" | cut -f1 -d:)
   local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
   local WORKLOAD_IDENTITY; WORKLOAD_IDENTITY="${FLEET_ID}.svc.id.goog:/allAuthenticatedUsers/"
-  local CA_LOCATION; CA_LOCATION=$(echo "${CA_NAME}" | cut -f4 -d/)
-  local CA_POOL; CA_POOL=$(echo "${CA_NAME}" | cut -f6 -d/)
-  local PROJECT; PROJECT=$(echo "${CA_NAME}" | cut -f2 -d/)
+  local CA_LOCATION; CA_LOCATION=$(echo "${CA_POOL_URL}" | cut -f4 -d/)
+  local CA_POOL; CA_POOL=$(echo "${CA_POOL_URL}" | cut -f6 -d/)
+  local PROJECT; PROJECT=$(echo "${CA_POOL_URL}" | cut -f2 -d/)
 
   # TODO : reduce retry and remove logging
   retry 10 gcloud privateca pools add-iam-policy-binding "${CA_POOL}" \
@@ -1620,6 +1625,15 @@ install_private_ca() {
     --role "roles/privateca.auditor" \
     --log-http \
     --verbosity debug
+
+  if [[ 1 == $(grep -o ":" <<< "${CA_NAME}" | wc -l) ]]; then
+    local CERT_TEMPLATE; CERT_TEMPLATE=$(echo "${CA_NAME}" | cut -f2 -d:)
+    retry 3 gcloud privateca templates add-iam-policy-binding "${CERT_TEMPLATE}" \
+      --member "group:${WORKLOAD_IDENTITY}" \
+      --role "roles/privateca.templateUser" \
+      --log-http \
+      --verbose
+  fi
 }
 
 does_istiod_exist(){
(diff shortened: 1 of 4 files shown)


zerobfd:
The code directionally looks good, but we don't edit the `asmcli` file directly. In the /asmcli directory there are several files that get merged to create the final tool. See https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/docs/developer-guide.md for more details on how to work with the repo.

asmcli/asmcli
```diff
@@ -1241,7 +1241,11 @@ OPTIONS:
                                       Name of the ca pool in the GCP CAS service used to
                                       sign certificates in the format
                                       'projects/project_name/locations/ \
-                                      ca_region/caPools/ca_pool'.
+                                      ca_region/caPools/ca_pool'. If specify certificate
```
shankgan, line 1244: nit: "if certificate template needs to be specified"

asmcli/asmcli
```diff
@@ -1614,6 +1619,13 @@ install_private_ca() {
     --location "${CA_LOCATION}" \
     --member "group:${WORKLOAD_IDENTITY}" \
     --role "roles/privateca.auditor"
+
+  if [[ 1 == $(grep -o ":" <<< "${CA_NAME}" | wc -l) ]]; then
+    local CERT_TEMPLATE; CERT_TEMPLATE=$(echo "${CA_NAME}" | cut -f2 -d:)
```
shankgan, line 1624: @pmansour - do you for-see any issues with using ":" as a delimiter to specify both the ca-pool and the certificate template in the same variable?

asmcli/asmcli
```diff
@@ -1597,11 +1601,12 @@ install_private_ca() {
   # If modify_gcp_component permissions are not granted, it is assumed that the
   # user has taken care of this, else Istio setup will fail
   local CA_NAME; CA_NAME="$(context_get-option "CA_NAME")"
+  local CA_POOL_URL; CA_POOL_URL=$(echo "${CA_NAME}" | cut -f1 -d:)
   local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
```
shankgan, line 1605: nit: maybe add an additional check to see if there is more than one ":' in the string, this errors out as an unacceptable format.

asmcli/asmcli
```diff
@@ -2060,16 +2072,28 @@ EOF
...
-  elif [[ "$(echo "${CA_NAME}" | cut -f1 -d/)" != "$(echo "${CA_NAME_TEMPLATE}" | cut -f1 -d/)" ]]; then
-    fatal "Malformed ca-name. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+  # check if CA_NAME is ca_pool:cert_template format
+  elif [[ $(grep -o ":" <<< "${CA_NAME}" | wc -l) == $(grep -o ":" <<< "${CA_NAME_TEMPLATE}" | wc -l) ]]; then
+    if [[ $(grep -o "/" <<< "${CA_NAME}" | wc -l) != $(grep -o "/" <<< "${CA_NAME_TEMPLATE}" | wc -l) ]]; then
+      fatal "Malformed ca-name with certificate template. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+    elif [[ "$(echo "${CA_NAME}" | cut -f1 -d/)" != "$(echo "${CA_NAME_TEMPLATE}" | cut -f1 -d/)" ]]; then
+      fatal "Malformed ca-name with certificate template. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+    fi
+  else
```
lei-tang, line 2091: nit: to be consistent with the comment above "# check if CA_NAME is ca_pool:cert_template format", consider adding a comment here.

asmcli/asmcli
liwenhao0810: Added in validate_private_ca().

asmcli/components/ca/private-ca.sh
```diff
@@ -1,13 +1,27 @@
...
+  local CA_NAME_TEMPLATE; CA_NAME_TEMPLATE="${CA_POOL_TEMPLATE}:${CT_TEMPLATE}"
 
   if [[ -z ${CA_NAME} ]]; then
     fatal "A ca-name must be provided for integration with Google Certificate Authority Service."
-  elif [[ $(grep -o "/" <<< "${CA_NAME}" | wc -l) != $(grep -o "/" <<< "${CA_NAME_TEMPLATE}" | wc -l) ]]; then
-    fatal "Malformed ca-name. ca-name must be of the form ${CA_NAME_TEMPLATE}."
-  elif [[ "$(echo "${CA_NAME}" | cut -f1 -d/)" != "$(echo "${CA_NAME_TEMPLATE}" | cut -f1 -d/)" ]]; then
-    fatal "Malformed ca-name. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+  # check if CA_NAME is ca_pool:cert_template format
+  elif [[ 0 < $(grep -o ":" <<< "${CA_NAME}" | wc -l) ]]; then
```
zerobfd, line 10: Failing lint:
```
In asmcli line 2084:
  elif [[ 0 < $(grep -o ":" <<< "${CA_NAME}" | wc -l) ]]; then
            ^-- SC2071: < is for string comparisons. Use -lt instead.
```

liwenhao0810: fixed.

asmcli/components/ca/private-ca.sh
```diff
@@ -1,13 +1,27 @@
...
-    fatal "Malformed ca-name. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+  # check if CA_NAME is ca_pool:cert_template format
+  elif [[ "${CA_NAME}" == *":"* ]]; then
+    if [[ $(grep -o ":" <<< "${CA_NAME}" | wc -l) != $(grep -o ":" <<< "${CA_NAME_TEMPLATE}" | wc -l) ]]; then
+      fatal "Malformed ca-name with certificate template. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+    elif [[ $(grep -o "/" <<< "${CA_NAME}" | wc -l) != $(grep -o "/" <<< "${CA_NAME_TEMPLATE}" | wc -l) ]]; then
+      fatal "Malformed ca-name with certificate template. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+    elif [[ "$(echo "${CA_NAME}" | cut -f1 -d/)" != "$(echo "${CA_NAME_TEMPLATE}" | cut -f1 -d/)" ]]; then
+      fatal "Malformed ca-name with certificate template. ca-name must be of the form ${CA_NAME_TEMPLATE}."
+    fi
```
pmansour, line 17: suggestion: you'd get much simpler bash error checks here and more accurate validations if you make the above template strings regexes, rather than trying to count the number of slashes and colons in each occurrence. For example, despite all the complex / hard-to-read validation logic here, it seems that the string "projects/foo/coffeeShops/CafeVita/espressoMachines/LaMarzocco" would be considered a valid CaPool.

asmcli/commands/install.sh
```diff
@@ -69,11 +69,12 @@ install_private_ca() {
   # If modify_gcp_component permissions are not granted, it is assumed that the
   # user has taken care of this, else Istio setup will fail
   local CA_NAME; CA_NAME="$(context_get-option "CA_NAME")"
+  local CA_POOL_URL; CA_POOL_URL=$(echo "${CA_NAME}" | cut -f1 -d:)
```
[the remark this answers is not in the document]
liwenhao0810: fixed

asmcli/commands/install.sh
```diff
@@ -86,6 +87,13 @@ install_private_ca() {
     --location "${CA_LOCATION}" \
     --member "group:${WORKLOAD_IDENTITY}" \
     --role "roles/privateca.auditor"
+
+  if [[ $(grep -o ":" <<< "${CA_NAME}" | wc -l) -eq 1 ]]; then
```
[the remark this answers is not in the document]
liwenhao0810: fixed

asmcli/components/ca/private-ca.sh
liwenhao0810: fixed using regex check.

asmcli/commands/install.sh
```diff
@@ -86,6 +87,13 @@ install_private_ca() {
...
+    local CERT_TEMPLATE; CERT_TEMPLATE=$(echo "${CA_NAME}" | cut -f2 -d:)
+    retry 3 gcloud privateca templates add-iam-policy-binding "${CERT_TEMPLATE}" \
```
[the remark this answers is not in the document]
liwenhao0810: Not sure how to check whether have the privileges to bind IAM roles, one thought here is that even it fails to bind the roles, it will not fail the installation. So from that perspective this is something user could work around.
But agreed on that is right things to do as following, if you know the command that check whether user have the permission to set the bindings, please let me know.

zerobfd (approved):
LGTM when the current threads are resolved

asmcli/commands/help.sh
```diff
@@ -110,7 +110,11 @@ OPTIONS:
                                       Name of the ca pool in the GCP CAS service used to
                                       sign certificates in the format
                                       'projects/project_name/locations/ \
-                                      ca_region/caPools/ca_pool'.
+                                      ca_region/caPools/ca_pool'. If certificate template
+                                      needs to be specified, the format is 'projects/ \
+                                      project_name/locations/ca_region/caPools/ \
+                                      ca_pool:projects/project_name/locations/ca_region/ \
+                                      certificateTemplates/cert_template'.
```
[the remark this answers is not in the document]
shankgan: @zerobfd - the UX isn't ideal as implemented now; we are combining two different parameters into the same flag. Would it be acceptable to add an extra flag to asmcli (--ca-template)?

asmcli/commands/install.sh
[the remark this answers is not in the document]
shankgan: A good compromise IMO would be to gate ALL the IAM role changes with the "can_modify_gcp_iam_roles" function. I think adding checks here could get complicated. How about we make the required IAM roles more easily available in our documentation?

asmcli/commands/help.sh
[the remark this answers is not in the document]
zerobfd: As long as the UX is decided before we merge anything into staging. Also please keep in mind that `asmcli` is going away as soon as humanly possible as soon as the API installation reaches parity, so any new args/changes is going to introduce more work for CAS customers to use and then migrate off of.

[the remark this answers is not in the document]
liwenhao0810: Agreed with keep as this for now and see if we need to add a flag here or directly in API installation.

asmcli/commands/install.sh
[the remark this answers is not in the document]
liwenhao0810: Added the conditional check "can_modify_gcp_iam_roles".

Issue #1117, opened 2022-02-10
Clarify revision label in outro()


danielmarzini:
We print out the revision label in the outro, but we don't change the revision label based on the type of installation has been launched. For instance, --managed with --channel rapid should return --asm-managed-rapid, asm-managed is return instead.

PR #1118, opened 2022-02-10
Fix issue #1117

danielmarzini:
Fix issue #1117 : We print out the revision label in the outro, but we don't change the revision label based on the type of installation has been launched. For instance, --managed with --channel rapid should return --asm-managed-rapid, asm-managed is return instead.

### DIFF
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -1672,16 +1672,31 @@ expose_istiod() {
 
 outro() {
   local OUTPUT_DIR; OUTPUT_DIR="$(context_get-option "OUTPUT_DIR")"
-  
+  local CHANNEL; CHANNEL="$(context_get-option "CHANNEL")"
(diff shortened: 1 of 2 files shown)


danielmarzini:
@mathieu-benoit could you please review this one?

mathieu-benoit:
Thanks but I don't have this "autority" 😄, but I'm sure @zerobfd will do this review.

zerobfd:
Thanks for the PR! The way that `asmcli` is generated, the file isn't edited directly. Can you make the change to asmcli/commands/install.sh as well? Otherwise the package building won't work.

danielmarzini:
Thank you @zerobfd, I applied the changes to that file. would you please tell me what are the errors in the build? I cannot see them.

costinm (approved):
Note that the plan (AFAIK) is to switch to 'default' revision, so we'll need to adjust this when it gets implemented.

PR #1121, opened 2022-02-11
Add options for telemetry

zerobfd:


jtrbs:
And do not set the "false" when one of the options for telemetry is used? https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asmcli/asmcli#L2869-L2871

zerobfd:
I thought it shouldn't matter since the overlays come afterwards and they don't have kpt setters.

bianpengyuan:
ENABLE_STACKDRIVER_MONITORING only applies to the control plane.

jtrbs:
@bianpengyuan thanks for the confirmation. For stackdriver enablement, shall `spec.values.telemetry.v2.stackdriver.enabled` https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asm/istio/istio-operator.yaml#L87-L88 share the same value as `ENABLE_STACKDRIVER_MONITORING` by default? Currently there is only a setter which set `ENABLE_STACKDRIVER_MONITORING` to false. That is for cases in which Stackdriver does not work (like no egress to Google backend). In those cases, customer would need to set `spec.values.telemetry.v2.stackdriver.enabled` to false as well. That means we need a separate setter for `spec.values.telemetry.v2.stackdriver.enabled`. Could be another improvement.

jtrbs (approved):

bianpengyuan:
> share the same value as ENABLE_STACKDRIVER_MONITORING by default?

Not necessarily. Control plane metrics are also used for other purpose than monitoring and is way cheaper than data plane, so it totally makes sense to me to keep control reporting telemetry to google but use prom for data plane.

jtrbs:
In this case, do you recommend that `spec.values.telemetry.v2.stackdriver.enabled` (data plane) is set to `false` by default for all platforms (currently prom is set to `false` https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asm/istio/istio-operator.yaml#L86) and optionally enabled by one of the overlay in this PR? This would be a backward-incompatible change, especially for GCP. So maybe such change should be done in some ASM new version. Or an explicit`asmcli flag -> kpt setter` for `spec.values.telemetry.v2.stackdriver.enabled` and let customer do whatever they want.

bianpengyuan:
No we should always enable data plane sd telemetry by default. User need to turn it off if they don't care about UI or concern about the cost.

jtrbs:
Got it. So `User need to turn it off if` will also apply when stackdriver is not supported in their platform env. It is user's responsibility to turn it off, not asmcli's responsibility to provide a flag to customize the stackdriver on or off during the installation.

PR #1125, opened 2022-02-17
added openshift profile for asm installation

aryan16:


asm/istio/options/openshift.yaml
```diff
@@ -0,0 +1,38 @@
+# Copyright 2020 Google LLC
```
bianpengyuan, line 1: s/2020/2022

asm/istio/options/openshift.yaml
```diff
@@ -0,0 +1,38 @@
...
+      namespace: kube-system
+  values:
+    global:
+      pilotCertProvider: "istiod"
+    cni:
+      cniBinDir: /var/lib/cni/bin
+      cniConfDir: /etc/cni/multus/net.d
+      chained: false
+      cniConfFileName: "istio-cni.conf"
+      excludeNamespaces:
```
bianpengyuan, line 31: No need to include line 31-33. It is already the default.

asm/istio/options/openshift.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - istio-system
+      - kube-system
+      logLevel: info
```
bianpengyuan, line 34: This is also the default. No need to include this.

asm/istio/options/openshift.yaml
aryan16: sorry my bad, I directly copied it. Will change it.

asm/istio/options/openshift.yaml
aryan16: sure

asm/istio/options/openshift.yaml
aryan16: cool

bianpengyuan (approved):

zerobfd (approved):

PR #1137, opened 2022-02-25
Use HTTPS_PROXY correctly when downloading packages.

zerobfd:


ZhengzheYang (approved):

jtrbs:
Not sure if this will impact multicloud/onprem CI. Previously `HTTPS_PROXY` was added in the script for connecting to the k8s control plane of clusters like AWS, BM in `kubectl` `istioctl`. Maybe we would need two different proxy env,  `ARTIFACT_HTTPS_PROXY` used to download packages, `K8S_HTTPS_PROXY` used for `kubectl` `istioctl`..

PR #1142, opened 2022-03-08
update gen-eastwest-gateway.sh

jtrbs:
This PR update the gen-eastwest-gateway.sh to the [latest one in Istio OSS](https://raw.githubusercontent.com/istio/istio/release-1.13/samples/multicluster/gen-eastwest-gateway.sh). The latest one uses injection template and is backward compatible. There is no need to specify setters like `--set spec.values.global.pilotCertProvider=kubernetes` at https://cloud.google.com/service-mesh/docs/unified-install/off-gcp-multi-cluster-setup#mesh-ca-or-ca-service. This will unify the east west gateway installation process.

### DIFF
--- a/asm/istio/expansion/gen-eastwest-gateway.sh
+++ b/asm/istio/expansion/gen-eastwest-gateway.sh
@@ -20,14 +20,12 @@ SINGLE_CLUSTER=0
 REVISION=""
 while (( "$#" )); do
   case "$1" in
-    # Node images can be found at https://github.com/kubernetes-sigs/kind/releases
-    # For example, kindest/node:v1.14.0
     --single-cluster)
       SINGLE_CLUSTER=1
       shift
     ;;
     --cluster)
-      CLUSTER=$2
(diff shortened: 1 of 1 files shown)


asm/istio/expansion/gen-eastwest-gateway.sh
```diff
@@ -92,15 +90,12 @@ IOP=$(cat <<EOF
 $IOP
         enabled: true
         k8s:
-          env:
-            # sni-dnat adds the clusters required for AUTO_PASSTHROUGH mode
-            - name: ISTIO_META_ROUTER_MODE
-              value: "sni-dnat"
```
jtrbs, line 98: Hi @howardjohn, is this change expected?

howardjohn: this setting hasn't been needed for a few versions, so as long as this isn't used with super old versions then yes.

zerobfd (approved):

PR #1152, opened 2022-03-16
Accept individual API enablement

zerobfd:

PR #1158, opened 2022-03-25
updated fleet registration for openshift cluster

aryan16:
Added an extra parameter while creating membership for openshift cluster.

### DIFF
--- a/asmcli/lib/installation_dependencies.sh
+++ b/asmcli/lib/installation_dependencies.sh
@@ -43,7 +43,8 @@ required_iam_roles() {
 }
 # [END required_iam_roles]
 
-old_required_apis() {
+# [START required_apis]
+required_apis() {
   local CA; CA="$(context_get-option "CA")"
     cat << EOF
 container.googleapis.com
@@ -57,24 +58,10 @@ gkehub.googleapis.com
 cloudresourcemanager.googleapis.com
 stackdriver.googleapis.com
 EOF
-   case "${CA}" in
+  case "${CA}" in
    mesh_ca)
      echo meshca.googleapis.com
      ;;
-   gcp_cas)
-     echo privateca.googleapis.com
-     ;;
-    *);;
-  esac
-}
-
-# [START required_apis]
-required_apis() {
-  local CA; CA="$(context_get-option "CA")"
-    cat << EOF
-mesh.googleapis.com
-EOF
-  case "${CA}" in
    gcp_cas)
      echo privateca.googleapis.com
      ;;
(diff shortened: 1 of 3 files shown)


asmcli/lib/installation_dependencies.sh
```diff
@@ -43,7 +43,8 @@ required_iam_roles() {
 }
 # [END required_iam_roles]
 
-old_required_apis() {
+# [START required_apis]
+required_apis() {
```
ZhengzheYang, line 47: is there a reason for the function name change? I think this was changed in #1152 for a reason, but I am not sure.

asmcli/lib/installation_dependencies.sh
```diff
@@ -66,20 +67,6 @@ EOF
...
-  local CA; CA="$(context_get-option "CA")"
-    cat << EOF
-mesh.googleapis.com
-EOF
-  case "${CA}" in
-   gcp_cas)
-     echo privateca.googleapis.com
-     ;;
-    *);;
-  esac
```
ZhengzheYang, line 82: same here. Is there a reason we delete these lines?

asmcli/lib/installation_dependencies.sh
```diff
@@ -409,4 +400,4 @@ get_auth_token() {
   local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
   local TOKEN; TOKEN="$(retry 2 gcloud --project="${PROJECT_ID}" auth print-access-token)"
   echo "${TOKEN}"
-}
+}
```
ZhengzheYang, line 403: new line

asmcli/lib/installation_dependencies.sh
```diff
@@ -320,6 +307,7 @@ register_cluster() {
   local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
   local KCF; KCF="$(context_get-option "KUBECONFIG")"
   local KCC; KCC="$(context_get-option "CONTEXT")"
+  local OPENSHIFT_CLUSTER; OPENSHIFT_CLUSTER="$(context_get-option "OPENSHIFT_CLUSTER_SETUP")"
```
ZhengzheYang, line 310: I don't see where you set this variable. You might need to call `context_set-option` to set it somewhere and maybe add it to the [context file](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/main/asmcli/lib/context.sh#L68) to get a default env var value.

asmcli/lib/installation_dependencies.sh
aryan16: I didn't change this, it was required to run `scripts/release-asm/precommit` for the PR

asmcli/lib/installation_dependencies.sh
aryan16: Same reason as above.

asmcli/lib/installation_dependencies.sh
ZhengzheYang: the precommit only merges the changes made by you so maybe you changed this by accident or have a stale commit. I would suggest keeping only the changes made by you and rebase on HEAD.

PR #1159, opened 2022-03-25
updated fleet registration for openshift clusters

aryan16:
Added an extra parameter while creating membership for openshift cluster.

### DIFF
--- a/asmcli/lib/context.sh
+++ b/asmcli/lib/context.sh
@@ -64,7 +64,8 @@ context_init() {
     "OFFLINE": "${OFFLINE:-0}",
     "INCLUDES_PROMETHEUS": "${INCLUDES_PROMETHEUS:-0}",
     "INCLUDES_STACKDRIVER": "${INCLUDES_STACKDRIVER:-1}",
-    "CHANNEL": "${CHANNEL:-}"
+    "CHANNEL": "${CHANNEL:-}",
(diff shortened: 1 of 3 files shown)


ZhengzheYang:
Code LGTM, waiting for @zerobfd to see if we should merge this since asmcli is in maintenance mode.

asmcli/lib/context.sh
```diff
@@ -64,7 +64,8 @@ context_init() {
     "OFFLINE": "${OFFLINE:-0}",
     "INCLUDES_PROMETHEUS": "${INCLUDES_PROMETHEUS:-0}",
     "INCLUDES_STACKDRIVER": "${INCLUDES_STACKDRIVER:-1}",
-    "CHANNEL": "${CHANNEL:-}"
+    "CHANNEL": "${CHANNEL:-}",
+    "OPENSHIFT_CLUSTER_SETUP": "${OPENSHIFT_CLUSTER_SETUP:-}"
```
zerobfd, line 68: Can we name this variable something more general? e.g. `PRIVATE_ISSUER` or `USE_JWKS_NOT_OIDC` or `NO_PUBLIC_ISSUER` or something. It doesn't fundamentally have anything to do with Openshift.

zerobfd (changes requested):

asmcli/lib/context.sh
aryan16: yeah sounds good, I used that name because it is only being used for openshift now. Good suggestion, will make the change.

zerobfd (approved):

PR #1172, opened 2022-04-04
Set HTTPS_PROXY when using kpt

zerobfd:


asmcli/lib/util.sh
```diff
@@ -257,12 +257,18 @@ EOF
...
 download_kpt_package() {
-  local SAMPLES_URL
+  local SAMPLES_URL HTTPS_PROXY
   SAMPLES_URL="${KPT_URL/asm@/samples@}"
+  HTTPS_PROXY="$(context_get-option "HTTPS_PROXY")"
 
   info "Downloading ASM kpt package..."
-  retry 3 kpt pkg get --auto-set=false "${KPT_URL}" asm
-  retry 3 kpt pkg get --auto-set=false "${SAMPLES_URL}" samples
+  if [[ -n "${HTTPS_PROXY}" ]]; then
```
ZhengzheYang, line 265: does `kpt pkg set` require http proxy? I wonder if it should be done in `kpt()` or something like we do with `kubectl`, but I am not sure.

ZhengzheYang (approved):

asmcli/lib/util.sh
zerobfd: Nope, set only modifies local files.

PR #1176, opened 2022-04-13
Print opts pulled from env vars

zerobfd:
This is a pretty common point of friction when following `export FOO` steps in installation/demo guides. This at least makes it more visible what's going on.


ZhengzheYang (approved):
neat!

PR #1180, opened 2022-04-19
adding git as a client dependency

sce-taid:


zerobfd (changes requested):
Thank you for the PR! The `asmcli` tool is generated from other files and usually isn't edited directly. In this case since it's a small change, if you just make the same edit to `asmcli/lib/validate.sh` then it should work. More details on the dev process are in the docs/ folder if you're interested.

sce-taid:
Hi,
github-asm-pkg-script-updated-main (asm-scriptaro-oss) is failing; could you have a look (e.g., at CB logs).
Thanks!

zerobfd:
Hello! It's failing because the change is now in validate.sh but not asmcli. The change should be in both of them. You had the correct change in the original PR, it just *also* needed to be in validate.sh.

sce-taid:
understood :) - thanks!

zerobfd (approved):

zerobfd:
Thanks for the PR and all of the iterations!

PR #1181, opened 2022-04-19
skip cni install if node pool wi is not enabled

ZhengzheYang:
This PR fixes 2 bugs and changes 1 behavior
1. it fixes the bug when the function `is_autopilot` is called too early to always return an error. If `asmcli` detects the cluster is autopilot, it shouldn't ever attempt to apply anything to the `kube-system`. It should always use managed CNI. The user shouldn't need to supply the `--use_managed_cni` flag. I moved that function to the `validate_control_plane` and refactored the validation for autopilot.
2. it fixes the bug that when we check for the node pool WI, we assumed only 1 node pool is present. It would directly fail on clusters with multiple node pools. Changed it to be a loop.
3. it changes the behavior that when the user's cluster does not have node pool WI enabled, it shouldn't just abort. It would set the env var in the `install-cni` container to skip the installation, and give a warning only. The env var is added with https://github.com/istio/istio/pull/38158

### DIFF
--- a/asmcli/commands/validate.sh
+++ b/asmcli/commands/validate.sh
@@ -91,15 +91,32 @@ validate_dependencies() {
 }
 
 validate_control_plane() {
-  if is_managed; then
+  if is_autopilot; then
+    validate_autopilot
+  elif is_managed; then
     if is_legacy; then
       # Managed legacy must be able to set IAM permissions on a generated user, so the flow
       # is a bit different
       validate_managed_control_plane_legacy
     fi
-    # Node-level Workload Identity is required for managed CNI
-    validate_node_pool_workload_identity
   else
     validate_in_cluster_control_plane
   fi
+}
+
+validate_autopilot() {
+  if ! is_managed; then
+    fatal "Autopilot clusters are only supported with managed control plane."
+  fi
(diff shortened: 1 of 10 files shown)


asmcli/commands/validate.sh
```diff
@@ -91,15 +91,32 @@ validate_dependencies() {
 }
 
 validate_control_plane() {
-  if is_managed; then
+  if is_autopilot; then
+    validate_autopilot
+  elif is_managed; then
```
zerobfd, line 96: Is there a reason this is an `elif` instead of two separate if blocks?

zerobfd (approved):

asmcli/commands/validate.sh
ZhengzheYang: I think they have the same effect. If it's not managed or legacy, it would've aborted, so it wouldn't reach the next if. I think `elif` makes it clear that it's either Autopilot or legacy MCP checks. I could make it two separate if blocks if you prefer?

zerobfd: Nope, that make sense. Just wanted to make sure I understood.

PR #1184, opened 2022-04-22
Fix name of istio remote secret for GKE cluster.

kyama902:
This PR fixes name of istio remote secret created by `asmcli create-mesh` subcommand.

In using multicluster service mesh, inter-cluster routing by `topology.istio.io/cluster` label in DestinationRule does not work. It causes HTTP 503 response with `no healthy upstream`.
This problem seems caused by mismatch between cluster name and istio remote secret name.

Cluster name is generated by `asmcli install` subcommand and it is form as `cn-PROJECT_ID-LOCATION-CLUSTER_NAME`.
But, for name of istio remote secret generated by `asmcli create-mesh` subcommand, it forms `istio-remote-secret-gke-PROJEC_ID-LOCATION-CLUSTER_NAME`.

This PR changes `SECRET_NAME` as `cn-PROJECT_ID-LOCATION-CLUSTER_NAME` form if remote cluster is GKE.

### DIFF
--- a/asmcli/asmcli
+++ b/asmcli/asmcli
@@ -423,10 +423,8 @@ install_one_remote_secret() {
   SECRET_NAME="${CTX1//_/-}"
   SECRET_NAME="${SECRET_NAME//\./-}"
   SECRET_NAME="${SECRET_NAME//@/-}"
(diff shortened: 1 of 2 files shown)


zerobfd (approved):
Thank you!

PR #1191, opened 2022-05-05
Add new hpa and pdb versions for samples

richardwxn:
Based on deprecation notice: https://kubernetes.io/docs/reference/using-api/deprecation-guide/.

Similar changes are added in OSS Istio.

I will add similar change to egress gateway part if this looks good


richardwxn:
@zerobfd Do you think whether this layout looks ok?

zerobfd (approved):
LGTM, let's make sure this doesn't change any of the commands that we use in Cloud docs when referring to the samples.

samples/gateways/istio-ingressgateway/autoscaling-v2.yaml
```diff
@@ -0,0 +1,21 @@
+# Optional: HorizontalPodAutoscaler will automatically scale the gateway replica count based on
+# Used on >=1.23 K8S versions
+# CPU utilization
+apiVersion: autoscaling/v2beta1
```
howardjohn, line 4: shouldn't this be `v2`?

richardwxn: fixed

richardwxn:
@zerobfd Can you help to merge it? Thanks

PR #1198, opened 2022-05-10
move autoscaling v2 sample yaml location

richardwxn:
1. move autoscaling v2 sample yaml location into subfolder for now since ASM 1.13 not yet officially support gke 1.23
2. remove pdb-v1beta1 since our least support version for ASM 1.13 is already gke 1.21


richardwxn:
@zerobfd PTAL

zerobfd (approved):

PR #1218, opened 2022-06-07
Configure managed CAS based on managed_cas.yaml

lei-tang:
Configure managed CAS based on managed_cas.yaml.


shankgan (approved):

asmcli/components/ca/managed-cas.sh
```diff
@@ -0,0 +1,20 @@
+configure_managed_cas() {
```
zerobfd, line 1: Since this is going under the experimental subcommand, can you prefix this with `x_` to follow convention?

zerobfd (changes requested):

asmcli/components/ca/managed-cas.sh
lei-tang: Done. Please take another look.

zerobfd (approved):

PR #1220, opened 2022-06-08
Add a function for enabling the workload certificate API

lei-tang:
Add a function for enabling the workload certificate API.

### DIFF
--- a/asmcli/components/ca/managed-cas.sh
+++ b/asmcli/components/ca/managed-cas.sh
@@ -18,3 +18,11 @@ x_configure_managed_cas() {
 
   configure_trust_domain_aliases
 }
+
+x_enable_workload_certificate_api() {
+  local WORKLOAD_CERT_API; WORKLOAD_CERT_API="$1"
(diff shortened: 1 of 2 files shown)


asmcli/components/ca/managed-cas.sh
```diff
@@ -18,3 +18,11 @@ x_configure_managed_cas() {
 
   configure_trust_domain_aliases
 }
+
+x_enable_workload_certificate_api() {
+  local WORKLOAD_CERT_API; WORKLOAD_CERT_API="$1"
+  local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
+
+  info "Enabling the workload certificate API for ${FLEET_ID} ..."
+  retry 2 run_command gcloud services enable --project="${FLEET_ID}" "${WORKLOAD_CERT_API}"
```
shankgan, line 27: did you define WORKLOAD_CERT_API elsewhere?
Also note that for managed_ca, gkehub.googleapis.com is also a requirement - however, think we need that to register clusters to a fleet. If not, this is also a required API

shankgan (approved):

PR #1222, opened 2022-06-08
Enable the workload certificate API

lei-tang:
Enable the workload certificate API.

### DIFF
--- a/asmcli/lib/installation_dependencies.sh
+++ b/asmcli/lib/installation_dependencies.sh
@@ -77,6 +77,9 @@ EOF
    gcp_cas)
      echo privateca.googleapis.com
      ;;
+   managed_cas)
+     echo workloadcertificate.googleapis.com
+     ;;
     *);;
   esac
 }
@@ -90,6 +93,9 @@ EOF
   case "${CA}" in
    gcp_cas)
      echo privateca.googleapis.com
+     ;;
+   managed_cas)
+     echo workloadcertificate.googleapis.com
      ;;
     *);;
   esac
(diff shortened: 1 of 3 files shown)


asmcli/lib/installation_dependencies.sh
```diff
@@ -186,6 +186,12 @@ enable_gcloud_apis(){
   # shellcheck disable=SC2046
   retry 3 gcloud services enable --project="${PROJECT_ID}" $(required_apis | tr '\n' ' ')
 
+  local CA; CA="$(context_get-option "CA")"
+  if [[ "${CA}" = "managed_cas" ]]; then
+    x_enable_gke_hub_api "autopush-gkehub.sandbox.googleapis.com"
```
zerobfd, line 191: If the dependencies aren't in prod yet then we generally don't push to the main development branch. Is there a strong reason that we need it in GitHub at this level of development?

zerobfd (changes requested):
If you look at the "required_apis()" function there's already a place where we determine apis based off of dependencies. We should put the logic there.

asmcli/lib/installation_dependencies.sh
lei-tang: The dependencies are not in prod yet. I agree with you on the the comments. I will update this PR when the dependencies are in prod. After the PR is updated, I will ask for your review again. Thanks!

lei-tang: The dependencies are not in prod yet. I agree with you on the the comments. I will update this PR when the dependencies are in prod. After the PR is updated, I will ask for your review again. Thanks!

lei-tang: @zerobfd I have updated the PR based on your comment. Please take another look. Thanks!

zerobfd (approved):

PR #1223, opened 2022-06-09
Enable workload certificate feature on the fleet

lei-tang:
Enable workload certificate feature on the fleet.

### DIFF
--- a/asmcli/components/ca/managed-cas.sh
+++ b/asmcli/components/ca/managed-cas.sh
@@ -26,3 +26,35 @@ x_enable_workload_certificate_api() {
   info "Enabling the workload certificate API for ${FLEET_ID} ..."
   retry 2 run_command gcloud services enable --project="${FLEET_ID}" "${WORKLOAD_CERT_API}"
 }
+
+x_exit_if_no_auth_token() {
+  local AUTHTOKEN; AUTHTOKEN="$(get_auth_token)"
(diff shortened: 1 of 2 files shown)


zerobfd (approved):

asmcli/components/ca/managed-cas.sh
```diff
@@ -26,3 +26,35 @@ x_enable_workload_certificate_api() {
...
+        'provision_google_ca': 'ENABLED'
+      }
+    }
+  }"
+
+  curl -H "Authorization: Bearer ${AUTHTOKEN}" \
+      -X POST -H "Content-Type: application/json" -H "Accept: application/json" \
+      -d "${BODY}" \
+      "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features?feature_id=workloadcertificate"
+}
```
shankgan, line 60: nit1: is there any reason we aren't using the gcloud api's? is it because we don't expect they have the newest version of gcloud?

nit2: I believe we also need to add the actual membership as dictated here. This curl just enables workloadcertificate feature on the fleet - go/workload-identity-platform-site/controller/user_guide/controller_user_guide#enable-feature-on-members

lei-tang: 1) Yes. 2) That will be in a follow-up PR.

PR #979, opened 2021-09-22
Simplify gateway samples

howardjohn:

PR #1244, opened 2022-06-20
Gateways samples - unprivileged container

mathieu-benoit:
Add back the unprivileged container setup for both ingress and egress gateways. This was originally here and was removed since [this PR](https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/979).

This PR updates the `Deployments` with [`securityContext` features](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) as well as the `Services` in order to use non-privilege ports on the container (`8443` and `8080`). With that we have more security features by default with these samples. Also, it allows these samples to run on any restricted and secured environment (think PSP, OpenShift, etc.).

### DIFF
--- a/samples/gateways/istio-egressgateway/deployment.yaml
+++ b/samples/gateways/istio-egressgateway/deployment.yaml
@@ -31,8 +31,20 @@ spec:
         app: istio-egressgateway
         istio: egressgateway
     spec:
+      securityContext:
+        fsGroup: 1337
+        runAsGroup: 1337
+        runAsNonRoot: true
+        runAsUser: 1337
       containers:
       - name: istio-proxy
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - all
+          privileged: false
+          readOnlyRootFilesystem: true
         image: auto # The image will automatically update each time the pod starts.
         resources:
           limits:
--- a/samples/gateways/istio-egressgateway/service.yaml
+++ b/samples/gateways/istio-egressgateway/service.yaml
@@ -24,8 +24,10 @@ spec:
   # Any ports exposed in Gateway resources should be exposed here.
   - name: http2
     port: 80
+    targetPort: 8080
   - name: https
     port: 443
+    targetPort: 8443
   selector:
     istio: egressgateway
     app: istio-egressgateway
--- a/samples/gateways/istio-ingressgateway/deployment.yaml
+++ b/samples/gateways/istio-ingressgateway/deployment.yaml
@@ -32,8 +32,20 @@ spec:
         app: istio-ingressgateway
         istio: ingressgateway
     spec:
+      securityContext:
+        fsGroup: 1337
+        runAsGroup: 1337
+        runAsNonRoot: true
+        runAsUser: 1337
       containers:
       - name: istio-proxy
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - all
+          privileged: false
+          readOnlyRootFilesystem: true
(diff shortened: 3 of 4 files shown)


mathieu-benoit:
FYI @zerobfd and @howardjohn

howardjohn (changes requested):
This is a downtime-causing event if users apply the new Services. In newer Kubernetes there is a best-of-both-worlds options using sysctl to allow binding to low ports. This is the only safe way I know to migrate existing users to rootless. However, it requires a newer k8s version. In the helm charts we have auto detection

mathieu-benoit:
Thanks @howardjohn, interesting, I didn't know about the downtime-causing event when changing the `targetPort` from `80` to `8080` in this example. I tested this myself and it was not that long/bad. How bad/long is it based on your experience? Any pointer that I could learn from about this?

I just looked at the Helm chart you mentioned (thanks for sharing!): https://github.com/istio/istio/blob/0e7940128eb8405b2283742cbf71241c1972ef3d/manifests/charts/gateway/templates/deployment.yaml#L67 and from there, I see what you mean with Kubernetes 1.22+.

Based on that, I see 2 options:
1. Keep my PR as-is with the `targetPort` to `8080` and `8443`. That's best practice that we could have to add more security, knowing that there is this downtime-causing event mentioned.
2. Update this PR to have the `CAP_NET_BIND_SERVICE` capability in order to bind to low port (`targetPort` pointing to `port`'s value like it was before this PR)

_Note: the new `sysctl` is not yet an avenue we could take because it requires end-users to have Kubernetes 1.22+._

Thoughts?

howardjohn:
> I'm also seeing this note in the Istio 1.7's release notes: https://istio.io/latest/news/releases/1.7.x/announcing-1.7/upgrade-notes/#gateways-run-as-non-root
> 
> > Note: the targetPort only modifies which port the gateway binds to. Clients will still connect to the port defined by port (generally 80 and 443), so this change should be transparent.
> 
> Which is exaclty what's in this PR.

This was before we got reports for months about all the outages we caused and how Istio doesn't care about compatibility, etc. We learned our lesson... (but I guess didn't update the release notes).

 I specifically don't want to switch the targetPort if possible (individual users can do it of course, but not recommend) since we cannot migrate back either (also causes downtime) and it makes the API model confusing (does Port refer to targetPort or serbice port? there is a long trail of issues around this we are trying to clean up).

Our approach was to have rootless "require" 1.21+,so the problem fixes itself with enough time passed.

That being said the samples don't follow that approach since we don't have a way to dynamically enable the sysctl option. We are pretty close to most users being on 1.21+ I think though.. 

Note - afaik CAP_NET_BIND_SERVICE doesn't work due to some obscure limitations around containers and ambient capabilities.

mathieu-benoit:
Oh ok, thanks for sharing all these insightful experiences.

RE: `NET_BIND_SERVICE`, I just tested and indeed it's not working as expected with my GKE cluster in 1.22, so doesn't look like a good solution, agree with you.

I think this ASM Ingress Gateway sample doesn't have the usage and scale that the Ingress Gateway Istio has upstream (when installing Istio via IstioOperator or Helm chart), so I still think it would be good to have security best practice in place here, even if there is a downtime-causing event with these `targetPort` values.

With all of that said in that thread, I'm closing this PR since it's not something we want implemented here.

howardjohn:
Yeah it's tricky since this is by nature a plaintext yaml instead of a template.. we could have a 1.21+ folder (or exactly what you have here), but then you end up with a weird combinatorial explosion of options.

All of this will eventually be replaced with the new Gateway API but it's a ways off

PR #1247, opened 2022-06-21
Reorder managed control plane install to prevent race

shankgan:


ZhengzheYang (approved):
cc @zerobfd

shankgan:
problem: control plane comes up first, we don't "configure" the cluster with ASM_OPTIONS until later. control plane bringup should only happen AFTER ASM_OPTIONS has been added to cluster

zerobfd (changes requested):
Make sure and re-generate asmcli before committing

PR #1252, opened 2022-06-27
Wait for workload certificate enablement

lei-tang:
Enabling workload certificate feature on a cluster will result in the cluster restarting and the cluster becoming unreachable for minutes. The PR waits for the k8s cluster becoming reachable and checks the workload certificate enablement status before proceeding with the installation.

### DIFF
--- a/asmcli/commands/validate.sh
+++ b/asmcli/commands/validate.sh
@@ -55,8 +55,7 @@ validate_dependencies() {
         enable_service_mesh_feature
       fi
       if [[ "${CA}" == "managed_cas" ]]; then
-        info "Wait 3 minutes for GKE HUB API enablement to propagate to the systems ..."
-        sleep 180
+        x_wait_for_gke_hub_api_enablement
         x_enable_workload_certificate_on_fleet "gkehub.googleapis.com"
       fi
     else
--- a/asmcli/commands/validate.sh
+++ b/asmcli/commands/validate.sh
@@ -55,6 +55,8 @@ validate_dependencies() {
         enable_service_mesh_feature
       fi
       if [[ "${CA}" == "managed_cas" ]]; then
+        info "Wait 3 minutes for GKE HUB API enablement to propagate to the systems ..."
+        sleep 180
         x_enable_workload_certificate_on_fleet "gkehub.googleapis.com"
       fi
     else
@@ -73,6 +75,7 @@ validate_dependencies() {
     local HUB_MEMBERSHIP_ID; HUB_MEMBERSHIP_ID="$(context_get-option "HUB_MEMBERSHIP_ID")"
     if [[ "${CA}" == "managed_cas" ]]; then
       x_enable_workload_certificate_on_membership "gkehub.googleapis.com" "${FLEET_ID}" "${HUB_MEMBERSHIP_ID}"
+      x_wait_for_enabling_workload_certificates "gkehub.googleapis.com" "${FLEET_ID}"
     fi
   elif should_validate && [[ "${USE_HUB_WIP}" -eq 1 || "${USE_VM}" -eq 1 ]]; then
     exit_if_cluster_unregistered
--- a/asmcli/components/ca/managed-cas.sh
+++ b/asmcli/components/ca/managed-cas.sh
@@ -75,6 +75,54 @@ x_enable_workload_certificate_on_membership() {
      -d "${ENABLEFEATURE}" "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features/workloadcertificate?update_mask=membership_specs"
 }
 
+x_wait_for_enabling_workload_certificates() {
+  local GKEHUB_API; GKEHUB_API="$1"
+  local FLEET_ID; FLEET_ID="$2"
+
+  info "Waiting for the workload certificates enablement. This may take up to 20 minutes ..."
+  # Enabling workload certificate feature on a cluster will result in
+  # the cluster restarting and the cluster becoming unreachable for minutes.
+  # The cluster restart may happen at a random time after the enablement.
+  # Therefore, wait 15 minutes before checking the connection to k8s is recovered.
+  sleep 900
+  # Check the connection to k8s is recovered.
+  local REACHABLE; REACHABLE=false
+  for i in {1..3}; do
+    local OUT; OUT=$(kubectl version -o json 2>/dev/null | jq .serverVersion.major -r 2>/dev/null)
+    if [[ $OUT -eq 1 ]]; then
+      info "The k8s cluster is reachable."
+      REACHABLE=true
+      break
+    else
+      info "The k8s cluster not reachable, try again ..."
+      sleep 60
+    fi
+  done
+  if ! $REACHABLE; then
+    fatal "The k8s cluster is not reachable, exit."
+  fi
+
+  # Check the status of workload certificate feature
+  local ENABLED; ENABLED=false
+  for i in {1..2}; do
+    local OUT; OUT="$(curl -H "X-Goog-User-Project: ${FLEET_ID}" \
+      -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+      "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features/workloadcertificate" 2>/dev/null | \
+      grep -i -c "certificateManagement.*ENABLED")"
+    if [[ $OUT -gt 0 ]]; then
+      info "Workload certificate management is enabled."
+      ENABLED=true
+      break
+    else
+      info "Workload certificate management is not enabled, try again ..."
+      sleep 60
+    fi
+  done
+  if ! $ENABLED; then
+    warn "The workload certificate management is not enabled. The workload certificates may not work until the management is enabled."
+  fi
+}
+
 x_install_managed_cas_for_mcp() {
   info "Configuring managed CAS for managed control plane..."
 
--- a/asmcli/components/ca/managed-cas.sh
+++ b/asmcli/components/ca/managed-cas.sh
@@ -113,18 +113,18 @@ x_wait_for_enabling_workload_certificates() {
   # Check the connection to k8s is recovered.
   local REACHABLE; REACHABLE=0
   for i in {1..3}; do
-    local OUT; OUT=$(kubectl version -o json 2>/dev/null | jq .serverVersion.major -r 2>/dev/null)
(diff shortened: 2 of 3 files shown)


asmcli/commands/validate.sh
```diff
@@ -55,6 +55,8 @@ validate_dependencies() {
         enable_service_mesh_feature
       fi
       if [[ "${CA}" == "managed_cas" ]]; then
+        info "Wait 3 minutes for GKE HUB API enablement to propagate to the systems ..."
+        sleep 180
```
shankgan, line 59: should this come after x_enable_workload_certificate_on_fleet?

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,54 @@ x_enable_workload_certificate_on_membership() {
...
+  fi
+
+  # Check the status of workload certificate feature
+  local ENABLED; ENABLED=false
+  for i in {1..2}; do
+    local OUT; OUT="$(curl -H "X-Goog-User-Project: ${FLEET_ID}" \
+      -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+      "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features/workloadcertificate" 2>/dev/null | \
+      grep -i -c "certificateManagement.*ENABLED")"
+    if [[ $OUT -gt 0 ]]; then
```
shankgan, line 112: is OUT supposed to be a string? could we just not do a -z here?

shankgan: IMO, this check should come before the previous one - the CLH propagation happens reasonably quickly. Its only the restart of the cluster that can take some time...

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,54 @@ x_enable_workload_certificate_on_membership() {
...
+x_wait_for_enabling_workload_certificates() {
+  local GKEHUB_API; GKEHUB_API="$1"
+  local FLEET_ID; FLEET_ID="$2"
+
+  info "Waiting for the workload certificates enablement. This may take up to 20 minutes ..."
+  # Enabling workload certificate feature on a cluster will result in
+  # the cluster restarting and the cluster becoming unreachable for minutes.
+  # The cluster restart may happen at a random time after the enablement.
+  # Therefore, wait 15 minutes before checking the connection to k8s is recovered.
+  sleep 900
```
shankgan, line 87: IMO, this i a long wait time..should we perhaps try a smaller initial wait followed by a larger number of retries?

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,54 @@ x_enable_workload_certificate_on_membership() {
...
+  # Check the connection to k8s is recovered.
+  local REACHABLE; REACHABLE=false
+  for i in {1..3}; do
```
shankgan, line 90: there is already a retry macro. Perhaps we can amend the existing logic to add a delay there? cc @zerobfd

asmcli/commands/validate.sh
zerobfd, line 59: If there needs to be a sleep, if should go in one of the x_ functions specific to managed CAS. It makes me nervous putting it into a function that's called in the non-experimental install path.

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,54 @@ x_enable_workload_certificate_on_membership() {
...
+  local FLEET_ID; FLEET_ID="$2"
+
+  info "Waiting for the workload certificates enablement. This may take up to 20 minutes ..."
+  # Enabling workload certificate feature on a cluster will result in
+  # the cluster restarting and the cluster becoming unreachable for minutes.
+  # The cluster restart may happen at a random time after the enablement.
+  # Therefore, wait 15 minutes before checking the connection to k8s is recovered.
+  sleep 900
+  # Check the connection to k8s is recovered.
+  local REACHABLE; REACHABLE=false
```
zerobfd, line 89: "false" and "true" are just strings in shell, you should use 0 or 1.

asmcli/components/ca/managed-cas.sh
zerobfd: I'm fine with this for the experimental release, we can refactor later if we need to.

zerobfd (changes requested):

asmcli/commands/validate.sh
lei-tang: This should be before x_enable_workload_certificate_on_fleet because before enabling workload certificate on fleet, the GKE HUB API enablement needs to be propagated to the systems.

asmcli/components/ca/managed-cas.sh
lei-tang: A smaller initial wait followed by a larger number of retries does not always work because the cluster restart may happen at a random time after the enablement. For example, if the initial wait is 1 minute, the cluster may not restart yet and thus passing the check. Then after 1 minute, the cluster begins restarting and causes the installation to fail.

asmcli/components/ca/managed-cas.sh
lei-tang: It sounds good to me to refactor later if we need to.

asmcli/commands/validate.sh
lei-tang: Done: it has gone to one of the x_ functions specific to managed CAS.

asmcli/components/ca/managed-cas.sh
lei-tang: Done.

lei-tang:
@shankgan @zerobfd Thank you for your reviews! All comments have been addressed. Please take another look.

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,59 @@ x_enable_workload_certificate_on_membership() {
      -d "${ENABLEFEATURE}" "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features/workloadcertificate?update_mask=membership_specs"
 }
 
+x_wait_for_gke_hub_api_enablement() {
+  info "Wait 3 minutes for GKE HUB API enablement to propagate to the systems ..."
```
williamaronli, line 79: waIt 3 minutes is it enough?

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,59 @@ x_enable_workload_certificate_on_membership() {
...
+
+  # Enabling workload certificate feature on a cluster will result in
+  # the cluster restarting and the cluster becoming unreachable for minutes.
+  # The cluster restart may happen at a random time after the enablement.
+  # Therefore, wait 15 minutes before checking the connection to k8s is recovered.
+  sleep 900
+  # Check the connection to k8s is recovered.
+  local REACHABLE; REACHABLE=0
+  for i in {1..3}; do
+    local OUT; OUT=$(kubectl version -o json 2>/dev/null | jq .serverVersion.major -r 2>/dev/null)
```
shankgan, line 116: @williamaronli - wonder if we should instead wait to check if node daemon set is installed, instead of checking if the kube api server is up?

asmcli/components/ca/managed-cas.sh
lei-tang: For the production environment, 3 minutes should be enough. For other environments, there may be some issues with the recent implementation, which leads to a long propagation latency.

shankgan: think it should be?..this isn't for the cluster restart though...

asmcli/components/ca/managed-cas.sh
williamaronli: Once the cluster is restarted successfully and the cluster should already is installed with zatar

Right now we use the command: ```kubectl get ds -n kube-system gke-spiffe-node-agent``` could we use this in this script?

shankgan (approved):

asmcli/components/ca/managed-cas.sh
shankgan: @lei-tang - was thinking this is better since this actually tests if the cluster is correctly configured/ready for managed certificates

asmcli/components/ca/managed-cas.sh
lei-tang: This isn't for the cluster restart. The GKE hub API enablement propagation latency for non-production environments can be tens of minutes, which seems to be a bug. The wait here is for normal propagation delays and 3-minute is a reasonable value.

williamaronli: Could we check ```gcloud alpha container fleet workload-certificate describe``` to make sure the provision google ca is enabled?

asmcli/components/ca/managed-cas.sh
```diff
@@ -75,6 +75,59 @@ x_enable_workload_certificate_on_membership() {
...
+  local GKEHUB_API; GKEHUB_API="$1"
+  local FLEET_ID; FLEET_ID="$2"
+
+  info "Waiting for the workload certificates enablement. This may take up to 20 minutes ..."
+  # Check the status of workload certificate feature
+  local ENABLED; ENABLED=0
+  for i in {1..2}; do
+    local OUT; OUT="$(curl -H "X-Goog-User-Project: ${FLEET_ID}" \
+      -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+      "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features/workloadcertificate" 2>/dev/null | \
```
williamaronli, line 93: can we directly use alpha command: ```gcloud alpha container fleet workload-certificate update --all-memberships --enable ``` instead of the https curl recall

asmcli/components/ca/managed-cas.sh
lei-tang: Done.

asmcli/components/ca/managed-cas.sh
lei-tang: It is equivalent to the curl call and does not require the installation of an alpha version of gcloud.

asmcli/components/ca/managed-cas.sh
lei-tang: The gcloud alpha call is equivalent to the curl call and the curl call does not require the installation of an alpha version of gcloud.

williamaronli (approved):

zerobfd (approved):

PR #1259, opened 2022-07-07
ca migration: minor changes

shankgan:


scripts/migration/ca-migration/migrate_ca
```diff
@@ -688,6 +710,7 @@ check_cert_in_pod() {
     fi
     if ! ${rootmap[${SN}]}; then
       echo "${POD}.${NS} does not trust ${CA_CERT}"
+      ERR="true"
```
zerobfd, line 713: I think using 0 and 1, then -eq or -ne to check, is safer.

zerobfd (approved):
Not sure what the argMap is doing over the code that was already there, but looks benign to me.

scripts/migration/ca-migration/migrate_ca
shankgan: done

PR #1267, opened 2022-07-15
Add experimental option managed_certificates to asmcli

xulingqing:
- Patched from @lei-tang asm_wip_integ_bug_bash_production_env_6_24_2022
- Synced to head and resolved some conflicts
- Add an experimental status checker for the option


JimmyCYJ:
LGTM

asmcli/commands/experimental/vm.sh
```diff
@@ -54,6 +54,7 @@ prepare_cluster_subcommand() {
   local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
   local CLUSTER_NAME; CLUSTER_NAME="$(context_get-option "CLUSTER_NAME")"
   local CLUSTER_LOCATION; CLUSTER_LOCATION="$(context_get-option "CLUSTER_LOCATION")"
+  local CA; CA="$(context_get-option "CA")"
```
zerobfd, line 57: Is this line necessary?

asmcli/commands/help.sh
```diff
@@ -290,6 +293,7 @@ FLAGS:
 
      --print_config
      --disable_canonical_service
+  --managed_certificates
```
zerobfd, line 296: I think this is three spaces too far to the left and doesn't align with the other flags

zerobfd (changes requested):
Looks good but one important thing: everywhere that we check if we're using managed_cas, we also need to check to make sure that the EXPERIMENTAL context option is also 1.

A convenient way to do that might be to create a "fail_if_not_experimental()" function that checks to make sure it's an experimental command, and if not fails with a message like "The selected features are in preview and only available when using the `experimental install` subcommand."

asmcli/commands/help.sh
xulingqing: fixed indentation.

asmcli/commands/experimental/vm.sh
xulingqing: Deleted this line.

xulingqing:
> fail_if_not_experimental

Add the function fail_if_not_experimental().

asmcli/lib/validate.sh
```diff
@@ -528,6 +536,7 @@ validate_args() {
   local ENABLE_GCP_COMPONENTS; ENABLE_GCP_COMPONENTS="$(context_get-option "ENABLE_GCP_COMPONENTS")"
   local ENABLE_REGISTRATION; ENABLE_REGISTRATION="$(context_get-option "ENABLE_REGISTRATION")"
   local ENABLE_NAMESPACE_CREATION; ENABLE_NAMESPACE_CREATION="$(context_get-option "ENABLE_NAMESPACE_CREATION")"
+  local EXPERIMENTAL; EXPERIMENTAL="$(context_get-option "EXPERIMENTAL")"
```
zerobfd, line 539: I don't think this line is necessary since it's also in the fail_if_not_experimental function.

asmcli/lib/installation_dependencies.sh
```diff
@@ -201,6 +201,11 @@ enable_gcloud_apis(){
   # shellcheck disable=SC2046
   retry 3 gcloud services enable --project="${PROJECT_ID}" $(required_apis | tr '\n' ' ')
 
+  local CA; CA="$(context_get-option "CA")"
+  if [[ "${CA}" = "managed_cas" ]]; then
```
zerobfd, line 205: Can you also add a fail_if_not_experimental call below this line as well?

zerobfd (approved):
LGTM after a couple small things.

asmcli/lib/validate.sh
xulingqing: deleted

asmcli/lib/installation_dependencies.sh
xulingqing: Added

JimmyCYJ (approved):

PR #1274, opened 2022-07-19
asmcli update x install, add --managed_certificates option

shankgan:


asmcli/commands/experimental/install.sh
```diff
@@ -7,31 +7,12 @@ x_install_subcommand() {
...
-  for i in {1..10}; do
-    if kubectl wait --for condition=established --timeout=10s crd/controlplanerevisions.mesh.cloud.google.com 2>/dev/null; then
-      break
-    fi
-    sleep 10
-  done
-
-  install_control_plane_revisions
-
+  install_control_plane
```
lei-tang, line 15: With the change here, it seems that x_install() will no longer work for the managed control plane. I think x_install should support both the in-cluster control plane and the managed control plane.

shankgan: don't think so. the install_control_plane function branches out into managed and regular control plane installation?

asmcli/components/ca/managed-cas.sh
```diff
@@ -92,7 +92,7 @@ x_wait_for_enabling_workload_certificates() {
       -H "Authorization: Bearer $(gcloud auth print-access-token)" \
       "https://${GKEHUB_API}/v1alpha/projects/${FLEET_ID}/locations/global/features/workloadcertificate" 2>/dev/null | \
       grep -i "certificateManagement.*ENABLED")"
-    if [[ -z "$OUT" ]]; then
+    if [[ -n "$OUT" ]]; then
```
lei-tang, line 95: Why changes -z to -n here?

asmcli/components/ca/managed-cas.sh
```diff
@@ -114,7 +114,7 @@ x_wait_for_enabling_workload_certificates() {
   local REACHABLE; REACHABLE=0
   for i in {1..3}; do
     local OUT; OUT=$(kubectl get ds -n kube-system gke-spiffe-node-agent 2>/dev/null | grep -i "gke-spiffe-node-agent")
-    if [[ -z "$OUT" ]]; then
+    if [[ -n "$OUT" ]]; then
```
lei-tang, line 117: Why changes -z to -n here?

shankgan: IIUC, this condition should check if OUT is not nil. That is accomplished with the -n flag? the -z flag is the opposite?

asmcli/components/ca/managed-cas.sh
shankgan: IIUC, this condition should check if OUT is not nil. That is accomplished with the -n flag? the -z flag is the opposite?

lei-tang: The -z flag checks the string is empty or not, whereas -n checks a variable is nil or not. A non-nil variable may have an empty string. So "-z" should be used here.

asmcli/components/ca/managed-cas.sh
lei-tang: The -z flag checks the string is empty or not, whereas -n checks a variable is nil or not. A non-nil variable may have an empty string. So "-z" should be used here.

shankgan: okay with leaving it as -z, but should the conditions be reversed then? if [[ -z ${OUT} ]]; then means cluster should be unreachable right?

asmcli/components/ca/managed-cas.sh
```diff
@@ -108,23 +108,21 @@ x_wait_for_enabling_workload_certificates() {
   # Enabling workload certificate feature on a cluster will result in
   # the cluster restarting and the cluster becoming unreachable for minutes.
   # The cluster restart may happen at a random time after the enablement.
-  # Therefore, wait 15 minutes before checking the connection to k8s is recovered.
```
lei-tang, line 111: Why this wait is removed?

shankgan: since we are waiting until spiffe-node-agent is installed, I thought this check is unnecessary. the time has instead been moved to the loop below. WDYT?

lei-tang: Moving the wait here to waiting until spiffe-node-agent is installed does not always work because the cluster restart may happen at a random time after the enablement. For example, if the  wait here is moved to waiting until spiffe-node-agent, when waiting for spiffe-node-agent, the cluster may have not restarted yet, thus spiffe-node-agent check may pass and the installation may fail at a later installation stage due to the cluster restart, which explains why the output from Lingqing's test is not the same as that of yours.

shankgan: added back wait

asmcli/components/ca/managed-cas.sh
```diff
@@ -84,7 +84,6 @@ x_wait_for_enabling_workload_certificates() {
   local GKEHUB_API; GKEHUB_API="$1"
   local FLEET_ID; FLEET_ID="$2"
 
-  info "Waiting for the workload certificates enablement. This may take up to 20 minutes ..."
```
lei-tang, line 87: This statement should not be deleted; it is useful to inform a user the long wait ahead.

asmcli/commands/experimental/install.sh
```diff
@@ -1,13 +1,8 @@
 x_install_subcommand() {
-  x_parse_install_args "${@}"
   context_set-option "PLATFORM" "gcp"
   context_set-option "EXPERIMENTAL" 1
-  x_validate_install_args
-  prepare_environment
 
-  x_validate_dependencies
-  x_configure_package
-  x_install
```
lei-tang, line 10: @zerobfd Can you review the removal of these six lines of code are ok?

asmcli/components/ca/managed-cas.sh
shankgan: moved the statement below

zerobfd (approved):

PR #1280, opened 2022-07-21
Temp workaround for attached clusters

zerobfd:


asmcli/lib/installation_dependencies.sh
```diff
@@ -389,14 +389,17 @@ register_cluster() {
...
-  local NOTFOUND; NOTFOUND="$(find_missing_strings "${WANT}" "${LABELS}")"
-
   local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
   local CLUSTER_NAME; CLUSTER_NAME="$(context_get-option "CLUSTER_NAME")"
   local CLUSTER_LOCATION; CLUSTER_LOCATION="$(context_get-option "CLUSTER_LOCATION")"
 
+  # temp workaround for attached clusters
+  if [[ -z "${CLUSTER_LOCATION}" ]]; then return 0; fi
+
+  local LABELS; LABELS="$(get_cluster_labels)";
```
chemikadze, line 399: Side note, I've noticed that get_cluster_labels does `gcloud container clusters describe` even for memberships, is it expected or should it use `gcloud container hub memberships describe`?

asmcli/lib/installation_dependencies.sh
```diff
@@ -389,14 +389,17 @@ register_cluster() {
...
-  local LABELS; LABELS="$(get_cluster_labels)";
-  local WANT; WANT="$(mesh_id_label)"
-  local NOTFOUND; NOTFOUND="$(find_missing_strings "${WANT}" "${LABELS}")"
-
   local PROJECT_ID; PROJECT_ID="$(context_get-option "PROJECT_ID")"
   local CLUSTER_NAME; CLUSTER_NAME="$(context_get-option "CLUSTER_NAME")"
   local CLUSTER_LOCATION; CLUSTER_LOCATION="$(context_get-option "CLUSTER_LOCATION")"
 
+  # temp workaround for attached clusters
+  if [[ -z "${CLUSTER_LOCATION}" ]]; then return 0; fi
```
chemikadze, line 397: Is it guaranteed to be unset only for attached clusters? Regular Anthos clusters don't seem have the issue despite failing to get labels and having CLUSTER_LOCATION unset, is there some interplay with `register_cluster`? In logs I have it appears like it returns on line 405.

chemikadze (approved):

asmcli/lib/installation_dependencies.sh
zerobfd: I'm 90% sure yes, and I'm 100% sure that attached clusters don't support the features that cluster labels enable so this shouldn't affect anything.

asmcli/lib/installation_dependencies.sh
zerobfd: I'll follow up on that one afterwards, but I did confirm this is working for on-prem/bare metal customers for the latest Anthos release. Not sure *why* it works but...

PR #1296, opened 2022-08-03
add cluster labels to Anthos on VMware clusters

cathchu:
ensures mesh_id label based on fleet membership is added to non-GKE clusters


asmcli/lib/installation_dependencies.sh
```diff
@@ -287,18 +287,26 @@ get_cluster_labels() {
...
+      --project="${PROJECT_ID}" \
+      --format='value(resourceLabels)[delimiter=","]')";
   echo "${LABELS}"
+  fi
 }
 
 generate_membership_name() {
   local MEMBERSHIP_NAME; MEMBERSHIP_NAME="$(context_get-option "HUB_MEMBERSHIP_ID")"
-  if [[ -n "${MEMBERSHIP_NAME}" ]]; then echo "${MEMBERSHIP_NAME}"; return; fi
+  if [[ ! -n "${MEMBERSHIP_NAME}" ]]; then echo "${MEMBERSHIP_NAME}"; return; fi
```
zerobfd, line 309: `-n` tests for a non-zero length string, used in asmcli as "not empty". Adding a ! means "not not empty", which I don't think is what we want.

Was this change necessary to get the labelling to work properly?

asmcli/lib/installation_dependencies.sh
```diff
@@ -317,8 +325,6 @@ generate_membership_name() {
       RAND="$(tr -dc "a-z0-9" </dev/urandom | head -c8 || true)"
       MEMBERSHIP_NAME="${CLUSTER_NAME:0:54}-${RAND}"
     fi
-  else
-    MEMBERSHIP_NAME="$(date +%s%N)"
```
zerobfd, line 321: I think removing this line means that anything off-GCP is going to need to be registered beforehand because we don't have a way of generating a membership name from nothing.

That's fine in principle, but I'd rather not add a potentially breaking change in case someone's relying on this. Can you think of a way to generate a more meaningful name from information that we're guaranteed to have even off-GCP?

(I don't really know how off-GCP clusters would get this far in the installation process without registering, but somehow it happens...)

zerobfd (changes requested):

asmcli/lib/installation_dependencies.sh
cathchu: Ah, no this looks to be fine. I'll remove it.

asmcli/lib/installation_dependencies.sh
cathchu: That makes sense, I was initially thinking it would be a requirement for the cluster to be registered (and without the `--enable-registration` or `--enable-all` flag, the script would fail). But I see that the registration uses this function to set parameters. Changing this back!

asmcli/lib/installation_dependencies.sh
```diff
@@ -287,13 +287,21 @@ get_cluster_labels() {
...
+    LABELS="$(retry 2 gcloud container hub memberships describe "${MEMBERSHIP_NAME}" \
+      --project="${PROJECT_ID}" \
+      --format='value(resourceLabels)[delimiter=","]')";
+  else
+    info "Reading labels for ${CLUSTER_LOCATION}/${CLUSTER_NAME}..."
+    LABELS="$(retry 2 gcloud container clusters describe "${CLUSTER_NAME}" \
+      --zone="${CLUSTER_LOCATION}" \
+      --project="${PROJECT_ID}" \
+      --format='value(resourceLabels)[delimiter=","]')";
   echo "${LABELS}"
```
zerobfd, line 303: I think this echo needs to go outside of the if statement. If `is_gcp` is false, then nothing gets printed.

zerobfd (approved):
One change necessary, then lgtm.

asmcli/lib/installation_dependencies.sh
cathchu: fixed.

PR #1307, opened 2022-08-11
Update get label method for off-GCP clusters

zerobfd:


asmcli/lib/installation_dependencies.sh
```diff
@@ -287,12 +287,20 @@ get_cluster_labels() {
...
-    --format='value(resourceLabels)[delimiter=","]')";
+  if ! is_gcp; then
+    local MEMBERSHIP_NAME; MEMBERSHIP_NAME="$(generate_membership_name)"
+    info "Reading labels for ${MEMBERSHIP_NAME}..."
+    LABELS="$(retry 2 gcloud container hub memberships describe "${MEMBERSHIP_NAME}" \
+      --project="${PROJECT_ID}" \
+      --format='value(labels)[delimiter=","]')";
+  else
+    info "Reading labels for ${CLUSTER_LOCATION}/${CLUSTER_NAME}..."
+    LABELS="$(retry 2 gcloud container clusters describe "${CLUSTER_NAME}" \
```
ZhengzheYang, line 299: Since we are requiring the clusters to be registered for all cases, can we just get the label from the membership all the time even for GKE clusters? not very important though and might not be that simple since with GKE we are getting the `resourceLabel`, not that I know the difference..

ZhengzheYang (approved):

asmcli/lib/installation_dependencies.sh
zerobfd: I'm not really sure if that would work since I think the data flow is different depending on whether or not it's on GCP. I'll follow up to see if we can remove this branch though.

PR #1320, opened 2022-09-08
fix: asmcli doesnt work without readlink binary on bsd-style OS

phspagiari:
Using `./asmcli` on a macos system returns an error since kpt isn't properly configured. 
As the commented line in the script says, the `stat` doesn't have the same behavior as `readlink`. It's required to use `readlink` to set the right path for kpt.

Using readlink (after installing using `brew install coreutils`) the script works as expected.
The option to install `kpt` manually doesn't work either. The version is fixed on `asmcli` to `0.39.3` and there are breaking changes in the CLI when we use the latest version.

### DIFF
--- a/asmcli/README.md
+++ b/asmcli/README.md
@@ -60,9 +60,11 @@ This script requires access to the following tools:
 - kubectl
 - sed
 - tr
+- readlink
 
-Google Cloud Shell pre-installs these tools except `kpt`, which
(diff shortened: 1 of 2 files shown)


LuisFigueiredo2021 (approved):

Felipemirand0 (approved):
LGTM

paulo-hs (approved):

PR #1347, opened 2022-10-26
MCP: make mcp options idempotent

shankgan:


zerobfd:
This means that we'll delete existing CNI=on/off config that's in the cluster. We could maybe split the values on commas (or match whatever code that parses the actual configmap) and check each k/v pair one by one.

PR #1355, opened 2022-11-11
Adding new --disable_ui option

stewartbutler:


asmcli/lib/parse.sh
```diff
@@ -245,6 +245,10 @@ parse_args() {
         context_set-option "PRINT_VERSION" 1
         shift 1
         ;;
+      --disable_ui)
+        context_set-option "UI_DISABLED" 1
```
zerobfd, line 249: I think for all of the other options the environment variable exactly matches the option name, so I think "DISABLE_UI" is more consistent.

asmcli/lib/validate.sh
```diff
@@ -330,7 +330,11 @@ exit_if_cluster_unlabeled() {
   local LABELS; LABELS="$(get_cluster_labels)";
   local REQUIRED; REQUIRED="$(mesh_id_label)";
   local NOTFOUND; NOTFOUND="$(find_missing_strings "${REQUIRED}" "${LABELS}")"
-
+  local UI_DISABLED; UI_DISABLED="$(context_get-option "UI_DISABLED")"
+  if [[ -n "${UI_DISABLED}" ]]; then
+    warn "UI Disabled by command line option; skipping cluster label check."
```
zerobfd, line 335: 1) I would do info instead of warn since ostensibly the user's intent is to disable it.

2) language nit: I think "Skipping cluster label check because UI not enabled." flows better

asmcli/lib/validate.sh
```diff
@@ -330,7 +330,11 @@ exit_if_cluster_unlabeled() {
   local LABELS; LABELS="$(get_cluster_labels)";
   local REQUIRED; REQUIRED="$(mesh_id_label)";
   local NOTFOUND; NOTFOUND="$(find_missing_strings "${REQUIRED}" "${LABELS}")"
-
+  local UI_DISABLED; UI_DISABLED="$(context_get-option "UI_DISABLED")"
+  if [[ -n "${UI_DISABLED}" ]]; then
```
zerobfd, line 334: Let's check for -eq 1 explicitly

zerobfd (changes requested):
Also, we should add DISABLE_UI to the context.sh initial struct + add it in the default zero function.

asmcli/lib/parse.sh
stewartbutler: Done.

asmcli/lib/validate.sh
stewartbutler: Done.

asmcli/lib/validate.sh
stewartbutler: Done.

stewartbutler:
All comments addressed, I think. Thanks for the pointer on context.sh.

asmcli/lib/validate.sh
```diff
@@ -330,7 +330,11 @@ exit_if_cluster_unlabeled() {
   local LABELS; LABELS="$(get_cluster_labels)";
   local REQUIRED; REQUIRED="$(mesh_id_label)";
   local NOTFOUND; NOTFOUND="$(find_missing_strings "${REQUIRED}" "${LABELS}")"
-
+  local DISABLE_UI; DISABLE_UI="$(context_get-option "DISABLE_UI")"
```
zerobfd, line 333: I realize I should have caught this sooner, but we should do this first thing in the function since the functions that get called above it might reach out to external APIs to get information.

zerobfd (approved):
One change, then LGTM

PR #1366, opened 2022-12-14
Update is_autopilot check

sergii-ssh:
On GKE 1.24+ `allowlistedworkloads.auto.gke.io` CRD is installed on both Standard and Autopilot clusters. Calling `container clusters describe` instead.


asmcli/lib/checks.sh
```diff
@@ -464,12 +464,14 @@ istio_namespace_exists() {
...
 is_autopilot() {
-  # TODO: wait till https://pkg.go.dev/google.golang.org/genproto/googleapis/container/v1#Cluster  
-  # publish the `Autopilot` field
-  # This is a temporary workaround to check if a cluster is Autopilot or GKE
-  # This CRD will be installed only if the cluster is Autopilot
   if [[ "${IS_AUTOPILOT}" -eq 1 ]]; then return; fi
-  if ! retry 2 kubectl get crd allowlistedworkloads.auto.gke.io 1>/dev/null 2>/dev/null; then
+
+  local _IS_AUTOPILOT; _IS_AUTOPILOT="$(gcloud container clusters describe "${CLUSTER_NAME}" \
+    --region "${CLUSTER_LOCATION}" \
```
ZhengzheYang, line 470: will `--region` always work? asking because some cluster might be zonal for gke Standard.

asmcli/lib/checks.sh
```diff
@@ -464,12 +464,14 @@ istio_namespace_exists() {
...
 
 is_autopilot() {
-  # TODO: wait till https://pkg.go.dev/google.golang.org/genproto/googleapis/container/v1#Cluster  
-  # publish the `Autopilot` field
-  # This is a temporary workaround to check if a cluster is Autopilot or GKE
-  # This CRD will be installed only if the cluster is Autopilot
   if [[ "${IS_AUTOPILOT}" -eq 1 ]]; then return; fi
-  if ! retry 2 kubectl get crd allowlistedworkloads.auto.gke.io 1>/dev/null 2>/dev/null; then
+
+  local _IS_AUTOPILOT; _IS_AUTOPILOT="$(gcloud container clusters describe "${CLUSTER_NAME}" \
```
ZhengzheYang, line 469: this is what worked for me:
```
gcloud container clusters  describe autopilot-cluster-1 --region ${LOCATION} --project ${PROJECT_ID} --format json | jq .autopilot.enabled
```
do we need to specify the format to be `json`?

ZhengzheYang (commented):
thanks!

asmcli/lib/checks.sh
zerobfd, line 469: _IS_AUTOPILOT implies a boolean to me. Can we somehow make this a 0 or 1?

zerobfd (changes requested):

asmcli/lib/checks.sh
sergii-ssh: yes, for zonal clusters region defaults to zone name.

asmcli/lib/checks.sh
sergii-ssh: yes, added

sergii-ssh: updated var name to match GCP field name

sergii-ssh (commented):
resolve

asmcli/lib/checks.sh
```diff
@@ -464,12 +464,15 @@ istio_namespace_exists() {
...
   if [[ "${IS_AUTOPILOT}" -eq 1 ]]; then return; fi
-  if ! retry 2 kubectl get crd allowlistedworkloads.auto.gke.io 1>/dev/null 2>/dev/null; then
+
+  local AUTOPILOT_MODE; AUTOPILOT_MODE="$(gcloud container clusters describe "${CLUSTER_NAME}" \
+    --region "${CLUSTER_LOCATION}" \
+    --project "${PROJECT_ID}" \
+    --format json | \
+    jq '.autopilot.enabled')"
+
+  if [[ "${AUTOPILOT_MODE}" = 'null' ]]; then
```
ZhengzheYang, line 475: will `${AUTOPILOT_MODE}` ever be `false`? maybe we should invert the check to be  `!= 'true'`.

ZhengzheYang (approved):

asmcli/lib/checks.sh
zerobfd: +1. I can approve/merge after this change.

sergii-ssh: thanks, updated

guangyw:
Can we have tests for this to at least catch it before the GKE version reaches regular channel?

ZhengzheYang:
> Can we have tests for this to at least catch it before the GKE version reaches regular channel?

Yeah, we do have internal test infra for different kinds of GKE cluster versions. Should we start testing with GKE 1.24? @sergii-ssh

PR #1371, opened 2022-12-14
Fix malformed membership ID in some cases

zerobfd:


ZhengzheYang (approved):

ZhengzheYang:
#1372 also tried to fix the CI. It seems like we need to follow https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke

PR #1407, opened 2023-02-25
Silence curl but keep showing errors

kynan:
The -s flag puts curl into silent mode, disabling the progress meter and
warnings or errors.

The -S flag in combination with -s re-enables error output in case curl
fails.


kynan:
Apparently the github-asm-pkg-script-updated-main failed, but I don't actually have sufficient permissions to see the Cloud Build output to determine *why* it failed, which isn't great :(

kynan:
Now it passed and again I don't know why 😅

zerobfd (approved):
LGTM, thank you!

It failed the first time because the content of `asmcli` and the files that are used to generate it didn't match, and there's a check to make sure that re-generating the tool ends up with the same content as the commit being tested.

Note for the future: It may be worth it to only do `-sSL` normally, but if the verbose flag is passed then just use `-L`. I'll merge this as-is though, and if anyone wants that change later we can handle it then.

PR #1426, opened 2023-04-08
Add logging output file in output dir

wonjekang:


zerobfd (changes requested):
Most of the logic looks good. Although I think instead of writing the log and moving it, I think it probably makes more sense just to delay writing to the log until the working directory is created and then start logging.

Maybe in the logic where we set up the working environment, we can set a LOG_FILE_CREATED variable to 1 and then in echo_log we can say something like `if [[ "${LOG_FILE_CREATED}" -eq 1 ]]; then echo $@ >> "${LOG_FILE_LOCATION}"; fi`

zerobfd (approved):
Looks perfect! I'll go ahead and approve, but can you add a test before submitting? I think something simple should work:

1) echo_log something
2) create a temporary file and set log_location to that file
3) echo_log something else
4) verify the log file has one line in it with something else

I think `asmcli/tests/lib/util.bats` might be the best place to put it.

PR #1428, opened 2023-04-13
Add --working-dir alias for output directory

wonjekang:
Add `working_dir` alias for OUTPUT_DIR context variable.


zerobfd (approved):
LGTM. Please also look over the tests to see if there's anywhere we should add one before submitting.

PR #1441, opened 2023-05-03
Remove fetching spiffe bundles from endpoints for default template

shankgan:
For Control plane Authentication, Istiod can authenticate the sidecar/gateways using JWT. Requiring Istiod to authenticate the dataplane using the mTLS cert signed by the CA is unnecessary.


zerobfd (approved):
LGTM, will merge when a domain expert takes a look.

howardjohn:
Checking something, please don't merge yet

zerobfd:
Just for posterity: there was an internal convo and we in fact didn't merge until the checking was done.

PR #1450, opened 2023-06-07
Swaps autoscaling v2 and v2beta1

cody-clark:
The quickstart is broken when the v2beta1 file is in the upper directory.

`error: resource mapping not found for name: "istio-ingressgateway" namespace: "" from "istio-ingressgateway/autoscaling-v2beta1.yaml": no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2beta1"`

Swapping the v2 hpa yaml into the upper directory and the v2beta1 into the subdirectory lets the quickstart proceed as expected.


cody-clark:
Accidentally had unintentional files added.

PR #1451, opened 2023-06-07
Hpa v2 swap

cody-clark:
The quickstart is currently broken:

```
error: resource mapping not found for name: "istio-ingressgateway" namespace: "" from "docs/shared/asm-ingress-gateway/autoscaling-v2beta1.yaml": no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2beta1"
```

Swapping the v2 hpa yaml file with the v2beta1 lets the quickstart proceed as expected.


zerobfd (approved):
This depends on the GKE cluster channel/version, right? Is there a corresponding update we should make to the quickstart to note the expected channel/version?

PR #1484, opened 2023-08-10
Use top level `istioctl` commands for `create-remote-secret` and `tag`

sergii-ssh:
Patches for https://github.com/istio/istio/pull/46051 and https://github.com/istio/istio/pull/46257


zerobfd (approved):
LGTM. I think since we use the version of istioctl downloaded by asmcli we shouldn't run into any compatibility problems, but it's worth keeping an eye out to make sure we don't get any reports after this change makes it to release.

PR #1487, opened 2023-08-28
Update verbose help text for consistency with public docs

zerobfd:


asmcli/commands/help.sh
```diff
@@ -183,6 +168,15 @@ FLAGS:
      --use_managed_cni                Deprecated. Managed CNI now defaults to be on.
      --managed                        Provision a remote, managed control plane
                                       instead of installing one in-cluster.
+                                      The channel matches with the GKE version:
+                                        GKE Channel  |  ASM Channel
+                                        -------------|----------------
+                                        Rapid        |  Rapid
+                                        Regular      |  Regular
+                                        Stable       |  Stable
+                                        Static       |  Regular
```
djazayeri, line 177: Maybe it's clearer to call this "No channel"?

Also, it has been suggested that No-channel should be Stable. At least for GSM data plane we've already decided to shift this way. Let's consider whether it might make sense to change this here.

asmcli/commands/help.sh
```diff
@@ -336,8 +330,9 @@ FLAGS:
   --version                           Print the version of this tool and exit.
 
 EXAMPLE:
-The following invocation will create a mesh using the provided 6 clusters
-by registering them to a fleet named "my_cluster" and install remote secrets on each pair of clusters:
+The following invocation will create a mesh using the provided 3 clusters
+by registering them to a fleet named "my-fleet" and install remote secrets (if
+necessary) on each pair of clusters:
```
djazayeri, line 335: nit: maybe "for each pair of clusters" instead of "on..."?

asmcli/commands/help.sh
djazayeri: Never mind the comment about no-channel => stable. That would change existing asmcli behavior, and we don't want that.

PR #1496, opened 2023-09-06
Merge staging-1.17 to release-1.17

sergii-ssh:


zerobfd (approved):
The diff looks good. I'm a bit confused as to why there are 30 commits but only 3 lines of change, but if that's how it turns out then I'm good with it.

sergii-ssh:
> The diff looks good. I'm a bit confused as to why there are 30 commits but only 3 lines of change, but if that's how it turns out then I'm good with it.

This happens because we `Squash and merge` those 30 commits from staging branch into single commit in release branch for every asmcli release.  Those changes are already there in release branch but commit SHAs doesn't match so git shows them in PR

PR #1498, opened 2023-09-15
Populate MonitoringConfig values to GCP_METADATA

wonjekang:


asmcli/lib/config.sh
```diff
@@ -66,6 +60,15 @@ configure_package() {
   configure_control_plane
 }
 
+configure_off_gcp_gcp_metadata(){
+  local HUB_MEMBERSHIP_ID; HUB_MEMBERSHIP_ID="${1}"
+  local PROJECT_ID; PROJECT_ID="${2}"
+  
+  kpt cfg set asm gcloud.core.project "$(get_monitoring_config_membership_project_id "${OFF_GCP_MEMBERSHIP_NAME}" "${PROJECT_ID}")"
```
zerobfd, line 67: Each one of these makes the same call via `gcloud`. Can you get all three values with one call to `gcloud container hub memberships describe` and then assign them to different variables?

zerobfd (changes requested):

zerobfd (approved):

PR #1515, opened 2023-10-12
Use specified channel only for revision name when installing

zerobfd:


zerobfd:
My GPG key expired on this machine, I'm going to amend the commit once I go through the process of generating a new subkey etc. etc. so that I can make sure and sign the commit.

djazayeri (approved):

PR #1567, opened 2024-01-29
rename rhel cni

hemendrateli:


wonjekang (commented):
I believe this update requires to change the flag name in prow tester. Could you send a followup CL for the fix?

https://source.corp.google.com/h/gke-internal/istio/istio/+/master-asm:prow/asm/tester/pkg/install/asm.go;l=229?q=cni-onprem-rhel&ss=h%2Fgke-internal%2Fistio%2Fistio%2F%2B%2Frefs%2Fheads%2Fmaster-asm

wonjekang (approved):

PR #1631, opened 2024-09-25
Remove profile cmd

sergii-ssh:
Follow up for https://github.com/istio/istio/pull/52468


asmcli/commands/print-config.sh
```diff
@@ -1,38 +0,0 @@
...
-    populate_fleet_info
-  fi
-  print_config >&3
-}
-
-print_config() {
-  local MANAGED; MANAGED="$(context_get-option "MANAGED")"
-
-  if [[ "${MANAGED}" -eq 1 ]]; then
-    cat "${MANAGED_MANIFEST}"
```
akshayjnambiar, line 30: for managed, it uses MANAGED_MANIFEST and do not call istioctl profile dump. Do we need to remove this(print_config) or just disable this for unmanaged?

sergii-ssh: `MANAGED_MANIFEST` has been removed a while ago - https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/pull/1449. And provisioning managed via asmcli is on a deprecation path - 
https://cloud.google.com/service-mesh/legacy/anthos-service-mesh/managed-anthos-service-mesh/provision-managed-anthos-service-mesh-asmcli

akshayjnambiar (approved):

PR #1650, opened 2024-11-29
Update cni-baremetal-rhel.yaml add spc_t seLinuxOption

kushagraj81:
Adding this for fixing the ongoing cni daemonset failing issue, due to privilege being set to false in securityContext


akshayjnambiar (approved):
LGTM

PR #1663, opened 2025-01-09
Updated the revision label validator

armaanverma:
When a revision label is passed with an Uppercase letter, the istioctl install steps fails without any context.

For other revision label conventions the validator throws the error. 

Also as per the docs upper case letters are not allowed - https://cloud.google.com/service-mesh/legacy/anthos-service-mesh/ca-migration#migrate_to

Docs


hemendrateli:
asmcli is in maintenance mode and we don't want to do any features updates. We will only fix critical customer issues in asmcli.

PR #1674, opened 2025-01-30
Managed CSC for in-cluster CSM

shavi2626:
**Change Description:**
Install Managed Canonical Service controller for in-cluster CSM.
1. Enable service mesh feature for both GCP and off-GCP platforms.
2. If in-cluster CSC deployment exists, update the deployment
3. Else, check and log the status of managed Canonical Service Controller.
4. Deprecate flag "--disable-canonical-controller" with warning message. Managed CSC will be enabled by default. Sample output [here](https://screenshot.googleplex.com/7worEfTygPp6oFk)


asmcli/commands/install.sh
```diff
@@ -101,6 +101,18 @@ install_canonical_controller() {
   info "...done!"
 }
 
+install_managed_canonical_controller() {
+  info "Installing managed canonical service controller..."
+  local IN_CLUSTER_CSC_DEP; IN_CLUSTER_CSC_DEP="$(kubectl get deployment/canonical-service-controller-manager \
+   -n asm-system --ignore-not-found=true || true)"
+  if [[ -z "$IN_CLUSTER_CSC_DEP" ]]; then
+    enable_service_mesh_feature
```
hemendrateli, line 109: Let's enable service mesh feature before control plane installation itself so that it can be used by other functions in future and we get sufficient time also for managed CSC to be enabled.

asmcli/commands/install.sh
```diff
@@ -101,6 +101,18 @@ install_canonical_controller() {
   info "...done!"
 }
 
+install_managed_canonical_controller() {
```
hemendrateli, line 104: nit `enable_managed_canonical_controller` instead of `install_managed_canonical_controller` ?

asmcli/commands/install.sh
```diff
@@ -266,7 +278,5 @@ install_control_plane() {
     if use_fleet_api; then install_fleet_api; else install_control_plane_revision; fi
   fi
 
-  if [[ "$DISABLE_CANONICAL_SERVICE" -eq 0 ]] && ! is_managed; then
```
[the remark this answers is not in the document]
hemendrateli: I see that we are doing this in install_managed_canonical_controller but we need to do `kubectl apply -f "${CANONICAL_CONTROLLER_MANIFEST}"` if in-cluster CP exists if there is change in in-cluster csc manifest.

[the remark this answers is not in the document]
shavi2626: Moved the logic to update the existing in-cluster CSC Deployment and checking the managed CSC status under one routine "verify_canonical_controller".

asmcli/commands/install.sh
shavi2626: Agree. Enabled service mesh in validate_dependencies() in commands/validate.sh. Once Chris replies, there will more clarity on what conditions to add before enabling the service mesh feature. For now enabled it for scenarios:
1. All off-GCP platforms
2. GCP platform where we can modify gcp components

asmcli/commands/install.sh
shavi2626: Agree. We are not installing the managed CSC here. Here renamed the routine "install_canonical_controller" to "verify_canonical_controller" as we will me just checking the state of managed CSC or updating the existing in-cluster CSC deployment

asmcli/commands/validate.sh
```diff
@@ -64,8 +62,9 @@ validate_dependencies() {
         exit_if_service_mesh_feature_not_enabled
       fi
     fi
+  else
+    enable_service_mesh_feature
```
hemendrateli, line 66: This will enable_service_mesh_feature for offGCP platforms in asmcli validate command also.

So asmcli install and asmcli validate both calls validate_dependencies. in asmcli validate command user can not pass `--enable*` flags so can_modify_gcp_components gives false and we exit_if_service_mesh_feature_not_enabled.

Now I think we can move out the logic from is_gcp section and add something like

```
if can_modify_gcp_components
   enable_service_mesh_feature
else
   exit_if_service_mesh_feature_not_enabled
``` 

So now we will need extra `enable_gcp_components` flag also for offGCP platforms.

shavi2626: Added the code in off-gcp section. Did not move out the code from gcp section because don't want to disrupt the order of commands if this section is used any time.
 "if [[ "${CA}" == "managed_cas" ]]; then
        x_wait_for_gke_hub_api_enablement
        x_enable_workload_certificate_on_fleet "gkehub.googleapis.com"
      fi"

asmcli/commands/install.sh
```diff
@@ -266,7 +274,5 @@ install_control_plane() {
...
```
hemendrateli, line 269: minor :- Can we cleanup DISABLE_CANONICAL_SERVICE flag from the code and remove --disable_canonical_service param from asmcli reference doc. As it is not used anywhere now so we should clean this up.

asmcli/commands/validate.sh
```diff
@@ -50,22 +50,25 @@ validate_dependencies() {
       if ! is_stackdriver_enabled; then
         enable_stackdriver_kubernetes
       fi
-      if needs_service_mesh_feature; then
```
hemendrateli, line 53: Nit :- We should delete needs_service_mesh_feature function code as it is not used anywhere now.

asmcli/lib/util.sh
```diff
@@ -853,3 +853,43 @@ get_monitoring_config_membership_json () {
...
+  local CSC_STATUS_AVAILABLE=0
+  local CS_ERROR="CANONICAL_SERVICE_ERROR"
+  local MEMBERSHIP_STATE;
+  local CODE;
+
+  for i in {1..10}; do
+    MEMBERSHIP_STATE=$( gcloud container fleet mesh describe --project "${FLEET_ID}" --format=json 2>/dev/null | \
+            jq '.membershipStates | if (. == null) then empty else . end | with_entries(select(.key| endswith("'/"${MEMBERSHIP_NAME}"'")))[]' )
+    if [[ -z "$MEMBERSHIP_STATE" ]]; then
+      echo "Membership state for ${MEMBERSHIP_NAME} not found"
```
hemendrateli, line 873: Don't we want to continue loop from here instead of going down as Membership state is not found.

asmcli/lib/util.sh
```diff
@@ -853,3 +853,43 @@ get_monitoring_config_membership_json () {
...
+  local CLUSTER_LOCATION; CLUSTER_LOCATION="$(context_get-option "CLUSTER_LOCATION")"
+  local MEMBERSHIP_NAME; MEMBERSHIP_NAME="$(generate_membership_name "${PROJECT_ID}" "${CLUSTER_LOCATION}" "${CLUSTER_NAME}")"
+  local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
+
+  local CSC_STATUS_AVAILABLE=0
+  local CS_ERROR="CANONICAL_SERVICE_ERROR"
+  local MEMBERSHIP_STATE;
+  local CODE;
+
+  for i in {1..10}; do
```
hemendrateli, line 869: 10 mins might be more, 5 mins should be good I think

asmcli/lib/util.sh
shavi2626: We can. But the else part below, will be common to this: to display retry message and sleep for 1 min. So did not continue here

shavi2626: It won't break in Line 875, Code will get empty string value or null. and we will fall to line 885.
Checked in the added unit test cases also ( test cases: 7,8,9 in test/lib/utils.bat). But to be safe w.r.t later changes in shell and its behaviour, will add continue statement here.

asmcli/lib/util.sh
shavi2626: updated to 5

asmcli/lib/util.sh
shavi2626: Updated code

asmcli/commands/validate.sh
shavi2626: Done

asmcli/commands/install.sh
shavi2626: Done

asmcli/commands/install.sh
```diff
@@ -92,12 +92,20 @@ apply_kube_yamls() {
...
-  info "Waiting for deployment..."
-  retry 3 kubectl wait --for=condition=available --timeout=600s \
+verify_canonical_controller() {
+  local IN_CLUSTER_CSC_DEP; IN_CLUSTER_CSC_DEP="$(kubectl get deployment/canonical-service-controller-manager \
+    -n asm-system --ignore-not-found=true || true)"
+  if [[ -z "$IN_CLUSTER_CSC_DEP" ]]; then
+    info "Checking Managed CanonicalService controller state..."
+    check_managed_canonical_controller_state
+  else
+    warn "Kindly migrate to managed canonical service controller. Refer <DocLink  to be added>"
```
hemendrateli, line 102: We can add something like "In-cluster canonical service controller is deprecated, please upgrade to managed canonical service controller. Refer [DocLink  to be added]"

asmcli/lib/util.sh
```diff
@@ -853,3 +853,45 @@ get_monitoring_config_membership_json () {
...
+      info "Managed Canonical Service Controller working successfully"
+      CSC_STATUS_AVAILABLE=1; break
+    elif [ "$CODE" = "WARNING" ]; then
+      if jq -r '.servicemesh.conditions[].code' <<< "$MEMBERSHIP_STATE" | grep -q "$CS_ERROR" ; then
+        warn "Managed Canonical Service Controller facing issues. Kindly refer to <wiki link>"
+        CSC_STATUS_AVAILABLE=1
+      fi
+      break
+    else
+      echo "Retry to get state code for the membership: $MEMBERSHIP_NAME"
```
hemendrateli, line 888: nit :- use RETRY_MESSAGE variable here.

shavi2626: done

asmcli/commands/install.sh
shavi2626: added

asmcli/lib/util.sh
```diff
@@ -853,3 +853,45 @@ get_monitoring_config_membership_json () {
...
+  local CLUSTER_NAME; CLUSTER_NAME="$(context_get-option "CLUSTER_NAME")"
+  local CLUSTER_LOCATION; CLUSTER_LOCATION="$(context_get-option "CLUSTER_LOCATION")"
+  local MEMBERSHIP_NAME; MEMBERSHIP_NAME="$(generate_membership_name "${PROJECT_ID}" "${CLUSTER_LOCATION}" "${CLUSTER_NAME}")"
+  local FLEET_ID; FLEET_ID="$(context_get-option "FLEET_ID")"
+  local RETRY_MESSAGE="Retry to get state code for the membership: ${MEMBERSHIP_NAME}"
+  local CSC_STATUS_AVAILABLE=0
+  local CS_ERROR="CANONICAL_SERVICE_ERROR"
+  local MEMBERSHIP_STATE;
+  local CODE;
+  # TO-DO(shavigupta): Update Doc links
```
hemendrateli, line 868: Nit - Generally the convention is to add "TODO" and not "TO-DO". This will ensure someone can just search by "TODO" in the codebase to get all TODOs.

asmcli/lib/util.sh
```diff
@@ -853,3 +853,45 @@ get_monitoring_config_membership_json () {
...
+      fi
+      break
+    else
+      echo "${RETRY_MESSAGE}"
+      sleep 60
+    fi
+  done
+
+  if [ ${CSC_STATUS_AVAILABLE} -eq 0 ]; then
+    warn "Maximum retries reached. Managed Canonical Service Controller status could not be determined. Kindly refer to <wiki link>"
```
hemendrateli, line 894: nit :- Maybe delete `Maximum retries reached` from the message as if code is warning then we are not retrying.

asmcli/lib/parse.sh
```diff
@@ -156,7 +156,8 @@ parse_args() {
         shift 1
         ;;
       --disable_canonical_service | --disable-canonical-service)
-        context_set-option "DISABLE_CANONICAL_SERVICE" 1
+        # Kept to not break customers' CI.
+        warn "The disable_canonical_service option is no longer supported. The managed Canonical Service Controller now defaults to be on."
```
hemendrateli, line 160: nit :- I would prefer "The disable_canonical_service option is no longer supported. The managed Canonical Service Controller is always provisioned."

shavi2626: Done

asmcli/lib/util.sh
shavi2626: Done

asmcli/lib/util.sh
shavi2626: Done

hemendrateli (approved):

PR #1676, opened 2025-03-18
Managed CSC for in-cluster CSM

shavi2626:
Change Description:
Install Managed Canonical Service controller for in-cluster CSM.

1. Enable service mesh feature for both GCP and off-GCP platforms.
2. If in-cluster CSC deployment exists, update the deployment
3. Else, check and log the status of managed Canonical Service Controller.
4. Deprecate flag "--disable-canonical-controller" with warning message. Managed CSC will be enabled by default.


hemendrateli:
I don't understand this, why our managed CSC changes are not there in main branch and we need to cherry pick into staging 1.25 ?

shavi2626:
> I don't understand this, why our managed CSC changes are not there in main branch and we need to cherry pick into staging 1.25 ?

Did not merge them to main branch, else have caused the test prow jobs to fail.

hemendrateli:
> > I don't understand this, why our managed CSC changes are not there in main branch and we need to cherry pick into staging 1.25 ?
> 
> Did not merge them to main branch, else have caused the test prow jobs to fail.

As we discussed on the chat, It's ok for CSC tests to fail in CI. We should merge the change.

hemendrateli (approved):

